Issue #3, opened 2020-05-27
Define CVMFS location(s) for common sbn software products


wesketchum:
@cjbackhouse brought this up in a slack channel: we don't have an obvious place to put common SBN software productions, like a SBN CVMFS area? For the moment, can request to publish all areas, but we should define the proper area(s) for this as part of the repository restructuring.

wesketchum:
ServiceDesk ticket issues for a common CVMFS area.

wesketchum:
This is Service Desk Ticket RITM0975699. Still in progress: OSG folks have been unusually slow to respond. But, all the prereqs on our side are complete.

cjbackhouse:
Looks from the ticket like this has happened. And I can ls `/cvmfs/sbn.opensciencegrid.org`
Who will be the day-to-day manager of this area? (who/how should I be asking for products to be installed there?)

wesketchum:
Yes!

Instructions for access from ticket here:

>So, Wes, you can start publishing in oasiscfs.fnal.gov.
>The local user that you have to log in as is cvmfssbn and currently you are the only person who can become cvmfssbn.
>You can add (or delete) more publishers from your experiment(s) in /home/cmvfssbn/.k5login.

@cjbackhouse I need your kerberos username, and I'll add you to the k5login. I've added Andrew, Gianluca, Tracy, Gennadiy, Joseph, Gray, and myself so far. Can add others as needed.

cjbackhouse:
Apologies, apparently github didn't notify me about this.
My kerberos principal is "bckhouse" (no "a").

wesketchum:
No worries. You've been added to the k5login: please test and see if you can login with the above instructions.

wesketchum:
This has been done, to my knowledge. Closing.

PR #4, opened 2020-05-29
Make the products directory at /cvmfs/sbnd.opensciencegrid.org/products/externals/ available

cjbackhouse:


wesketchum:
This may be mitigated/should be updated based on the outcome of #3 , so should keep track of that.

mastbaum:
@wesketchum is that a topic of discussion for tomorrow's meeting? Should decide with @cjbackhouse to merge now or wait based on the timeline for an sbn area.

cjbackhouse:
Overtaken by events. We now have /cvmfs/sbn.opensciencegrid.org/

PR #5, opened 2020-05-29
Setup more modern versions of gdb and valgrind

cjbackhouse:


cjbackhouse:
Filed https://github.com/SBNSoftware/sbncode/issues/53 instead

PR #36, opened 2020-10-26
Added baseline variable from Event.hh to CAFs

FernandaPsihas:


wesketchum:
@FernandaPsihas , does this need to be part of an upcoming 2020A release, or is it more future-looking?

cjbackhouse:
This is very important for oscillation analysis to get the SBND oscillations correct.

FernandaPsihas:
Yes, this needs to be in whatever 2020A release we use to produce CAFs.

mastbaum:
@FernandaPsihas just to confirm, looks like this field already existed in `sbncode/StandardRecord/SRTrueInteraction.h` but wasn't filled until now?

In any case, looks good to me and matches what I had in sbnanalysis, as suggested by Zarko: https://github.com/SBNSoftware/sbncode/blob/853a2281d0192dabe893e4f584b170942bcb5dc1/sbnanalysis/core/ProcessorBase.cxx#L454

@wesketchum as reviewer on this should I go ahead merge to develop via GH?

FernandaPsihas:
Yep, @mastbaum 
The variable existed already and now it is just being filled for the osc analysis.

wesketchum:
I'll handle the merge (well, just did): since it's a trivial one, no problems doing it now. Will be part of the next release.

PR #50, opened 2020-11-05
Restore the status quo from before the sbncaf_srproxy branch

cjbackhouse:
Restore the status quo from before the sbncaf_srproxy branch overwrote it. No interesting CAFAna macros will run without the SRProxy generated files installed.


wesketchum:
Sorry, made the commit history a bit jumbled, but just updated this to the recent version of larsoft. No other changes. Checked it builds with c7.

wesketchum (approved):

cjbackhouse:
thanks

PR #51, opened 2020-11-05
Feature/ext osclib

cjbackhouse:
Pick up OscLib as an external product rather than duplicating it in sbncode.
    
The product has abbreviated of the names, which entails a lot of mechanical changes.


sbncode/CAFAna/Core/CMakeLists.txt
```diff
@@ -3,8 +3,13 @@ include_directories( $ENV{IFDHC_FQ_DIR}/inc )
 # Allow "StandardRecord/" to find the right directory
 include_directories(..)
 
+include_directories($ENV{OSCLIB_INC})
+
 art_make(LIBRARY_NAME CAFAnaCore
          LIB_LIBRARIES StandardRecordProxy
+                       # TODO how am I supposed to link to an external library?
```
cjbackhouse, line 10: @wesketchum just ${OSCLIB} didn't work here like I expected. I think there a couple other instances of this pattern too.

cjbackhouse:
I'd like to merge this before the sbnana split too. Otherwise I will have to recreate it over in sbnana. The osclib dependency that it adds will only be incurred by sbnana, which is one of the main points of doing the split.

wesketchum:
I think that's not a problem at all: we'll do this merge before doing the plot.

cjbackhouse:
Superseded by #78 and merged there

PR #59, opened 2020-11-17
Simple Flash Matching migration and update

ikatza:
I'm migrating the Flash Matching code, the simple weighted light and charge centres one, from `sbndcode` to here so that it can be more easily shared.

I've made a number of updates to accommodate for changes in ICARUS and to better log issues and the reasons why a particle might not be scored and have an associated T0.

This PR should be merged along with updates to `sbndcode` and `icaruscode`, and respective metrics files on `sbnd_data` and `icarus_data`.


gputnam:
Thanks so much Iker!

To the reviewers -- Bruce, Michelle (others?). Can we merge this in tomorrow morning? Having this code in sbncode ASAP is very necessary to do the required work preparing for the PAC.

mstancar:
Yes please merge

On Nov 18, 2020, at 6:58 PM, gputnam <notifications@github.com<mailto:notifications@github.com>> wrote:



Thanks so much Iker!

To the reviewers -- Bruce, Michelle (others?). Can we merge this in tomorrow morning? Having this code in sbncode ASAP is very necessary to do the required work preparing for the PAC.

brucehoward-physics:
I sent Iker a DM on Slack this evening with a question about the flash locations. In a cosmic+nu sample the flashes all seem to be in cryo0 for ICARUS. Since that's where we typically do the neutrino generation that might be by design, but I think the cosmics are distributed in both cryos(?). Am not sure if there are implications to this.

In any case, running the fm_scores fcl file, the scores extended to larger values with a nu+cosmic file than a nue file, as I might expect.

gputnam:
@Bruce -- this module only looks at in-time-light, which probably explains why you only see flashes in the cryostat with the neutrino.

ikatza:
Actually it only runs on Cryostat 0. This is due to my ignorance of how to run ICARUS files and wether the module would fail if the input file doesn't have anything on the other Cryostat.

wesketchum:
Sorry, I'm not sure I understand the status here.

Is this validated to work with ICARUS or not? Is it important to merge regardless of that, to make things available for SBND?

brucehoward-physics:
We've tested that it runs but it was only giving flashes in cryo0 hence my curiosity and follow-up with Iker last night/this morning.

In talking with Iker I think we can test today if it will produce flashes in both cryos if we change a few lines. And then if it still works in this case when there is only light in cryo0.

brucehoward-physics:
Tested with the cryo1 stuff uncommented and ran on a nu+cosmic (so stuff in both cryos) and a nue file (which should be cryo0 only). In both cases, the fm_scores_icarus fcl ran and produced the output trees for both instances, fmatchCryo0 and fmatchCryo1

There were more entries in cryo0 than cryo1 in the nu+cosmic case. Possibly for the reason Gray notes above. In the nu-only case, there was one flash matched in the other cryo.

wesketchum:
OK, great, so this is good to go then? I can create a path release that can move alongside the ICARUS patch release too. (@SFBayLaser )

ikatza:
**Status:** 

- I have tested in SBND for neutrino only and neutrino + cosmics events, works ok. 
- It has been tested in ICARUS for neutrino only and neutrino + cosmics events, both by me and @brucehoward-physics.
- Runs ok for both ICARUS cryostats, irregardless if the input file only has products from one. I've updated the fcls.
- I have produced up to date metrics files for both experiments. The specific `sbndcode` and `icaruscode` PRs highlight this. 
- I'm currently discussing about where to put the respective fcl files for each repository.

I defer to the software managers on this last point. But want to note that we _could_ delay the decision as it doesn't prevent proper functioning.

Other than that, the verbosity and logging could be improved/changed. But it can definitively wait to a future PR.

brucehoward-physics:
Wanted to check in to see if this can be/is going to be merged? Thanks for any updates.

mendezdpatricia:
It seems like all the tests look fine(?). In addition, yesterday we also tested that it worked for producing ICARUS CAFs with the flash matching info. 
When can we merge this 😬@SFBayLaser @wesketchum ?

wesketchum:
Working on merging this now. Sorry for the delay!

PR #67, opened 2021-01-27
Feature/dmendez fd nuesel

mendezdpatricia:
Making this pull request as advertised to do for all active branches.
This is the branch that I've been using for the purpose of the nue selection in ICARUS.
I don't see why it wouldn't be ok to merge this into develop, but I also don't see a problem with leaving this as a branch as long as I get some help with making sure that it still works once the sbncode-sbnana split happens.


wesketchum:
Thanks Diana. Can you suggest/assign someone in this PR to review this to verify (at least within reason) that it's ok to merge to develop? From your description it sounds to be the case. Assigning me is an option, but I'm not as familiar with the details so could miss something obvious.

mendezdpatricia:
Right, this should affect anyone else's work. It mostly only includes a bunch of scripts that use SBNAna (Vars,Cuts,Spectra) to perform the event selection, all which are in SBNAna/ananue/eventselection , and a few additions and edits that should only affect the FD's Vars and Cuts. 
Assigned Ed and Chris to review.

etyley (approved):
These changes all look good to me from an SBND NuE Selection point of view.

wesketchum:
OK, this is now in release/v09_15_00 and is awaiting the split.

PR #76, opened 2021-01-28
Merging Developed Stuff

gputnam:
This pull request incorporates changes from a number of different studies in SBN -- low-energy proton identification, track charge reconstruction, single-to-noise for deconvolution filtering, analysis of track Calorimetry, etc.

They exist as standalone Analysis and Producer modules organized into the NuMuSelection (for numu specific stuff) and TPCReco (for more general stuff) directories.

The changes here do not affect CAFMaker or other modules in common use.

My hope is that this pull request will be merged in before the split of sbncode is done.


wesketchum:
OK, this is now in release/v09_15_00 and is awaiting the split.

Note that I think this was rather old, and needed a lot of products removed/moved to sbnobj. Please do not use this branch again for work once we have the new release, and instead use a new branch based off of develop (after the release is finished).

PR #80, opened 2021-02-02
Use fullpath of shared object, indentation whitespace removal

ikatza:


ikatza:
I would suggest to also rename the shared object of `CAFMaker` so that it also comes with the full path.

In my opinion it's easier to let the build facility (I guess `cetbuild`) to decide the names instead of manually setting them. But I wouldn't claim I know how it works, I only copy paste code.

wesketchum (approved):

wesketchum:
Agreed on CAFMaker renaming. Doing that now, with the new release.

PR #87, opened 2021-02-11
Add in more MCFlux stuff. Add in range-based pion momentum.

gputnam:
Depends on: https://github.com/SBNSoftware/sbnanaobj/pull/5.

### DIFF
--- a/sbncode/LArRecoProducer/RangePAllPID_module.cc
+++ b/sbncode/LArRecoProducer/RangePAllPID_module.cc
@@ -18,6 +18,8 @@
 #include "messagefacility/MessageLogger/MessageLogger.h"
 #include "lardata/Utilities/AssociationUtil.h"
 
+#include "TDatabasePDG.h"
+
 #include "LArReco/TrackMomentumCalculator.h"
 #include "sbnobj/Common/Reco/RangeP.h"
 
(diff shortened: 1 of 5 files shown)


wesketchum:
@FernandaPsihas or @etyley : could one of you give a quick look over on this to make sure it's ok to merge in? ~Today would be good so it can be in the release this week (though I think we can merge it to develop whenever it's ready regardless).

wesketchum:
(please also check out the sbnanaobj branch needed as well)

sbncode/LArRecoProducer/RangePAllPID_module.cc
```diff
@@ -49,9 +49,12 @@ class sbn::RangePAllPID : public art::EDProducer {
   art::InputTag fTrackLabel;
 
 };
-const static std::vector<int> PIDs {13, 2212};
-const static std::vector<std::string> names {"muon", "proton"};
+const static std::vector<int> PIDs {13, 211, 2212};
+const static std::vector<std::string> names {"muon", "pion", "proton"};
 
+// P.A. Zylaet al.(Particle Data Group), Prog. Theor. Exp. Phys.2020, 083C01 (2020)
```
etyley, line 55: I am curious why this is hardcoded vs using the `TDatabasePDG` that is used in the CAFMaker, although I doubt it will change the physics output in this case.

etyley:
I added some comments to this and the `sbnanaobj` PRs which depend how strict you want to be regarding sticking to conventions but physics-wise this all looks good to me

sbncode/LArRecoProducer/RangePAllPID_module.cc
gputnam: Fixed!

gputnam:
@wesketchum I have addressed the comments in this pull request and the upstream one.

wesketchum:
Done and done. Thanks all!

PR #89, opened 2021-02-22
Added missing CMake include directory

PetrilloAtWork:
I could not make CMake happy without adding `${ART_DIR}/Modules` to its include path (`ArtMake.cmake`).
It is not clear to me if the presence or order of other repositories can make the build work without this modification.


PetrilloAtWork:
Pull request from the wrong branch... set new pull request #90.

PR #91, opened 2021-03-02
Get Calorimetry Analysis working in most recent icaruscode.

gputnam:
title. 

This update doesn't change anything related to CAF. 

It's an update to a standalone module which has been used by a couple people in SBND/ICARUS to evaluate the calorimetry. There have been some updates in the ICARUS channel mapping which breaks stuff. This pull request remedies those issues.


wesketchum (approved):

wesketchum:
@gputnam looks ok to me -- asking @SFBayLaser to have a look if he also has time just to double-check, but shouldn't block for next release.

PR #93, opened 2021-03-03
Fill new variables added in sbnanaobj branch feature/gtruth

cjbackhouse:
The majority of these come from the simb::GTruth object.


gputnam:
Does update introduce any problems if you run it over something other than neutrino MC?

I.e. does it work even if the GTruth object isn't present.

wesketchum (approved):

cjbackhouse:
I admit I haven't tested it, but I look for objects under the same genLabel as existing code. I think that means that if you specify "" it should just skip filling this?
If you can recommend  a no-genie sample I can test.

gputnam:
Well.... it's merged in now, so I guess we'll find out.

PR #94, opened 2021-03-17
Feature/bckhouse metadata service

cjbackhouse:
This is a direct port of the metadata classes from sbndcode, followed by some refactoring to allow an external user to grab the metadata maps.

Configuration changes:
- the service name is now MetadataSBN rather than TFileMetadataSBND
- GenerateTFileMetadata no longer exists. You may set the json filename blank if you don't want it
- There is an Experiment parameter, defaulting to the $EXPERIMENT env var. If we decide to unify sbnd_project and icarus_project we could set that to "sbn"

### DIFF
--- a/sbncode/CMakeLists.txt
+++ b/sbncode/CMakeLists.txt
@@ -27,6 +27,7 @@ add_subdirectory(OpDet)
 add_subdirectory(OpT0Finder)
 
 add_subdirectory(CAFMaker)
+add_subdirectory(Metadata)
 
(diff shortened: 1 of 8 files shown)


cjbackhouse:
I am happy to do that, or have someone else do that. I just didn't want to pull the rug out from under anything.

etyley:
If you are okay to do that, I will make a complementary SBNDUtil pull request to point production to the new service so they can all be merged in one release?

PR #98, opened 2021-03-23
Feature/bckhouse_caf_wgts

cjbackhouse:
CAFMaker logic to fill weights into CAFs


wesketchum:
Don't know if @mastbaum can take a quick look or if he's the right person, but just in case requested review.

sbncode/CAFMaker/CAFMaker_module.cc
```diff
@@ -665,11 +753,13 @@ void CAFMaker::produce(art::Event& evt) noexcept {
     }
   }
 
+  // TODO eventually we want to handle multiple sources of weights
```
mastbaum, line 756: Yes: in the future there may be multiple producers (GENIE weights, flux weights, etc.). Within a single run of the event weight producer we can mix these calculators arbitrarily, but do want to support factoring this out, generating weights across multiple jobs, etc.

mastbaum (approved):
This looks good (within my limited CAF experience). See note about future support of weights from multiple producers.

sbncode/CAFMaker/CAFMaker_module.cc
cjbackhouse: Yep. I was too scared to attempt this right now, especially since there is no example file configured that way to test with. I guess we could open an issue to track this.

PR #103, opened 2021-03-26
Fill the ParameterSet info in beginRun()

cjbackhouse:
so that it's soon enough for downstream modules (CAFMaker) to use it.

Turns out that we had eventweight+cafmaker working, but only in separate jobs. If you run them in a single job, which is the plan, then you run into this need for time travel.


etyley:
I have tested this locally for SBND and it now runs and produces the expected CAFs. If we could get a new release cut then I can start production tests.

PR #108, opened 2021-04-01
Feature/dmendez icarus cafweights

mendezdpatricia:
Edits to Icarus cafmaker for event weights.

### DIFF
--- a/sbncode/CAFMaker/cafmakerjob_icarus.fcl
+++ b/sbncode/CAFMaker/cafmakerjob_icarus.fcl
@@ -18,7 +18,7 @@
 #include "crthitconverter_producer.fcl"
 #include "pmtflashtrigger_producer.fcl"
 #include "flashmatch_simple_icarus.fcl"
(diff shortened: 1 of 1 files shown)


cjbackhouse:
The genieweight stuff looks fine. The sbnd fcls have (too many) variants with/without genieweight, with different readout lengths, with/without SCE etc, whereas here it looks like there's a single fcl and you've just turned on genieweight and changed the geometry unconditionally. I guess this is a icarus question about how things should be organized.

For the PROLOG issue, it would be good to figure out which of the headers actually caused the problem, and add the BEGIN/END_PROLOG there if appropriate.

wesketchum:
@cjbackhouse so is this ok? If so can you do the review interface here and say approve? Trying to use these tools so long as they are useful.

cjbackhouse (approved):
This is all good. Comments were about potential future directions to tidy up.

etyley (approved):
I have nothing to add to what Chris has said

Issue #107, opened 2021-03-29
SBN Event Weight throws exception when running on cosmic-only sample.


etyley:
When running tests for the upcoming production I found that the event weight refused to run on a sample without neutrino interactions, meaning we must use separate workflows in the production, which was not the expected behaviour after discussion in the SBN analysis infrastructure meeting. The error was:
```%MSG-s ArtException:  PostEndJob 29-Mar-2021 03:23:51 CDT ModuleEndJob
---- EventProcessorFailure BEGIN
  EventProcessor: an exception occurred during current event processing
  ---- ScheduleExecutionFailure BEGIN
    Path: ProcessingStopped.
    ---- ProductNotFound BEGIN
      Found zero products matching all selection criteria
        C++ type: std::vector<simb::MCTruth>
        Module label: 'generator'
        Product instance name: ''
        Process name: (empty)The above exception was thrown while processing module SBNEventWeight/genieweight run: 13 subRun: 42 event: 2
    ---- ProductNotFound END
    Exception going through path runprod
  ---- ScheduleExecutionFailure END
---- EventProcessorFailure END
%MSG
```

cjbackhouse:
I was thinking like this:
https://github.com/SBNSoftware/sbncode/compare/feature/evtwgt_cosmic

So you can either set the generator label to `""`, or if you want to run the same fcl over both genie and cosmics you can leave it set to "`generator`" but set the option to allow it to continue even if it doesn't find MCTruths.

I'm not proposing a merge just yet because I didn't quite understand the fcl structure. Where should I be writing that `AllowMissingTruth` exists, and writing a big comment about it?

gputnam:
I want to add one other comment to this issue. On this point (from the other issue):

> I would argue that CAFMaker is working as designed - if you don't intend to have weights in the CAF, set the label to "". If you aren't sure / want best-effort, set StrictMode: false. I don't know why the weights should necessarily treated differently to any other product in this respect.

I disagree. The problem is that the large majority of times we run CAF, the input files don't contain systematic weights (even MC files). This isn't like a piece of reconstruction information that should always be present. Having a base configuration of the CAFMaker that fails over the majority of files is a problem -- we already saw issues with getting this to work for the workshop production.

I don't totally care what direction we take to fix this. We could have the weights label be `""` in the base config, and add in an extra weights config for running CAFMaker specifically on files that contain weights. Or we could have the code that grabs the weights fail gracefully. Either is fine with me.

[...]

PR #110, opened 2021-04-01
Feature/evtwgt_cosmic

cjbackhouse:
There are several different small fixes for cosmics/geniewgt/cafs in here.

Some discussed here https://github.com/SBNSoftware/sbncode/issues/107
and the final commit with the fcl changes in the meeting today. I don't think there's an associated issue.

### DIFF
--- a/sbncode/CAFMaker/CAFMaker.fcl
+++ b/sbncode/CAFMaker/CAFMaker.fcl
@@ -13,7 +13,7 @@ standard_cafmaker:{
     DataTier:             "caf" # Provide a string to override automatic filename
     FileExtension:        ".caf.root"
     GeneratorInput:       "generator"
-    SystWeightLabel:      "genieweight"
+    SystWeightLabel:      "" # If weights exist they are likely "genieweight"
     StrictMode:           false    # Abort if any required product is not found, unless the label is empty
 }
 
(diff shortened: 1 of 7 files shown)


cjbackhouse:
I didn't do anything to the icarus cafmaker fcl. I think whether that wants to be split or not is up for discussion.

I tested that `lar -c cafmakerjob_sbnd_3drift_sce_runfmatch.fcl /pnfs/sbnd/mc/reco2/artroot/official/MCP2020A/prodcorsika_proton_intime_filter/sbndcode/v09_09_00_01/run_number/00/00/00/13/subrun_number/00/00/00/42/reco2-68cbb816-750b-4ec1-847a-ecec01877cfb.root` runs. It also works with the `_geniewgt` suffix, though wastes some effort setting up the calculators. I think that means that in principle icarus could get away with one fcl.

gputnam:
This looks good to me, except I think @etyley had suggested that the SystWeightLabel be set to "" by default (in CAFMaker.fcl) and overridden to "genieweight" for the specific configuration that needs it.

gputnam (changes requested):
see comment above.

cjbackhouse:
Done

gputnam (approved):

gputnam:
Great! Thanks.

PR #118, opened 2021-05-19
feature/jzettleNewFMObject

jzettle:
Pull request for the addition of a new sbn::SimpleFlashMatch data product replacing the anab::T0 object that was used previously. This is largely done to improve the amount of stored information from the simple flash match object in the CAF files.  Tested on SBND and ICARUS simulations and shown to fill the correct object in the CAF files.


ikatza:
@FernandaPsihas could you please review the changes to the files that are under CAFMaker?

FernandaPsihas (approved):

ikatza (approved):

cjbackhouse:
@miquelnebot I think it would probably be worth putting this through a build test with the corresponding PR from sbnanaobj. Otherwise looks good to me so far as I can see.

PR #119, opened 2021-05-26
Feature/jlarkin fakereco

jacoblarkin:
Fill fake reco variables for slices in CAFs.
This depends on the fake reco pull request I made in sbnanaobj.

### DIFF
--- a/sbncode/CAFMaker/FillTrue.cxx
+++ b/sbncode/CAFMaker/FillTrue.cxx
@@ -109,6 +109,26 @@ namespace caf {
 
   }//FillSliceTruth
 
+ void FillSliceFakeReco(const std::vector<art::Ptr<recob::Hit>> &hits,
+                         const std::vector<art::Ptr<simb::MCTruth>> &neutrinos,
+                         const std::vector<caf::SRTrueInteraction> &srneutrinos,
+                         const cheat::ParticleInventoryService &inventory_service,
+                         const detinfo::DetectorClocksData &clockData,
+                         caf::SRSlice &srslice, caf::SRTruthBranch &srmc,
+                         const std::vector<art::Ptr<sim::MCTrack>> &mctracks,
+                         const std::vector<geo::BoxBoundedGeo> &volumes, TRandom &rand,
+                         bool allowEmpty)
+  {
+    caf::SRTruthMatch tmatch = MatchSlice2Truth(hits, neutrinos, srneutrinos, inventory_service, clockData);
+    if(tmatch.index >= 0) {
+      bool do_fill = false;
+      caf::SRFakeReco this_fakereco;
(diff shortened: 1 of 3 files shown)


sbncode/CAFMaker/FillTrue.h
```diff
@@ -38,6 +38,16 @@ namespace caf
...
 
+  void FillSliceFakeReco(const std::vector<art::Ptr<recob::Hit>> &hits,
+                         const std::vector<art::Ptr<simb::MCTruth>> &neutrinos,
+                         const std::vector<caf::SRTrueInteraction> &srneutrinos,
+                         const cheat::ParticleInventoryService &inventory_service,
+                         const detinfo::DetectorClocksData &clockData,
+                         caf::SRSlice &srslice, caf::SRTruthBranch &srmc,
+                         const std::vector<art::Ptr<sim::MCTrack>> &mctracks,
+                         const std::vector<geo::BoxBoundedGeo> &volumes, TRandom &rand,
+                         bool allowEmpty = false);
```
cjbackhouse, line 49: `allowEmpty` doesn't seem to do anything. Just remove it.

sbncode/CAFMaker/FillTrue.cxx
```diff
@@ -109,6 +109,26 @@ namespace caf {
...
+                         const std::vector<art::Ptr<sim::MCTrack>> &mctracks,
+                         const std::vector<geo::BoxBoundedGeo> &volumes, TRandom &rand,
+                         bool allowEmpty)
+  {
+    caf::SRTruthMatch tmatch = MatchSlice2Truth(hits, neutrinos, srneutrinos, inventory_service, clockData);
+    if(tmatch.index >= 0) {
+      bool do_fill = false;
+      caf::SRFakeReco this_fakereco;
+      do_fill = FRFillNumuCC(*neutrinos[tmatch.index], mctracks, volumes, rand, this_fakereco);
+      if(do_fill) srslice.freco = this_fakereco;
```
cjbackhouse, line 127: I think we can simplify this a bit. It looks to me like `FRFillNumuCC` doesn't touch anything in the `fakereco` object in the case it returns false, so I think this can just be
```if(tmatch.index.index >= 0) FRFillNumuCC(..., srslice.freco)```

File: sbncode/CAFMaker/FillTrue.cxx [hunk 3]
```diff
@@ -109,6 +109,26 @@ namespace caf {
...
+    }
```
cjbackhouse, line 128: What happens if `tmatch.index` is bad, or `FRFillNumuCC` returns false for whatever reason? I think it might be best to have some flag in the fake reco object to indicate if the fake reco was done or not.

cjbackhouse (approved):

PR #121, opened 2021-06-03
Fix flashX unfolding bug

ikatza:
I missed updating a variable, thus the global X coordinate of the flash wasn't correct.

 The bug itself only affects ICARUS, and not SBND.


miquelnebot (approved):
Looks OK

PR #123, opened 2021-06-04
fix constructor for MeVPrtlTruth

mroda88:


mroda88:
trigger build

miquelnebot (approved):
OK I assume

PR #127, opened 2021-06-14
Feature/gputnam caloana

gputnam:
Updates to the CalorimetryAnalysis module.

Small update to just some analysis -- orthogonal to most activities in sbncode.


PetrilloAtWork (changes requested):
Summary of the comments:

1. debugging messages must be moved out of `std::cout` before merging;
2. there is once again the suspicion that some code is not ICARUS-ready; in this case my impression is that the issue is not fundamental, but rather due to the carefree jumping of the original code between plane number and view; I have proposed a change that might overcome that issue in the place I have identified; this change needs to be counter-reviewed though!
3. there is the usual request that missing data products trigger a big fat exception unless explicitly allowed.

I consider the first point a requirement, the second almost a requirement.
The rest of the comments should be considered by the author and acted upon according to his judgement.

sbncode/TPCReco/CalorimetryAnalysis/run_calorimetry_analysis_icarus_sce_multistream.fcl
```diff
@@ -0,0 +1,160 @@
...
+services.SpaceCharge: @local::icarus_spacecharge
+services.SpaceCharge.EnableCalEfieldSCE: true
+services.SpaceCharge.EnableCalSpatialSCE: true
+services.SpaceCharge.EnableSimEfield: true
+services.SpaceCharge.EnableSimEfieldSCE: true
+services.SpaceCharge.EnableSimSpatialSCE: true
+services.SpaceCharge.EnableCorrSCE: true
+services.SpaceCharge.InputFilename: "SCEoffsets/SCEoffsets_ICARUS_E500_voxelTH3.root"
+services.SpaceCharge.RepresentationType: "Voxelized_TH3"
+services.SpaceCharge.service_provider: SpaceChargeServiceICARUS
```
PetrilloAtWork, line 43: How is `SpaceCharge` different from `SpaceChargeService`?

As an aside, I have opened an issue with `icaruscode`, SBNSoftware/icaruscode#236, that would be relevant here since there is way too much boilerplate configuration. @grayputnam's help in defining the content re that issue would be welcome.

sbncode/TPCReco/CalorimetryAnalysis/run_calorimetry_analysis_icarus_sce_multistream.fcl
```diff
@@ -0,0 +1,160 @@
...
+  # @table::icarus_detsim_services
+  RandomNumberGenerator: {} #ART native random number generator
+  TFileService: { 
+    fileName: "calorimetryout.root"
+  }
+  message:      { debugModules: ["*"] destinations: { debugmsg:{type: "cout" threshold: "INFO"} } }
+  TimeTracker:  { printSummary: true }
+  AuxDetExptGeoHelperInterface: { service_provider: "icaruscode/CRT/CRTGeometryHelper" }
+  BackTrackerService:  @local::standard_backtrackerservice
+  ParticleInventoryService: @local::standard_particleinventoryservice
```
PetrilloAtWork, line 30: Pretty much all of this is overwritten by the last two `@table` references, with possibly one exception:
```suggestion
  AuxDetExptGeoHelperInterface: { service_provider: "icaruscode/CRT/CRTGeometryHelper" }
```
which I am not sure is even necessary (is CRT needed? is `AuxDetGeometry` loaded at all?).

sbncode/TPCReco/CalorimetryAnalysis/run_calorimetry_analysis_icarus_sce_multistream.fcl
```diff
@@ -0,0 +1,160 @@
...
+}
+
+outputs:
+{
+  outpid:
+  {
+    module_type: RootOutput
+    fileName:    "%ifb_%tc-%p.root"
+
+    fastCloning: false
```
PetrilloAtWork, line 60: Why are we disabling fast cloning?

sbncode/TPCReco/CalorimetryAnalysis/run_calorimetry_analysis_icarus_sce_multistream.fcl
```diff
@@ -0,0 +1,160 @@
...
+  }
+}
```
PetrilloAtWork, line 62: Considering that no _art_/ROOT output is produced, this should be omitted not to confuse:
```suggestion
```

wesketchum:
Assigning @gputnam  ... status on this and the review, @PetrilloAtWork ? Feel we should be pushing to resolve this.

PetrilloAtWork:
As far as I can tell, the debug messages are still there in `std::cout`, so I assume Gray hasn't found the time to go through the suggestions.

wesketchum:
Status on this? @gputnam

gputnam:
This PR has been stale for a while -- I'm closing it.

PR #128, opened 2021-06-17
New track calo skimmer module for ICARUS calibration stuff.

gputnam:
New module. Orthogonal to existing stuff. Depends on https://github.com/SBNSoftware/sbnobj/pull/17.


sbncode/Calibration/TrackCaloSkimmerSelectStoppingTrack_tool.cc
```diff
@@ -0,0 +1,183 @@
...
+    (t.hit_max_time_p2_tpcE < 0. || t.hit_max_time_p2_tpcE < fFidTickMax) &&
+    (t.hit_min_time_p2_tpcW < 0. || t.hit_min_time_p2_tpcW > fFidTickMin) &&
+    (t.hit_max_time_p2_tpcW < 0. || t.hit_max_time_p2_tpcW < fFidTickMax);
+
+  std::cout << "Track times: " << t.hit_min_time_p2_tpcE << " " << t.hit_max_time_p2_tpcE << " " << t.hit_min_time_p2_tpcW << " " << t.hit_max_time_p2_tpcW << " is FID: " << time_is_fid << std::endl;
+
+  // compute the median dqdx of the last 5 cm
+  std::vector<double> endp_dqdx;
+  for (const sbn::HitInfo &h: t.hits2) {
+    if (h.oncalo && h.rr < 5.) endp_dqdx.push_back(h.dqdx);
```
wesketchum, line 146: Suggestion: make configurable?

sbncode/Calibration/TrackCaloSkimmerSelectStoppingTrack_tool.cc
```diff
@@ -0,0 +1,183 @@
...
+  for (const geo::BoxBoundedGeo &g: fFiducialVolumes) {
+    geo::Point_t end {t.end_x, t.end_y, t.end_z};
+
+    if (g.ContainsPosition(end)) {
+      end_is_fid = true;
+      break;
+    }
+  }
+
+  std::cout << "Track end: " << t.end_x << " " << t.end_y << " " << t.end_z << " is FID: " << end_is_fid << std::endl;
```
wesketchum, line 132: Wrap in a verbosity option.

sbncode/Calibration/TrackCaloSkimmerSelectStoppingTrack_tool.cc
```diff
@@ -0,0 +1,183 @@
...
+
+  // Collection plane times need to be fiducial
+  bool time_is_fid = \
+    (t.hit_min_time_p2_tpcE < 0. || t.hit_min_time_p2_tpcE > fFidTickMin) &&
+    (t.hit_max_time_p2_tpcE < 0. || t.hit_max_time_p2_tpcE < fFidTickMax) &&
+    (t.hit_min_time_p2_tpcW < 0. || t.hit_min_time_p2_tpcW > fFidTickMin) &&
+    (t.hit_max_time_p2_tpcW < 0. || t.hit_max_time_p2_tpcW < fFidTickMax);
+
+  std::cout << "Track times: " << t.hit_min_time_p2_tpcE << " " << t.hit_max_time_p2_tpcE << " " << t.hit_min_time_p2_tpcW << " " << t.hit_max_time_p2_tpcW << " is FID: " << time_is_fid << std::endl;
```
wesketchum, line 141: Wrap in a verbosity option.

sbncode/Calibration/TrackCaloSkimmerSelectStoppingTrack_tool.cc
```diff
@@ -0,0 +1,183 @@
...
+    if (endp_dqdx.size() % 2 == 0) {
+      unsigned other_middle = middle - 1;
+      std::nth_element(endp_dqdx.begin(), endp_dqdx.begin() + other_middle, endp_dqdx.end());
+      med_dqdx = (med_dqdx + endp_dqdx[other_middle]) / 2.;
+    }
+  }
+
+  bool valid_med_dqdx = ((med_dqdx > 0.) && (med_dqdx > fEndMediandQdxCut)) || (fEndMediandQdxCut < 0.);
+
+  std::cout << "Track MED dqdx: " << med_dqdx << " is valid: " << valid_med_dqdx << std::endl;
```
wesketchum, line 164: Wrap in a verbosity option.

sbncode/Calibration/TrackCaloSkimmerSelectStoppingTrack_tool.cc
```diff
@@ -0,0 +1,183 @@
...
+  bool valid_stopping_fit = t.n_fit_point > 2 || !fRequireFit;
+
+  // Make a cut on the stopping fits
+  bool valid_stopping_expR = (t.exp_fit_R < fExpFitRCut) || (fExpFitRCut < 0.) || !fRequireFit;
+
+  double reduced_residual_offset = (t.const_fit_residuals - t.exp_fit_residuals) / t.n_fit_point;
+
+  bool valid_stopping_residuals = (reduced_residual_offset < fFitResidualsCut) || (fFitResidualsCut < 0.) || !fRequireFit;
+
+  std::cout << "Fit is valid: " << valid_stopping_fit << " " << valid_stopping_expR << " " << valid_stopping_residuals << std::endl;
```
wesketchum, line 176: Wrap in a verbosity option.

sbncode/Calibration/TrackCaloSkimmer_module.cc
```diff
@@ -0,0 +1,460 @@
...
+// Generated at Mon May 17 09:46:34 2021 by Gray Putnam using cetskelgen
+// from cetlib version v3_11_01.
+////////////////////////////////////////////////////////////////////////
+
+#include "TrackCaloSkimmer.h"
+
+#include "art/Utilities/make_tool.h"
+
+// Global functions / data for fitting
+const size_t MAX_N_FIT_DATA = 30;
```
wesketchum, line 15: Make configurable?

sbncode/Calibration/TrackCaloSkimmer_module.cc
```diff
@@ -0,0 +1,460 @@
...
+
+static int N_FIT_DATA = -1;
+static double FIT_RR[MAX_N_FIT_DATA];
```
wesketchum, line 18: Use a std::vector? Does this need to be static I guess?

sbncode/Calibration/TrackCaloSkimmer_module.cc
```diff
@@ -0,0 +1,460 @@
...
+
+  std::vector<art::Ptr<recob::PFParticle>> PFParticleList;
+  try {
+    art::ValidHandle<std::vector<recob::PFParticle>> pfparticles = e.getValidHandle<std::vector<recob::PFParticle>>(fPFPproducer);
+    art::fill_ptr_vector(PFParticleList, pfparticles);
+  }
+  catch(...) {
+      std::cout << "PFP's with tag: " << fPFPproducer << " not present.\n";
+      // Real data may have missing products -- just ignore the event
+      if (e.isRealData()) return;
```
wesketchum, line 111: Do we really want this to fail silently if there is no data? I think we probably do want it to crash in that case...right?

sbncode/Calibration/TrackCaloSkimmer_module.cc
```diff
@@ -0,0 +1,460 @@
...
+    const std::vector<const recob::TrackHitMeta*> &trkHitMetas = fmtrkHits.isValid() ? fmtrkHits.data(trkPtr.key()) : emptyTHMVector;
+
+    float t0 = std::numeric_limits<float>::signaling_NaN();
+    if (fmT0.isValid() && fmT0.at(p_pfp.key()).size()) t0 = fmT0.at(p_pfp.key()).at(0)->Time();
+
+    if (fRequireT0 && fmT0.at(p_pfp.key()).size() == 0) {
+      continue;
+    }
+
+    std::cout << "Processing new track! ID: " << trkPtr->ID() << " time: " << t0 << std::endl;
```
wesketchum, line 191: Wrap in verbosity option.

sbncode/Calibration/TrackCaloSkimmer_module.cc
```diff
@@ -0,0 +1,460 @@
...
+    FillTrack(*trkPtr, pfp, t0, trkHits, trkHitMetas, calo, rawdigits, track_infos);
+  }
+}
+
+// helpers
+float HitMinTime(const std::vector<sbn::HitInfo> &hits, bool TPCE) {
+  double min = -1;
+
+  for (const sbn::HitInfo &h: hits) {
+    bool hit_is_TPCE = h.tpc <= 1;
```
wesketchum, line 202: <=1 ? So, either 0 or 1? Is that right?

sbncode/Calibration/TrackCaloSkimmer_module.cc
```diff
@@ -0,0 +1,460 @@
...
+    if (h.oncalo && hit_is_TPCE == TPCE) {
+      if (min < 0. || h.time < min) min = h.time;
+    } 
+  }
+
+  return min;
```
wesketchum, line 208: This always returns -1 if hit_is_TPCE is false. Is that the desired behavior?

sbncode/Calibration/TrackCaloSkimmer_module.cc
```diff
@@ -0,0 +1,460 @@
...
+  double max = -1;
+
+  for (const sbn::HitInfo &h: hits) {
+    bool hit_is_TPCE = h.tpc <= 1;
+    if (h.oncalo && hit_is_TPCE == TPCE) {
+      if (max < 0. || h.time > max) max = h.time;
+    } 
+  }
+
+  return max;
```
wesketchum, line 221: See comments on HitMinTime too.

sbncode/Calibration/TrackCaloSkimmer_module.cc
```diff
@@ -0,0 +1,460 @@
...
+    if ((track.End() - othr.start).r() < 50. || (track.End() - othr.end).r() < 50.) {
+      fTrack->tracks_near_end_dist.push_back(std::min((track.End() - othr.start).r(), (track.End() - othr.end).r())); 
+      fTrack->tracks_near_end_costh.push_back(
+        (track.End() - othr.start).r() < (track.End() - othr.end).r() ? 
+          track.EndDirection().Dot(othr.dir) : track.EndDirection().Dot(othr.enddir)); 
+    }
+  }
+
+  // Save?
+  bool select = false;
```
wesketchum, line 298: could alternatively default to true, and only not select if all of the selection tools return false.

sbncode/Calibration/TrackCaloSkimmer_module.cc
```diff
@@ -0,0 +1,460 @@
...
+  // Save!
+  if (select) {
+    std::cout << "Track Selected!\n";
+    fTree->Fill();
+  }
+}
+
+void sbn::TrackCaloSkimmer::DoTailFit() {
+  // Try fitting the constant and exponentials to the tail of dQ/dx v. RR on the collection plane
+  std::vector<double> fit_rr;
```
wesketchum, line 321: not sure how big these are, but it may be better to reserve for the total number of hits, so that the push_backs are a bit faster.

sbncode/Calibration/TrackCaloSkimmer_module.cc
```diff
@@ -0,0 +1,460 @@
...
+      fit_dqdx.push_back(h.dqdx);
+    }
+  }
+
+  // Copy the fit data to the global array
+  for (unsigned i = 0; i < fit_rr.size() && i < MAX_N_FIT_DATA; i++) {
+    FIT_RR[i] = fit_rr[i];
+    FIT_DQDX[i] = fit_dqdx[i];
+
+    std::cout << "FIT: " << FIT_RR[i] << " " << FIT_DQDX[i] << std::endl;
```
wesketchum, line 336: Wrap in verbosity option.

sbncode/Calibration/TrackCaloSkimmer_module.cc
```diff
@@ -0,0 +1,460 @@
...
+
+    double C = fFitConst.GetParameter(0);
+
+    double cresiduals = -1;
+    ConstResiduals(nparam, NULL, cresiduals, &C, 0);
+
+    fTrack->const_fit_C = C;    
+    fTrack->const_fit_residuals = cresiduals;
+
+    std::cout << "EXP FIT. A: " << A << " R: " << R << " resid: " << residuals << std::endl;
```
wesketchum, line 373: Wrap in verbosity option.

wesketchum (changes requested):
Larger suggestions:
* not clear if the tool is very ICARUS focused or not, but if it is could name it "ICARUS" somehow to not worry as much about use in SBND.
* I'm a little worried that the HitInfo object use may just bloat memory. Do you have any memory usage stats for this module from the memory tracker?
* Has this been checked on ICARUS data, with both west and east cryostats? If not, strongly suggest that is done before this is merged in.

sbncode/Calibration/TrackCaloSkimmer_module.cc
gputnam: For the way ICARUS data has been processed, there are art::Events that have missing data products (if the reconstruction doesn't get run). So this is required functionality for processing such files.

sbncode/Calibration/TrackCaloSkimmer_module.cc
gputnam: Yeah. I'd be happy to switch this out for a call to the Geometry service if such a functionality exists.

sbncode/Calibration/TrackCaloSkimmerSelectStoppingTrack_tool.cc
gputnam: Fixed.

sbncode/Calibration/TrackCaloSkimmerSelectStoppingTrack_tool.cc
gputnam: Fixed.

sbncode/Calibration/TrackCaloSkimmerSelectStoppingTrack_tool.cc
gputnam: Fixed.

sbncode/Calibration/TrackCaloSkimmerSelectStoppingTrack_tool.cc
gputnam: Fixed.

sbncode/Calibration/TrackCaloSkimmerSelectStoppingTrack_tool.cc
gputnam: Fixed.

sbncode/Calibration/TrackCaloSkimmer_module.cc
gputnam: I don't think I need to -- one shouldn't be fitting more than the last ~5cm of the track and so there's no case where you would need more than 30 points. Plus it's easier to make static with a fixed size.

sbncode/Calibration/TrackCaloSkimmer_module.cc
gputnam: When I looked around online this was the suggested way to work with TFitter. I could switch this out for a singleton class in principle, but the fit being performed is simple enough that this works for me.

sbncode/Calibration/TrackCaloSkimmer_module.cc
gputnam: Fixed.

sbncode/Calibration/TrackCaloSkimmer_module.cc
gputnam: Yeah -- the times are non-negative, so -1 is a sensible value to signify no such time exists.

sbncode/Calibration/TrackCaloSkimmer_module.cc
gputnam: same reply here

sbncode/Calibration/TrackCaloSkimmer_module.cc
gputnam: That's what the code currently does?

sbncode/Calibration/TrackCaloSkimmer_module.cc
gputnam: It shouldn't be more than 10-15 points.

sbncode/Calibration/TrackCaloSkimmer_module.cc
gputnam: Fixed.

sbncode/Calibration/TrackCaloSkimmer_module.cc
gputnam: Fixed.

gputnam:
> Larger suggestions:
> 
>     * not clear if the tool is very ICARUS focused or not, but if it is could name it "ICARUS" somehow to not worry as much about use in SBND.
> 
>     * I'm a little worried that the HitInfo object use may just bloat memory. Do you have any memory usage stats for this module from the memory tracker?
> 
>     * Has this been checked on ICARUS data, with both west and east cryostats? If not, strongly suggest that is done before this is merged in.

I've addressed the first round of comments! On these points -- I thought we wanted it usable in SBND? In any case, I've added a comment to specify its origin in ICARUS. 

The memory footprint of this module is not significant compared to the stage0/stage1 processing. Just did a test which returned <2GB of memory. 

And yes, it has been run in both cryostats on a large number of ICARUS data events.

sbncode/Calibration/TrackCaloSkimmer_module.cc
gputnam: On second glance -- I think I misunderstood the comment.

> This always returns -1 if hit_is_TPCE is false. 

This is not correct. If hit_is_TPCE is false, it only accounts for hits in the "West" TPC

sbncode/Calibration/TrackCaloSkimmer_module.cc
wesketchum: OK, I think this needs some followup with @SFBayLaser  and ICARUS team on cases where reco doesn't get run/missing data products. I've made little conditions/things like this in the past and it's always come back to bite me, as it inevitably hides problems.

If there's a valid reason that we won't have these objects on some events, ok. But reco failing probably means the whole job should fail and we should never get to this point.

sbncode/Calibration/TrackCaloSkimmer_module.cc
wesketchum: OK. Given it may be doing this frequently, I'd suggest that maybe this be part of the class and just cleared after use ... to avoid allocations in loops and the like. But, if performance is good and it's not worth it, it's fine.

sbncode/Calibration/TrackCaloSkimmer_module.cc
wesketchum: Yeah, but it took me a bit to convince myself of that. Modifying the code may make that more clear.

sbncode/Calibration/TrackCaloSkimmer_module.cc
wesketchum: OK ,I guess I just don't understand what this function is supposed to be doing then? Can you just briefly walk through that.

sbncode/Calibration/TrackCaloSkimmer_module.cc
wesketchum: OK, a little worried that the residual range fit is not configurable ... so I'd recommend at the very least making sure to throw an error if this 30 number is not enough? You may already have that in place, so sorry if you do and I missed it.

sbncode/Calibration/TrackCaloSkimmer_module.cc
wesketchum: OK, no strong preference from me here then.

wesketchum:
Thanks Gray! Resolved most, added a few comments. On the bigger ones, that all sounds good, just wanted to be sure.

sbncode/Calibration/TrackCaloSkimmer_module.cc
gputnam: This function returns the minimum hit time of all the hits on a track in one of the geographic TPC's -- either East or West.

sbncode/Calibration/TrackCaloSkimmer_module.cc
gputnam: It's just De Morgan (https://en.wikipedia.org/wiki/De_Morgan%27s_laws). Also -- one thing I want to do in the piece of code is to save which tool selected the track (so as to tell a cathode crosser from a stopping track, for example). That is most easily done this way.

sbncode/Calibration/TrackCaloSkimmer_module.cc
gputnam: I don't believe it means the reco is failing. From what I can tell, there is an optical filter that runs and the reconstruction only gets run if the optical filter passes. If the filter fails, then the raw data is still there but the reconstruction does not get run.

sbncode/Calibration/TrackCaloSkimmer_module.cc
gputnam: I think the performance overhead of this step is minimal.

sbncode/Calibration/TrackCaloSkimmer_module.cc
cjbackhouse: art filters are a total unintuitive pain. Way out of scope here, but I kind of think the reco should be putting an explicit "failed optical filter" flag in the event, and the downstream recos should be seeing that and, say, putting an empty list of tracks. Then downstream can distinguish between mysteriously missing products and the actual expected program flow.

wesketchum: Not sure I agree @cjbackhouse , but that's a good and different conversation.

@gputnam if that's the case, then the FCL should be configured such that the analysis module only runs on events that pass that filter, using the `SelectEvents: ["name_of_trigger_path"]` in the analysis module configuration (like we do for output streams). I would still remove this more silent failure here.

sbncode/Calibration/TrackCaloSkimmer_module.cc
wesketchum: OK, I got it. Just some documentation suggestions then: add a comment that 0 or 1 is 'TPC E' in ICARUS (note, may not be true for SBND...?) and that the function is using the bool TPCE input to say whether you want the east (true) or west (false) TPCs.

I'd actually strongly consider a 'HitMinTimeTPCEast' and 'HitMinTimeTPCWest' function that can call this one underneath just to avoid confusion, as it's really not intuitive.

sbncode/Calibration/TrackCaloSkimmer_module.cc
wesketchum: (see below comment on adding a comment to help document)

sbncode/Calibration/TrackCaloSkimmer_module.cc
wesketchum: OK, that's fine. Note if the goal is to know which tool passed, you stop as soon as the first tool passes, so ... you may want to consider a different way to do that.

sbncode/Calibration/TrackCaloSkimmer_module.cc
gputnam: Isn't that only possible to do if the analysis job is in the same larsoft job as the filtering? What if I run on the output of stage0/stage1 processing with the analysis module?

sbncode/Calibration/TrackCaloSkimmer_module.cc
gputnam: Added comments to document this behavior.

wesketchum (approved):

PR #130, opened 2021-06-22
FluxReader Addition

marcodeltutto:
This PR adds the FluxReader package to `sbncode`. This was developed by Zarko Pavlovic and I porting it to sbncode.

FluxReader allows reading gSimple or Dk2Nu files direclty into LArSoft. For each neutrino ray, an MCTruth dataproduct is created with associated MCFlux. In the FluxReader directory there are two LArSoft plugins:

- **FluxReader** (`art::Source`): reads each neutrino from the gSimple/Dk2Nu file, creates MCTruth, MCFlux (and Dk2Nu and NuChoice) dataproducts and places them in an art::Event.
- **FluxGeoFilter** (`art::Filter`): filters events based on the neutrino position. It moves along the neutrino direction, and only keeps events if the neutrino intersects at least one of the specified volumes. The default volume is volTPCActive.

Documentation on this code and instructions on how to run it are on the wiki: https://sbnsoftware.github.io/AnalysisInfrastructure/SimulationInfrastructure/FluxReader.


sbncode/FluxReader/job/fluxfilter.fcl
```diff
@@ -0,0 +1,12 @@
+# Configuration fhicl file for filter module FluxGeoFilter
+
+BEGIN_PROLOG
+flux_filter: {
+  module_type: "FluxGeoFilter"
+
+  # volWorld, volDetEnclosure, volCryostat, volTPC, volTPCActive, ...
+  volumes: [ "volTPCActive" ]
```
jzennamo, line 8: This only works for SBND, correct? Should we label it as such in the fcl name "_sbnd.fcl"?

jzennamo:
OK, it looks like there are some legacy fcls in here that point back to uboone. Could we remove those or move them to a "legacy" directory or something? Other than that this is very work and I am happy for it to be merged as is.

sbncode/FluxReader/job/fluxfilter.fcl
marcodeltutto: Yes, correct. Done with commits 60dcb63, fa717c8.

marcodeltutto:
Thanks for looking at it! I moved all the legacy fcls to a `legacy` directory, as you suggested.

jzennamo (commented):
aw, I think I did this wrong?

jzennamo (approved):
Looks great! Thanks Marco!

PR #131, opened 2021-06-24
BackTracker messages are useless

ikatza:
We have all experience the barrage of BackTracker messages...

```
%MSG
%MSG-w BackTracker:  MCParticleHitMatching:gaushitTruthMatch@BeginModule  24-Jun-2021 18:18:27 CDT run: 1 subRun: 0 event: 50
caught exception
---- BackTracker BEGIN
  No sim::SimChannel corresponding to channel: 8658
---- BackTracker END
```

This commits ignores them

### DIFF
--- a/sbncode/CAFMaker/cafmakerjob_sbnd.fcl
+++ b/sbncode/CAFMaker/cafmakerjob_sbnd.fcl
@@ -31,7 +31,7 @@ services:
   RandomNumberGenerator: {} #ART native random number generator
   TFileService: { fileName: "fullchain_production_hist_prod5.root" closeFileFast: false }
 #  scheduler:    { wantTracer: false wantSummary: true }
-  message:      { debugModules: ["*"] destinations: { debugmsg:{type: "cout" threshold: "INFO"  categories:{BackTracker: { limit: 0 }} } } }
+   message:     @local::sbnd_message_services_prod
   TimeTracker:  { printSummary: true }
   # #  @table::standard_services
 
--- a/sbncode/CAFMaker/cafmakerjob_sbnd.fcl
+++ b/sbncode/CAFMaker/cafmakerjob_sbnd.fcl
@@ -31,7 +31,7 @@ services:
   RandomNumberGenerator: {} #ART native random number generator
   TFileService: { fileName: "fullchain_production_hist_prod5.root" closeFileFast: false }
 #  scheduler:    { wantTracer: false wantSummary: true }
-  message:      { debugModules: ["*"] destinations: { debugmsg:{type: "cout" threshold: "INFO"} } }
+  message:      { debugModules: ["*"] destinations: { debugmsg:{type: "cout" threshold: "INFO"  categories:{BackTracker: { limit: 0 }} } } }
   TimeTracker:  { printSummary: true }
   # #  @table::standard_services
 


sbncode/CAFMaker/cafmakerjob_sbnd.fcl
```diff
@@ -31,7 +31,7 @@ services:
   RandomNumberGenerator: {} #ART native random number generator
   TFileService: { fileName: "fullchain_production_hist_prod5.root" closeFileFast: false }
 #  scheduler:    { wantTracer: false wantSummary: true }
-  message:      { debugModules: ["*"] destinations: { debugmsg:{type: "cout" threshold: "INFO"} } }
+  message:      { debugModules: ["*"] destinations: { debugmsg:{type: "cout" threshold: "INFO"  categories:{BackTracker: { limit: 0 }} } } }
```
PetrilloAtWork, line 34: Since we are here: `debugModules: ["*"]` has been neutered ages ago (see e.g. `lar --print-description message`). It can be removed.

ikatza:
Hi, I followed @etyley advice to instead use the default messaging service for SBND.

This makes it easier to maintain, and as such it's beneficial.

But to actually get rid of the barrage of BackTracker messages the default configuration should change https://github.com/SBNSoftware/sbndcode/pull/118

sbncode/CAFMaker/cafmakerjob_sbnd.fcl
ikatza: Not using `debugModules` anymore.

PetrilloAtWork (commented):
I hope I am not making up a crisis out of the understandable attempt to suppress a pointless warning...

What can I say? this design is broken, and it's not Iker's fault.
We are using now a message configuration which belongs to SBND, which is stored in `sbndcode`, which `sbncode` does not depend from and that is therefore not supposed to see.
There is a load of similar cases in this job. It's not only the configuration, either: for example, the very _geometry_ file is in `sbndcode`. So: this whole configuration file belongs to `sbndcode`.
Ok, I have unloaded that.

Now, more to the point. The change not only suppresses completely `BackTracker` messages (:+1:), but also it suppresses _all_ the `INFO` level messages: :+1:? :-1:?
I would elicit the comment from the maintainer of the configuration (@cjbackhouse?) on the topic.
My personal opinion: the configuration that Iker has selected (and that as he points out still does not suppress the warning message) was meant for production. That means that the job is supposed to have been validated, no human is routinely looking at the logs, and the only important output is error messages. In my mind, this type of output should be enabled for production only and by special override (`services.message: @local::sbnd_message_services_prod`), while the default one should be designed for interactive processing: see information level messages, check that everything looks ok. That would be `sbnd_message_services_interactive` and my suggestion would be to move to that.
Either way, the issue won't be solved until `message_stdout_warning_sbnd` in `messages_sbnd.fcl` is changed to suppress the evil warning.

cjbackhouse:
I don't think I am in charge of these fcls. If I was I would have a loooot to say.
Agree that we shouldn't suppress all info messages. It should be possible to specify the backtracker in particular in the message config. Though, maybe the real solution is to downgrade the messages (to debug?) in backtracker itself. If this is a normal thing that is expected to happen during normal processing, there's no real reason to emit a message at all.

PetrilloAtWork:
Ok, let's explore for a moment that possibility. @ikatza, do you understand what causes these messages, and whether this is indeed useless _in general_? If we don't have that understanding yet, we can do the proper investigation (I can do that), and if the outcome is that these messages are not due to the specifics of SBND but are a general nuisance, we can ask LArSoft to demote them to `DEBUG` as @cjbackhouse suggests. It will take slightly longer, but not _that much_ longer (especially compared to how longer _I_ am making this simple business just with my comments).

ikatza:
In my opinion, this fcl should on `sbndcode` and not in here. Maybe we could open an issue to discuss about moving this and other fcls down to `sbdcode` and `icaruscode`. But that's not the point of this PR.

My impression is that the standard production fcls use this messaging configuration. In my opinion `INFO` messages are too numerous to have them all displayed on the terminal, and I much prefer inspecting the `messages.log` file than the terminal output.

I do not understand what causes the warning, nor if it's serious or not. I assumed it isn't, given how we've been ignoring them.

@wforeman reached out to me with a workaround on the `AnalysisTree` module. I'll paste his message in here:

> There’s a workaround for this that’s in the AnalysisTree module, where you check if a hit has a SimChannel before calling the backtracker. It looks like this (line 2268):

```cpp
 if(DoesHitHaveSimChannel(fSimChannels,hitlist[i])) {
          std::vector<sim::TrackIDE> eveIDs = bt_serv->HitToTrackIDEs(clockData, hitlist[i]);
```

> The function itself is very simple:
```cpp
// Function to check if there was a SimChannel made for a hit (useful when checking for noise hits)
bool sbnd::AnalysisTree::DoesHitHaveSimChannel( std::vector<const sim::SimChannel*> chans, art::Ptr<recob::Hit> const& hit){
  bool isMatch = false;
  for(size_t sc = 0; sc < chans.size(); ++sc){
    if( chans[sc]->Channel() == hit->Channel() ) { isMatch = true; break; }
  }
  return isMatch;
}
```

I do not know if this workaround could be applied on the reco and cafmaker modules.

PetrilloAtWork:
Related: [Redmine issue #26010](https://cdcvs.fnal.gov/redmine/issues/26010), `larsim` pull request LArSoft/larsim#70.

ikatza:
The solution Gianluca linked as already been merged to `larsim` and I confirm that we no longer get the barrage of messages.

Thus, there's no point anymore for this PR. I'll close it now.

PR #133, opened 2021-06-28
FHiCL validation for beam information retrievers

PetrilloAtWork:
This is a mostly complete conversion of the configuration of the module into using FHiCL configuration validation.

In summary: the configuration description is centralised in the data structure `Config` and the constructor is using it (wrapped per FHiCL validation protocol). Where parameters were extracted (i.e., in the module constructor) now... parameters are still extracted, with a specific syntax.
In `Config`, each configuration parameter type and name is defined, together with a description (`Comment`) and the default value where provided. The most confusing part is the use of the same word for the parameter name (that is the key that we write in the configuration file, e.g. `Name{ "TimePadding" }` for `TimePadding`) and the C++ data member of `Config` that will contain the value of that parameter (also `TimePadding`). These do not need to be the same, but it has been a convenient, albeit possibly confusing, convention. In a single case they are different: `RawDataLabel` variable is still associated to a FHiCL key `raw_data_label`, that is, the FHiCL configuration is something like `raw_data_label: "TriggerUDPwhatever"`.
> One recommendation that escaped my notice in my first review (the one with fewer than 100 comments): the configuration parameter `raw_data_label` does not follow the convention of the other parameters, which would be `RawDataLabel`. I did not change it, but it should be fixed.

To see the resulting documentation, run `lar --print-description BNBRetriever` after having compiled the code.

Bonus track: I was compiling with Clang: an additional commit fixes the issue Clang found (inconsistent tagging of `override` methods).


PetrilloAtWork:
This is a proposal for @jzennamo's branch/pull request, and Joe is in charge for it.
He should be able to merge it himself if he thinks it worth.

PR #134, opened 2021-06-30
Remove GENIE v2 flag

mroda88:
While looking for something else, I stumbled upon some fcl files that still have GENIE v2 flags. They are still valid, but they should not be used as before.  
They are endemic and should be removed if found in general purpose fcl files. I had a look in the fcl file chain and these were the only one I found.


mroda88:
@miquelnebot this is a simple PR, to avoid bad habits.

mroda88:
Check if you see the notifications

miquelnebot (approved):

PR #136, opened 2021-07-02
Update razzle-dazzle Weights

etyley:
Update razzle-dazzle to point to weights files on sbnd_data


miquelnebot (approved):
new directory needed, OK

PR #140, opened 2021-07-07
Porting MultiPart event generator from icaruscode to sbncode

marcodeltutto:
This PR adds the MultiPart event generator that is in `icaruscode`.
See SBNSoftware/icaruscode#196


sbncode/EventGenerator/MultiPart/MultiPartRain_module.cc
```diff
@@ -0,0 +1,464 @@
...
+    size_t _multi_max;
+    
+    // verbosity flag
+    unsigned short _debug;
+};
+
+void MultiPartRain::abort(const std::string msg) const
+{
+    std::cerr << "\033[93m" << msg.c_str() << "\033[00m" << std::endl;
+    throw std::exception();
```
mroda88, line 106: I would prefer to avoid throwing empty exception as they are not very informative. but ok.

sbncode/EventGenerator/MultiPart/MultiPartVertex_module.cc
```diff
@@ -0,0 +1,499 @@
...
+  x = fFlatRandom->fire(_xrange[0],_xrange[1]);
+  y = fFlatRandom->fire(_yrange[0],_yrange[1]);
+  z = fFlatRandom->fire(_zrange[0],_zrange[1]);
+
+  if(_debug>0) {
+    std::cout << "Generating a rain particle at (" 
+	      << x << "," << y << "," << z << ")" << std::endl;
+  }
+}
+/*
```
mroda88, line 368: Commented code could go away. It creates confusion.

mroda88 (approved):
I understand this code import is kind of urgent. Since I didn't see anything major, I'm happy to approve. Although I have a few comments. A few just code related. 

I would like to add one idea for future developments. I'll write it here, in the hope it won't get lost. 
The approach taken by this code is ok, but it can be improved greatly to make the events look like neutrino events. 
The recipe is the following:

1. simply select randomly an energy from the neutrino spectrum
2. evaluate the energy in the centre of mass of an hypothetical interaction
3. then generate the particle in the CoM by simply adding a number of particle in a way similar to how it's done already but the kinematics is to be constructed using things like [TGenPhaseSpace](https://root.cern.ch/doc/master/classTGenPhaseSpace.html)
4. Boost back in the lab frame. 

This approach will give us events more similar to a realistic neutrino interaction. 
True, this will give us less control: this is why this approach should be in addition to what we have here.

henrylay97:
As mentioned the other day, these failures will continue until we update `sbnanaobj` to the new version of root, to match that of LArSoft. The failures are not related to this particular PR but we can't see its potential effect until the build errors are sorted.

marcodeltutto:
> I understand this code import is kind of urgent. Since I didn't see anything major, I'm happy to approve. Although I have a few comments. A few just code related.
> 
> I would like to add one idea for future developments. I'll write it here, in the hope it won't get lost.
> The approach taken by this code is ok, but it can be improved greatly to make the events look like neutrino events.
> The recipe is the following:
> 
> 1. simply select randomly an energy from the neutrino spectrum
> 2. evaluate the energy in the centre of mass of an hypothetical interaction
> 3. then generate the particle in the CoM by simply adding a number of particle in a way similar to how it's done already but the kinematics is to be constructed using things like [TGenPhaseSpace](https://root.cern.ch/doc/master/classTGenPhaseSpace.html)
> 4. Boost back in the lab frame.
> 
> This approach will give us events more similar to a realistic neutrino interaction.
> True, this will give us less control: this is why this approach should be in addition to what we have here.

Thanks @mroda88, we'll keep this in mind for future updates.

PR #141, opened 2021-07-10
Refactor configs and products out of sbncode for MeVPrtl Event Generator

gputnam:
Depends on https://github.com/SBNSoftware/sbnobj/pull/20 and https://github.com/SBNSoftware/icaruscode/pull/200.


mroda88 (approved):
I didn't run it myself, but there is nothing physics related here, just a a bit of refact. 
All looks good to me.

PR #144, opened 2021-07-14
Avoid crashing when there are no valid weights if StrictMode=false

gputnam:
Title.


cjbackhouse (approved):
I think this is fine. In general we should make sure to go through CAFMaker and ensure that StrictMode behaves the way it's supposed to everywhere.

PR #146, opened 2021-07-14
Fill Objects for MeVPrtl in StandardRecord

gputnam:
Title. Depends on https://github.com/SBNSoftware/sbnanaobj/pull/21 and https://github.com/SBNSoftware/sbncode/pull/141.

### DIFF
--- a/sbncode/CAFMaker/FillTrue.cxx
+++ b/sbncode/CAFMaker/FillTrue.cxx
@@ -109,6 +109,42 @@ namespace caf {
 
   }//FillSliceTruth
 
+
+ void FillMeVPrtlTruth(const evgen::ldm::MeVPrtlTruth &truth,
+                       caf::SRMeVPrtl &srtruth) {
+   // Fill stuff!!
+   srtruth.dcy.x = truth.decay_pos.X();
+   srtruth.dcy.y = truth.decay_pos.Y();
+   srtruth.dcy.z = truth.decay_pos.Z();
+   srtruth.dcyT  = truth.decay_pos.T();
+
+   srtruth.mom.x = truth.mevprtl_mom.X();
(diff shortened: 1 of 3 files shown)


gputnam:
In the ordering, https://github.com/SBNSoftware/sbncode/pull/141 should be merged in first and then we can point this one to develop.

gputnam:
Since https://github.com/SBNSoftware/sbncode/pull/141 is now merged I edited this to point to develop.

mroda88 (approved):
LGTM

sbncode/CAFMaker/FillTrue.cxx
```diff
@@ -109,6 +109,42 @@ namespace caf {
...
+   srtruth.ray_weight = truth.ray_weight;
+   srtruth.decay_weight = truth.decay_weight;
+   srtruth.C1 = truth.C1;
+   srtruth.C2 = truth.C2;
+   srtruth.C3 = truth.C3;
+   srtruth.C4 = truth.C4;
+   srtruth.C5 = truth.C5;
+
+   switch(truth.gen) {
+     case evgen::ldm::kDissonantHiggs:
```
mroda88, line 137: We need to change this following the changes in sbnanaobj.

mroda88 (changes requested):
I think now there is a problem of compatibility with https://github.com/SBNSoftware/sbnanaobj/pull/21#

sbncode/CAFMaker/FillTrue.cxx
gputnam: No actually the way I changed the code there's no change to sbncode. There is now an update to sbnana, which is in another PR.

mroda88 (approved):
Coo, we are good to go

PR #147, opened 2021-07-15
Update Calorimetry in CAFMaker

gputnam:
Don't save calibrated charge in CAFMaker Calorimetry. Also, depending on configuration save certain "Calo-Points".

As discussed in a previous SBN-Analaysis-Framework meeting, this implements per-calo-point saving on Track's in neutrino slices.

Depends on https://github.com/SBNSoftware/sbnanaobj/pull/22.


sbncode/CAFMaker/FillReco.cxx
```diff
@@ -384,19 +384,66 @@ namespace caf
...
+        p.t = dprop.ConvertXToTicks(xyz[i].x(), calo.PlaneID());
+
+        // lookup the wire -- the Calorimery object makes this
+        // __way__ harder than it should be
+        for (const art::Ptr<recob::Hit> &h: hits) {
+          if (h.key() == tps[i]) {
+            p.wire = h->WireID().Wire;
+            p.sumadc = h->SummedADC();
+            p.integral = h->Integral();
+          }
```
cjbackhouse, line 428: Should this search always succeed? What if it fails? We get the default-initialized values? Do we want to assert instead?

cjbackhouse (approved):
Minor comments

PR #148, opened 2021-07-15
Cleanup notebooks

gputnam:
This is my bad -- I accidentally included some figures from a prior PR with jupyter notebooks. This PR removes those plots.

It doesn't touch any cpp code.


miquelnebot (approved):
quick fix ok

PR #149, opened 2021-07-20
Updates to Track Skimming. Add in end-point-hit information and truth information

gputnam:
Depends on updates to the sbnobj repository: https://github.com/SBNSoftware/sbnobj/pull/22.

This adds in truth information and end-point-hit saving to the track calorimetry skimmer.

This PR has some urgency -- I would like it to be merged in for this week's release if possible so it can be included in a data processing campaign in ICARUS. 

I would ask @mrmooney to be the reviewer if he is available so that it can be reviewed promptly.

### DIFF
--- a/sbncode/CAFMaker/FillTrue.cxx
+++ b/sbncode/CAFMaker/FillTrue.cxx
@@ -7,11 +7,6 @@
 #include <algorithm>
 
 // helper function declarations
-caf::Wall_t GetWallCross( const geo::BoxBoundedGeo &volume,
-        const TVector3 p0,
-        const TVector3 p1);
-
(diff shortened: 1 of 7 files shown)


mrmooney (approved):
Looks fine to me, though please confirm that logic of Lines 595-600 in TrackCaloSkimmer_module.cc is intended (hinfo.p.x, y, z not initialized if no SpacePoint found to be associated with Hit) - are default x/y/z values desired in this case?

Not essential for now, but we may want to keep SpacePoints that have no associated collection plane Hit as it may help improve reconstruction of the Michel decay angle (or does our reconstruction require a collection plane Hit for a SpacePoint to be formed?)

PR #150, opened 2021-07-21
Add flux event weight

Ooohu:
Add a BNBFlux folder for flux uncertainty evaluation.

PR #151, opened 2021-07-22
Generalize CAFMaker to be able to accept multiple SystWeightLabels

cjbackhouse:
This is necessary to support both genieweight and fluxweight, which are computed in different modules.

The fluxweight modules are being added in https://github.com/SBNSoftware/sbncode/pull/150 but this change is independently useful.

I haven't been able to test properly yet. There is only a very small/trivial fluxweight file available, and I am struggling with CAFMaker fcl problems, and something to do with incompatible geometries. It would be good to generate, say, 1000 events with all the genie and flux weights turned on to make sure the resulting CAFs look sane.


cjbackhouse:
I was able to check that this runs properly over a file Keng supplied that had both genieweight and fluxweight objects in it.

Switching review request to someone who probably has more bandwidth available than Andy.

sbncode/CAFMaker/CAFMaker_module.cc
```diff
@@ -783,10 +786,16 @@ void CAFMaker::produce(art::Event& evt) noexcept {
     // corresponding to a neutrino) that could plausibly be reweighted. This
     // avoids the need for special configuration for cosmics or single particle
     // simulation, and real data.
-    if(!fmpewm && mctruth->NeutrinoSet()) fmpewm = FindManyPStrict<sbn::evwgh::EventWeightMap>(mctruths, evt, fParams.SystWeightLabel());
+    if(fmpewm.empty() && mctruth->NeutrinoSet()){
```
gputnam, line 789: Should this check be here? There might be systematic uncertainties for generators other than neutrinos. 

At least, this may be something to revisit in the future.

sbncode/CAFMaker/CAFMaker_module.cc
```diff
@@ -799,10 +808,10 @@ void CAFMaker::produce(art::Event& evt) noexcept {
           }
           const unsigned int idx = fWeightPSetIndex[it.first];
           if(idx >= srneutrinos.back().wgt.size()) srneutrinos.back().wgt.resize(idx+1);
-	  srneutrinos.back().wgt[idx].univ = it.second;
+          srneutrinos.back().wgt[idx].univ = it.second;
```
gputnam, line 811: Filling the SR objects really shouldn't go in CAFMaker_module itself -- it should be split out into one of the functions inside FillTrue.

sbncode/CAFMaker/CAFMaker_module.cc
cjbackhouse: This is like this due to previous complaints about errors running CAFMaker over non-neutrino MC with the reweight label still set. I think we actually need to have a proper conversation about what the convention is for set/unset product labels and existing/missing products before we can know what the right answer is here.

sbncode/CAFMaker/CAFMaker_module.cc
cjbackhouse: I pushed a couple of commits moving things to FillTrue

cjbackhouse:
Conflict fixed.

sbncode/CAFMaker/CAFMaker_module.cc
gputnam: Ok! In any case this can be deferred for longer term I think.

gputnam:
g2g from my perspective.

PR #152, opened 2021-07-28
Simple-Flash Match: multi-matcher, unbiased, tuneable, fast.

ikatza:
Substantial improvements to discern in-time and out-of-time interactions.

Particularly, interactions occurring on the far half side of the TPCs will be much better matched. Also now _near time_ interactions, eg cosmics that may be synchronous to the beam, will be time-tagged and match.

- Multi-flash matching, allowing many to many matching. At a minimal cost of processing time.
- Completely decoupled the charge and flash computations.
- New fcl parameters to fine tune the module to improve performance, or to aid with commissioning tasks.
- In particular more control to filter small depositions of charge and light
- Enforce flash and charge concurrence
- Metrics are now created with non split slices
- Substantial improvements in the reconstruction of the flash centre: (flash_x, flash_y, flash_z)
- ..... Improved ratio metric construction for ICARUS
- ..... Robuster method to compute `hypo_x_rr` and `hypo_x_ratio`, both show less bias
- ..... Use skew to unbias `flash_y` and `flash_z`, not yet optimised

More details on:
https://sbn-docdb.fnal.gov/cgi-bin/sso/ShowDocument?docid=22643

This is the main PR. `sbndcode, icaruscode, sbnd_data, icarus_data` will be updated too, they all depend on this PR.

### DIFF
--- a/sbncode/FlashMatch/CMakeLists.txt
+++ b/sbncode/FlashMatch/CMakeLists.txt
@@ -31,6 +31,8 @@ art_make(
 
   larpandora_LArPandoraInterface
 
+  # caf_RecoUtils //uncomment for creating metrics
+  # nusimdata_SimulationBase //uncomment for creating metrics
   ${MF_MESSAGELOGGER}
   ${FHICLCPP}
   ${ROOT_GEOM}
--- a/sbncode/FlashMatch/CMakeLists.txt
+++ b/sbncode/FlashMatch/CMakeLists.txt
@@ -30,9 +30,9 @@ art_make(
   lardataobj_Simulation
 
   larpandora_LArPandoraInterface
+  larsim_Utils
 
-  # caf_RecoUtils //uncomment for creating metrics
-  # nusimdata_SimulationBase //uncomment for creating metrics
+  nusimdata_SimulationBase
   ${MF_MESSAGELOGGER}
   ${FHICLCPP}
   ${ROOT_GEOM}
--- a/sbncode/FlashMatch/FlashPredict.hh
+++ b/sbncode/FlashMatch/FlashPredict.hh
@@ -33,22 +33,19 @@
 #include "lardata/DetectorInfoServices/DetectorClocksServiceStandard.h"
 #include "lardata/Utilities/AssociationUtil.h"
 #include "lardataobj/RecoBase/PFParticle.h"
-#include "lardataobj/RecoBase/Track.h"
-#include "lardataobj/RecoBase/Vertex.h"
+// #include "lardataobj/RecoBase/Track.h"
 #include "lardataobj/RecoBase/PFParticleMetadata.h"
 #include "lardataobj/RecoBase/Hit.h"
 #include "lardataobj/RecoBase/SpacePoint.h"
 // #include "lardataobj/RecoBase/Slice.h"
 #include "lardataobj/RecoBase/OpHit.h"
-#include "lardataobj/RecoBase/OpFlash.h"
-#include "lardataobj/AnalysisBase/T0.h"
+// #include "lardataobj/RecoBase/OpFlash.h"
 #include "larpandora/LArPandoraInterface/LArPandoraHelper.h"
 
 #include "TTree.h"
 #include "TF1.h"
 #include "TFile.h"
 #include "TH1.h"
-#include "TSpline.h"
 
 #include "sbncode/OpT0Finder/flashmatch/Base/OpT0FinderTypes.h"
 #include "sbncode/OpDet/PDMapAlg.h"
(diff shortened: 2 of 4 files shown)


sbncode/FlashMatch/CMakeLists.txt
```diff
@@ -31,6 +31,8 @@ art_make(
 
   larpandora_LArPandoraInterface
 
+  # caf_RecoUtils //uncomment for creating metrics
```
etyley, line 34: I don't think that having to comment and recompile the code in order to be able to create the metrics is a great idea, it massively increases the maintenance burden if there are any upstream changes to the simulation or reconstruction. It would be preferable to have just behind a fcl parameter

sbncode/FlashMatch/FlashPredict.hh
```diff
@@ -83,59 +79,207 @@ public:
...
                         const std::set<unsigned>& tpcWithHits) const;
   unsigned sbndPDinTPC(const int pdChannel) const;
   unsigned icarusPDinTPC(const int pdChannel) const;
+  double wallXWithMaxPE(const OpHitIt opH_beg,
+                        const OpHitIt opH_end) const;
+  std::list<double> wiresXGl() const;
+  double driftDistance() const;
   double flashXGl(const double hypo_x, const double flash_x) const;
-  double driftDistance(const double x) const;
+  double foldXGl(const double x_gl) const;
```
etyley, line 274: This name is pretty cryptic to me, some comments to explain what these functions do a bit more verbosely would help

sbncode/FlashMatch/FlashPredict.hh
```diff
@@ -148,87 +292,74 @@ private:
   const double fBeamWindowStart, fBeamWindowEnd;
   const double fFlashStart, fFlashEnd;
   const unsigned fTimeBins;
-  const bool fSelectNeutrino, fOnlyPrimaries;
+  const bool fSelectNeutrino;
+  const bool fOnlyCollectionWires;
+  const bool fForceConcurrence;
```
etyley, line 297: Same thing here, some comments to explain what `ForceConcurrence` means would make the code easier to understand.

sbncode/FlashMatch/FlashPredict.hh
```diff
@@ -148,87 +292,74 @@ private:
...
   unsigned fTPCPerDriftVolume;
   const unsigned fOpDetNormalizer;
   const double fTermThreshold;
-  std::list<double> fWiresX_gl;
-  // std::vector<double> fPMTChannelCorrection;
 
-  unsigned fPeakCounter = 0;
-  std::vector<recob::OpHit>::iterator fOpH_beg, fOpH_end;
-  const art::ServiceHandle<geo::Geometry> geometry;
+  const unsigned kRght = 0;
```
etyley, line 329: If these are known at compile time, is there a reason not to use a `constexpr`

sbncode/FlashMatch/FlashPredict.hh
```diff
@@ -148,87 +292,74 @@ private:
...
   };
-  std::array<Fits, 3> rrFits;
-  std::array<Fits, 3> peFits;
-  const std::array<std::string, 3> suffixes{"l", "h", "m"};
-  // const std::string kPolFit = "pol3";
+  TH2D* fRRH2; TH2D* fRatioH2;
+  const unsigned kMinEntriesInProjection = 100;
+  std::array<Fits, 3> fRRFits;
+  std::array<Fits, 3> fRatioFits;
+  const std::array<std::string, 3> kSuffixes{"l", "h", "m"};// low, high, medium
```
etyley, line 346: If these are known at compile time, is there a reason not to use a `constexpr`

sbncode/FlashMatch/FlashPredict_module.cc
```diff
@@ -25,64 +25,94 @@ FlashPredict::FlashPredict(fhicl::ParameterSet const& p)
...
 
-  fPDMapAlgPtr = art::make_tool<opdet::PDMapAlg>(p.get<fhicl::ParameterSet>("PDMapAlg"));
-  fGeoCryo = std::make_unique<geo::CryostatGeo>(geometry->Cryostat(fCryostat));
-  fNTPC = geometry->NTPC();
-  for (size_t t = 0; t < fNTPC; t++) {
-    const geo::TPCGeo& tpcg = fGeoCryo->TPC(t);
-    fWiresX_gl.push_back(tpcg.LastPlane().GetCenter().X());
+  // TODO: check params are sane:
+  if(fFlashStart > 0. || fFlashEnd < 0.){
+    throw cet::exception("FlashPredict");
```
etyley, line 70: Adding a description explaining why the exception was thrown could save a lot of debugging later down the line.

sbncode/FlashMatch/FlashPredict_module.cc
```diff
@@ -25,64 +25,94 @@ FlashPredict::FlashPredict(fhicl::ParameterSet const& p)
...
+
+  if(fSBND && !fICARUS) {
+    if(fUseOppVolMetric) {
+      throw cet::exception("FlashPredict")
+        << "UseOppVolMetric: " << std::boolalpha << fUseOppVolMetric << "\n"
+        << "Not supported on SBND. Stopping.";
+    }
     fTPCPerDriftVolume = 1;
     fDriftVolumes = fNTPC/fTPCPerDriftVolume;
+    // TODO: make these few fcl params
```
etyley, line 87: Status of this `TODO`, I think this would be better if these should be fcl params.

sbncode/FlashMatch/FlashPredict_module.cc
```diff
@@ -566,113 +558,127 @@ void FlashPredict::loadMetrics()
...
   mf::LogInfo("FlashPredict") << "Finish loading metrics";
 }
 
 
-bool FlashPredict::computeChargeMetrics(const flashmatch::QCluster_t& qClusters)
+double FlashPredict::cheatMCT0(const std::vector<art::Ptr<recob::Hit>>& hits,
+                               const std::vector<art::Ptr<simb::MCParticle>>& mcParticles)
+{
+  auto clockData = art::ServiceHandle<detinfo::DetectorClocksService const>()->DataForJob();
+  const std::vector<std::pair<int, float>> matches =
```
etyley, line 630: The truth match utils within LArSoft make this much simpler: https://github.com/LArSoft/larsim/blob/33c27d3378360a0b9fbb134263ac3e45221b58bc/larsim/Utils/TruthMatchUtils.h#L52-L62

etyley (changes requested):
I have added some general comments on coding conventions but somebody from the PDS group should probably check the actual physics of the code. I strongly feel that un-commenting code to create the metrics is not the best way forwards and this should only require a fcl parameter switch.

sbncode/FlashMatch/CMakeLists.txt
ikatza: That's a good point. I've removed all of uncomments lines, now there's a new fcl specific to create metrics.

sbncode/FlashMatch/FlashPredict.hh
ikatza: Good idea.

sbncode/FlashMatch/FlashPredict.hh
ikatza: Turns out C++ has a thing against making std::string constexpr. Whilst there are ways around it, there's really no point. I've stopped using the method that called for `kSuffixes`.

sbncode/FlashMatch/FlashPredict_module.cc
ikatza: Forgot about these!

Made commits to the specific repos with updated fcls.

sbncode/FlashMatch/FlashPredict_module.cc
ikatza: Very good advise, thanks I didn't know about it.

sbncode/FlashMatch/FlashPredict_module.cc
ikatza: Done.

sbncode/FlashMatch/FlashPredict.hh
ikatza: `ForceConcurrence` is a tad too complex to explain on an inline comment. Still I put one elsewhere.

ikatza:
I've addressed @etyley requests. Except for putting more comments. 

I agree is important but I would rather postpone that to have the time to do it properly and following whatever convention we agree on (I guess oxygen style?). At this point is more work than what I'm able to put into this and I would like use these new updates for an analysis.

etyley (approved):
These changes look good to me, though I still think this could do with another review from someone more familiar with the code

jzettle (approved):
I don't see anything pressing to add in a readthrough, although I am by no means an expert in the details of the code here.

ikatza:
Is it possible to run the CI with the other pull requests?
Otherwise it will not run.

Talking about: 
https://github.com/SBNSoftware/sbndcode/pull/129
https://github.com/SBNSoftware/icaruscode/pull/213

PR #154, opened 2021-08-02
move experiment genie_xsec dependency to sbncode

### LINKED ISSUE #15: Move dependency on GENIE cross section package from experiment repositories to sbncode
Both experiments depend on GENIE cross section, and there is value in having them share the same version of those.
Technically, the dependency can be moved into `sbncode`.

miquelnebot:
This is a request to move experiment genie_xsec dependency to sbncode.
Open issue details:
https://github.com/SBNSoftware/sbncode/issues/15#issuecomment-889241974
Closes #15 
ExperimentPRs:
https://github.com/SBNSoftware/sbndcode/pull/131
https://github.com/SBNSoftware/icaruscode/pull/222

### DIFF
--- a/ups/product_deps
+++ b/ups/product_deps
@@ -38,7 +38,7 @@ larsoft             v09_27_00
 sbnobj              v09_11_15_01
 sbnanaobj           v09_16_12
 sbndaq_artdaq_core  v0_07_08_of0
-genie_xsec          v3_00_04a
+genie_xsec          v3_00_06
 
 # list products required ONLY for the build
 # any products here must NOT have qualifiers
--- a/ups/product_deps
+++ b/ups/product_deps
@@ -38,7 +38,7 @@ larsoft             v09_27_00
 sbnobj              v09_11_15_01
 sbnanaobj           v09_16_12
 sbndaq_artdaq_core  v0_07_08_of0
-genie_xsec          v3_00_06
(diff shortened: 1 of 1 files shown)


marcodeltutto (approved):
Thanks Miquel!

ups/product_deps
```diff
@@ -38,6 +38,7 @@ larsoft             v09_27_00
 sbnobj              v09_11_15_01
 sbnanaobj           v09_16_12
 sbndaq_artdaq_core  v0_07_08_of0
+genie_xsec          v3_00_04a
```
mroda88, line 41: While we are at it, we should update the dependency to the latest version. v3_00_06

mroda88 (changes requested):
Can we give it a try with v3_00_06? The spline's dependencies are ok.

mroda88 (changes requested):
I'm sorry my previous message was a little too early. There are not splines v3_00_06. The original commit was ok

mroda88 (approved):
Perfect!

wesketchum (approved):
Looks good to me, though obviously need to make sure CI tests are ok.

jzennamo:
@miquelnebot @marcodeltutto @mroda88 Just wanting to check on this, are these 2 CI failures expected? It looks like there are a slightly different number of hits and the code is a bit faster.

PR #156, opened 2021-08-03
Updates to Configurations for ICARUS MCP

gputnam:
This PR is organizing updates to CAFMaking configurations for use in ICARUS for the upcoming MC production.

This PR also updates the organization of the CAFMaker fcl files. It moves the configurations to fcl/caf. The configurations to run in SBND are in fcl/caf/sbnd. The configurations to run in ICARUS are moved into icaruscode in a sibling PR.

This so far is a draft PR to collect updates needed for the MCP.

_[additional irrelevant text was edited away]_


wesketchum:
@grayputnam , do you think this will get in for this upcoming production (finished in 6 days...), or push to sometime after?

gputnam:
Hey Wes! My hope that is that it can be pushed in the next 6 days.

jzennamo:
Hi @gputnam, it looks like this PR is deleting the fcl that was provided to us for running CAFs in ICARUS (i.e. `cafmakerjob_icarus_sce.fcl `), should we plan to use a different fcl?

gputnam:
Hey! It actually moves that configuration to icaruscode with the sibling PR: https://github.com/SBNSoftware/icaruscode/pull/223. So the necessary configurations are all there!

gputnam:
Ok I need to do a couple more tests tomorrow but this should be ready for review! I will mark it as not a draft.

sbncode/CAFMaker/FillTrue.cxx
```diff
@@ -310,6 +310,7 @@ namespace caf {
       srneutrino.E = nu.Nu().EndMomentum().Energy();
       srneutrino.momentum = nu.Nu().EndMomentum().Vect();
       srneutrino.position = nu.Nu().Position().Vect();
+      srneutrino.time = nu.Nu().Position().T() / 1000. /* ns -> us */;
```
wesketchum, line 313: I think this goes against the recent unit consistency efforts to keep things in units of GeV/cm/ns where reasonable. @PetrilloAtWork, @cjbackhouse , or @etyley could maybe reply.

sbncode/EventGenerator/MeVPrtl/Tools/IMeVPrtlDecay.h
```diff
@@ -46,7 +46,9 @@ class IMeVPrtlDecay: virtual public IMeVPrtlStage
 protected:
     double TimeOfFlight(const MeVPrtlFlux &flux, TVector3 decay) {
       // TODO: should the neutrino TOF be subtracted here to get the correct T0?
-      return flux.pos.T() + (flux.pos.Vect() - decay).Mag() * (1. / flux.mom.Beta()) / Constants::Instance().c_cm_per_ns;
+      //
+      // TODO: implement time of flight
```
wesketchum, line 50: Is this undoing a TOF correction? I think you're the expert here so I defer to you, just wondered about this choice.

sbncode/EventGenerator/MeVPrtl/config/numi_kaon_common.fcl
```diff
@@ -11,6 +11,10 @@ icarus_numi_beam2detector_rotation: [   0.921228671,    0.0226872648,   0.388359
 
 icarus_numi_beam_origin: [0, 0, 0, 4.8195e2, 79.5800e2, 794.3955e2] # cm
```
wesketchum, line 12: Please see https://github.com/SBNSoftware/icaruscode/pull/230 and make sure this is consistent? (It looks like it's not). 

This and other geometry pieces look like they'd be better pulled from the geometry service to not need to worry about config in every place? Is there a reason not to do this? If not, I think it's sufficient for now to check the geometry is consistent, but we should make an issue to change this for the next major release.

sbncode/CAFMaker/FillTrue.cxx
etyley: The timing is actually done in us rather than ns (This was more a documentation of what was currently in use rather than a deliberate decision): https://github.com/SBNSoftware/sbnanaobj/blob/feature/etyley_CAFUnits/sbnanaobj/StandardRecord/README.md

wesketchum: Thank you! apologies for the noise.

mroda88 (commented):
Hi @grayputnam can we just have a comments or two on why the changes in the generation of a proper random seed are requiring changes in the MeV Portal code? I agree with @wesketchum that the changes look like something is still in progress here.

sbncode/EventGenerator/MeVPrtl/Tools/IMeVPrtlDecay.h
gputnam: It is. I had some code to try to implement the TOF of the (massive) LDM. But I decided I didn't trust it so for this production I turned it off. Planning to implement it for the futurue.

sbncode/EventGenerator/MeVPrtl/config/numi_kaon_common.fcl
gputnam: hmmm interesting it looks like there's been another update. I'll keep it in mind for the October production, not going to worry about it for now.

As far as if I can pull this from the Geometry service -- I'd love to! But I don't see how. There's now volume definition that matches the ICARUS AV (this is exactly the problem w/ "both cryostats" in GENIE). And the configuration for the beam position you linked to is XML not FCL so I don't see how I can automatically include it.

gputnam:
@mroda88 the changes to MeVPrtl are orthogonal to the random number generation. That comment was made by Gianluca about one of the aspects of this PR -- but I think he edited my comment in a way that made that unclear. I have since fixed his update to clarify that.

sbncode/EventGenerator/MeVPrtl/config/numi_kaon_common.fcl
PetrilloAtWork: Indeed `Geometry` (as the service) does not know where the targets are.
The transformation matrix is also not known to LArSoft: it's passed straight to `GENIEHelper`.
I don't see any easy way to get it from LArSoft; a module might be written specifically for that purpose, though.

sbncode/EventGenerator/MeVPrtl/config/numi_kaon_common.fcl
```diff
@@ -11,6 +11,10 @@ icarus_numi_beam2detector_rotation: [   0.921228671,    0.0226872648,   0.388359
...
 
+# icarus_detector_box: [-364.49, 364.49, -173.41, 143.41, -909.951, 879.951]
+# icarus_detector_box: [-369.63. 369.63, -224.61, 177.71, -994.950652270838, 994.950652270838]
+icarus_detector_box: [-378.49, 378.49, -191.86, 144.96, -904.950652270838, 904.950652270838]
```
PetrilloAtWork, line 16: Now this is something that `Geometry` knows.
I _recommend_ to try to figure out how to get this from LArSoft for the future.
@grayputnam: is this anything that has already been discussed? if not, please open an issue about it.

P.S. I really appreciate the centralisation of these coordinates into a single point.

sbncode/LArRecoProducer/transfer_flashmatch_producer.fcl
```diff
@@ -5,4 +5,19 @@ transfer_flashmatch_sce_sbnd: {
...
+  FlashLabel: fmatchCryoE
+  PFParticleLabelOut: pandoraGausSCECryoE
+}
+
+transfer_flashmatch_sce_icarus_cryoW: {
+  module_type: TransferPFParticleFlashMatch
+  PFParticleLabelIn: pandoraGausCryoW
+  FlashLabel: fmatchCryoW
+  PFParticleLabelOut: pandoraGausSCECryoW
+}
```
PetrilloAtWork, line 21: This also looks like something belonging to `icaruscode`.

PetrilloAtWork (approved):
There are a couple of action items for the future:
* move ICARUS configuration into `icaruscode` (part of this has already happened, but there are detector-specific pieces still around);
* extract detector coordinates from the geometry service rather than encoding them in the configuration.

PetrilloAtWork:
> @mroda88 the changes to MeVPrtl are orthogonal to the random number generation. That comment was made by Gianluca about one of the aspects of this PR -- but I think he edited my comment in a way that made that unclear. I have since fixed his update to clarify that.

Aaargh! I apologise: that description was meant to be in the [`icaruscode` side of this pull request](https://github.com/SBNSoftware/icaruscode/pull/223)... I can see why it is confusing people!
I have moved to where it belongs.

cjbackhouse (commented):
Not really an expert here, but looks broadly fine.

PR #157, opened 2021-08-03
Improve bookeeping of not-candidate slices

ikatza:
Now my module creates associations for every slice, and not just the ones that are potential neutrino candidates.

There will be warnings related to the size of the assns vector, as now there will be more assns than before. The size should be the same as the `recob::Slice` vector.

### DIFF
--- a/sbncode/FlashMatch/FlashPredict.hh
+++ b/sbncode/FlashMatch/FlashPredict.hh
@@ -369,6 +369,7 @@ private:
   static constexpr int kQNoOpHScr = -1;
   static constexpr int kNoChrgScr = -2;
   static constexpr int k0VUVPEScr = -3;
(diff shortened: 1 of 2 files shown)


ikatza:
This PR is meant to fix potential inconsistencies with the CI system.

As long as we get the expected warning (because the product size will grow for most cases) I believe we can skip the code review.

pgreen135:
trigger build

ikatza:
As expected, one of the events have now 17 entries on the assns vector (same as the number of slices).

The other event already had 14 entries, just like the number of slices.

henrylay97 (approved):
Looks generally sensible to me from a brief look through the code. CI outputs look as expected now, and its good to see there are no differences from the event that was flagged last time.

PR #158, opened 2021-08-04
Add missing definition for fmatch

ikatza:
The `cafmakerjob_sbnd_fmatch.fcl` was missing the definition of `fmatch`. I suspect this was on `cafmakerjob_sbnd.fcl` but got cleaned away, as it shouldn't have been there since `sbndcode` does run `fmatch` by default.

No changes needed for the `cafmaker_icarus` side of things.


ikatza:
Can you please review it @etyley? I believe this is similar to a change you made recently.

etyley (approved):
All looks good to me

PR #159, opened 2021-08-05
feature/caf_plane_vectors

cjbackhouse:
Depends on https://github.com/SBNSoftware/sbnanaobj/pull/23
Adjust the filling of the variables in CAFMaker


cjbackhouse:
This is based on the units branch. The diff should be saner once that is merged.

etyley (approved):
This all looks good to me. This also implements the change of shower energy units from MeV to GeV, deprecating https://github.com/SBNSoftware/sbncode/pull/120, so I will close that PR.

cjbackhouse:
I think it makes more sense to merge that first, and then this

etyley:
Okay, I will leave it up then (I think it is waiting for your approval still)

cjbackhouse:
> Okay, I will leave it up then (I think it is waiting for your approval still)

Oops. Done.

gputnam:
Hi all -- following up on conversation over slack.

I disagree with the idea behind this change to the StandardRecord object. I think that in practice in analysis code it is easier to access per-plane variables (PID, e.g.) as separate elements in the C++ class rather than as a list. 

I recognize that this update simplifies the code in CAFMaker, but that is just because per-plane objects are stored in larsoft as vectors. Once we pull them out of vectors into separate elements, it is easier for analysis to use that modality. And we should optimize for simplicity of analysis code, not for simplicity in CAFMaker.

etyley:
I think that this actually simplifies the analysis code as it makes it easier to access the bestplane'th object, for example here: https://github.com/SBNSoftware/sbnana/pull/41/files#diff-ac9316eb3ece875355f4ddeb844af3609be711583738c0d3e645d0d70bd66d36L274-R274

gputnam:
Well... a few things here. 

First, part of the simplification in the sbnana code is that the new code does not do any error checking on "bestplane". If I am reading the code correctly I believe that is a bug -- and that it will crash in the cases where bestplane is -1.

Second, we may be encountering a difference in analysis between C++ and python. In the pythonic/columnar method, I want the structure to be as "flat" as possible so that it can be read into a table. But it is also my understanding that we want StandardRecord to be as flat as possible. Making these variables into lists adds in a layer of indirection that is not necessary and goes against this principle.

Third, I think putting these variables into lists sends the wrong message to analyzers about how to treat them -- that (for example) the PID on each plane is the same basic thing, and that (for example) it is ok to sum the chi2's on each plane to get a total PID. When in reality the PID on different planes often tell you different things and it is often best to just use one of them.

One thing I notice about this PR is that the shower/track per-plane vars are structured differently. In the Track case, the per-plane variables are separated into objects (SRTrackCalo and SRTrkChi2PID) that are repeated per plane. In the shower case, each individual variable is saved as a float which are all repeated per plane. In my experience I have found that the Track structure works well. I agree that the shower case seems clunky, but I would advocate that the better change is to group those per-plane float's into a single object which is repeated per-plane.

cjbackhouse:
> First, part of the simplification in the sbnana code is that the new code does not do any error checking on "bestplane". If I am reading the code correctly I believe that is a bug -- and that it will crash in the cases where bestplane is -1.

I hadn't realized it was possible for that to be unset. If it is, it's still a one line early return to deal with that.

> Second, we may be encountering a difference in analysis between C++ and python. In the pythonic/columnar method, I want the structure to be as "flat" as possible so that it can be read into a table. But it is also my understanding that we want StandardRecord to be as flat as possible. Making these variables into lists adds in a layer of indirection that is not necessary and goes against this principle.

As you can see in the other PR, this structure is completely transparent from the C++ side. And I don't think there is any real performance implication. The values are still consecutive in memory, just in a different tree.

> Third, I think putting these variables into lists sends the wrong message to analyzers about how to treat them -- that (for example) the PID on each plane is the same basic thing, and that (for example) it is ok to sum the chi2's on each plane to get a total PID. When in reality the PID on different planes often tell you different things and it is often best to just use one of them.

I don't think you should ever sum - but isn't averaging, or indexing by bestplane, or looping over all three planes trying to satisfy some criterion, far more common than wanting to access a specific plane? I can't think of any circumstance where you would uniformly prefer U over V or vice versa.

> One thing I notice about this PR is that the shower/track per-plane vars are structured differently. In the Track case, the per-plane variables are separated into objects (SRTrackCalo and SRTrkChi2PID) that are repeated per plane. In the shower case, each individual variable is saved as a float which are all repeated per plane. In my experience I have found that the Track structure works well. I agree that the shower case seems clunky, but I would advocate that the better change is to group those per-plane float's into a single object which is repeated per-plane.

Agree about array-of-structs vs struct-of-arrays. We could/should easily do that.

gputnam:
I think we should setup a zoom chat to discuss this -- progress on a resolution is going to be slow over github. Is there a time that works for you (and any other interested parties)?

cjbackhouse:
I am broadly free the rest of this week. @etyley? Who else?

etyley:
I am also broadly free apart from the usual sbn meetings

cjbackhouse:
Discussed at today's sbn ana infra meeting. Agreed that at the least Gray needs to check that things technically work for him, before we can make any design decision. And also that this is likely to miss the boat for the current production, and that's ok.

gputnam:
Following up on a discussion the @cjbackhouse and I had over slack -- I am happy with this PR as long as it happens alongside a change in the way array's are flattened in the flatmaker script. This PR can be merged alongside a change to the upstream repository that contains flatmaker.

cjbackhouse:
Specifically that would show up here as just a bump of the `srproxy` version. I'm swamped though, so it's not clear when I'll get to that.

wesketchum:
Thanks guys. Was gonna ping for an update, but the above sounds good.

I've cut this week's release already, so natural next point of insertion would be next week. Adding 'external' tag to remind us we need the srproxy update.

henrylay97:
As requested, the manual trigger results @miquelnebot 

Both succeeded with warnings at build stage (nothing to worry about, the usual version mismatch with larsoft) 
The full ci_test results can be found here for [c7:prof](https://dbweb8.fnal.gov:8443/LarCI/app/ns:sbnd/build_detail/phase_details?build_id=sbnd_ci/7391&platform=Linux%203.10.0-1160.31.1.el7.x86_64&phase=ci_tests&buildtype=slf7%20c7:prof) and [e20:prof](https://dbweb8.fnal.gov:8443/LarCI/app/ns:sbnd/build_detail/phase_details?build_id=sbnd_ci/7390&platform=Linux%203.10.0-1160.31.1.el7.x86_64&phase=ci_tests&buildtype=slf7%20e20:prof).

The warning for reco2 is due to SBNSoftware/sbndcode#187 which was merged this morning and for which the reference files have not been updated.

TL;DR this PR is good to go from a CI point of view.

PR #163, opened 2021-08-12
Heuristic method for determining detector ID

cjbackhouse:


sbncode/CAFMaker/CAFMaker_module.cc
```diff
@@ -341,8 +340,33 @@ void CAFMaker::beginJob() {
...
+
+  std::string gdml = geom->GDMLFile();
+  gdml = basename(gdml.c_str()); // strip directory part
+  std::cout << "CAFMaker: Attempting to deduce detector from GDML file name: '" << gdml
+            << "' and configured detector name: '" << geom->DetectorName() << "'. ";
+  // Do we find the string in either of the names?
+  const bool hasSBND = ((gdml.find("sbnd") != std::string::npos) ||
+                        (geom->DetectorName().find("sbnd") != std::string::npos));
+  const bool hasIcarus = ((gdml.find("icarus") != std::string::npos) ||
+                          (geom->DetectorName().find("icarus") != std::string::npos));
```
PetrilloAtWork, line 354:
```
  const bool hasSBND = ((gdml.find("sbnd") != std::string::npos) ||
                        (geom->DetectorName().find("sbnd") != std::string::npos));
  const bool hasIcarus = ((gdml.find("icarus") != std::string::npos) ||
                          (geom->DetectorName().find("icarus") != std::string::npos));
```
While `geo:GeometryCore::DetectorName()` is guaranteed to be lower case, the GDML file name may have any case.
I _suggest_ the `gdml` string to be converted to lower case.

sbncode/CAFMaker/CAFMaker_module.cc
```diff
@@ -341,8 +340,33 @@ void CAFMaker::beginJob() {
 
 //......................................................................
 void CAFMaker::beginRun(art::Run& run) {
-  // fDetID = geom->DetId();
-  fDet = (Det_t)1;//(Det_t)fDetID;
+  // Heuristic method to determine the detector ID
+  const geo::GeometryCore* geom = lar::providerFrom<geo::Geometry>();
+
+  std::string gdml = geom->GDMLFile();
+  gdml = basename(gdml.c_str()); // strip directory part
```
PetrilloAtWork, line 347: I suppose you are using POSIX `basename()` because linking to `<filesystem>` library with Clang 7 and GCC 9 is cumbersome?

PetrilloAtWork (approved):
The heuristic is reasonable, and I can't think of anything better than that.
I have left behind a _suggestion_.
Another general _suggestion_ is that currently we don't have a way out if neither detector name not geometry file fulfil the logic; these are parameters that can't be changed at will, so when the heuristic fails we are stuck. An override path may be useful.
On the other end, until ICARUS decides to have a `icarus_is_so_much_better_than_sbnd.gdml` geometry (or vice versa), a failure here seems unlikely.

sbncode/CAFMaker/CAFMaker_module.cc
cjbackhouse: Honestly, I wasn't really aware of it. See new PR

PR #165, opened 2021-08-18
feature/bckhouse_detid - again

cjbackhouse:
Belatedly addressing review comments from https://github.com/SBNSoftware/sbncode/pull/163

### DIFF
--- a/sbncode/CAFMaker/CAFMakerParams.h
+++ b/sbncode/CAFMaker/CAFMakerParams.h
@@ -30,6 +30,11 @@ namespace caf
       Comment("Provide a string to override the automatic filename.")
     };
 
+    Atom<std::string> DetectorOverride { Name("DetectorOverride"),
+      Comment("Override the automatically detectected detector using 'sbnd' or 'icarus'. This parameter should usually be unset - ''"),
+      ""
+    };
+
     Atom<string> DataTier        { Name("DataTier") };
     Atom<string> FileExtension   { Name("FileExtension") };
     Atom<string> GeneratorLabel  { Name("GeneratorInput") };
--- a/sbncode/CAFMaker/CAFMaker_module.cc
+++ b/sbncode/CAFMaker/CAFMaker_module.cc
@@ -24,6 +24,7 @@
 #include <time.h>
 #include <algorithm>
 #include <cmath>
+#include <filesystem>
 #include <iomanip>
 #include <iostream>
 #include <iterator>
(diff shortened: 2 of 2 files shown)


wesketchum:
Is this required to have CAFs working in current production?

cjbackhouse:
No, these are cleanups requested in the review of the last PR that I didn't get to in time.

PetrilloAtWork:
@cjbackhouse have you tested on both Clang and GCC? I thought the reason you had not used `std::path::basename()` was because of issues with linking that _do (or did) exist_. Since you did not change `CMakeLists.txt` I can only hope that `cet_build_tools` knows what to do.
If you haven't tested, I urge @miquelnebot to launch an integration test (which should cover Clang, right?) as soon as  possible.

cjbackhouse:
I didn't use `std::path` because I didn't know about it. It worked fine on gcc, but agree we should launch a test to check clang if we think there's potential for it to be an issue.

wesketchum:
trigger build

cjbackhouse:
SBND c7:prof build failed with
```1553: sbncode/sbncode/CAFMaker/CMakeFiles/CAFMaker_module.dir/CAFMaker_module.cc.o: In function `std::__1::__fs::filesystem::path::filename() const':
1554: /cvmfs/larsoft.opensciencegrid.org/products/clang/v7_0_0/Linux64bit+3.10-2.17/bin/../include/c++/v1/filesystem:1099: undefined reference to `std::__1::__fs::filesystem::path::__filename() const'
1555: clang-7: error: linker command failed with exit code 1 (use -v to see invocation)```
so it looks like you were right @PetrilloAtWork :cry: 
any idea what does need to be specified to the linker?

PetrilloAtWork:
:-|
I should take a look at the mail thread on the topic; and then we can reiterate the question if needed.
I really hope this is solved above us, since that's somehow the point of having an infrastructure for building.

PetrilloAtWork:
Ok for going back to Jan 1, 1970.

I have just seen this experiment detection code (or a version of it) somewhere else.
That makes me think that factorising it in its own utility may be useful.

cjbackhouse:
> @cjbackhouse what is the state of this PR?

We are stuck because of the `filesystem` problems with clang

The right thing to do here is to merge the first, unrelated, commit af71276a4f7217f8be0b185c43ef451e22d80a4a
and the second hunk of the second commit, which converts things to lowercase, and leave the last little bit that uses filesystem pending as a new PR.

I don't have time to do the git surgery myself right now, but if you can figure out how, go ahead...

cjbackhouse:
OK, I have performed the necessary surgery on this branch so that the filesystem change is in its own commit, and last in the sequence.

To merge, rather than clicking the merge button, if someone does `git checkout develop; git merge 255f67314548a08a189c343a21896e38854bacd2` I think that should work. And then that would leave this PR about just the remaining filesystem change, and we can shelve it semi-indefinitely. Things work fine in 1970s mode, and I have no real desire to learn about boost.

cjbackhouse:
@miquelnebot I am a bit confused why the override and lowercase parts of this are still showing up in the "files changed" tab, even though I do see them now included in `develop`.

The reason I actually noticed is I saw these same changes included in Joseph's PR, which is against development, so shouldn't show them. Are they going to start showing up as phantom changes in all PRs?

PetrilloAtWork (approved):
Approved, even if:
* I still see changes in the "Files changed" that I think @cjbackhouse has removed
* from @miquelnebot words, merging is proceeding anyway

But all is good as long as it compiles.
:+1:

PR #166, opened 2021-08-19
Feature/lnguyen caloskimmer

VCLanNguyen:
Add a detector flag that identify sbnd or icarus (copy from heuristic method in CAFMaker) for CaloSkimmer.


gputnam:
Hi Lan! Thanks for the PR. One suggestion -- if both hasSBND and hasICARUS are false, the code should throw an exception. We would want to know if the module can't handle the input configuration.

sbncode/Calibration/fcl/sbnd/sbnd_trackcalo_skimmer.fcl
```diff
@@ -0,0 +1,106 @@
...
+    CaloLifeTimeForm: 0
+    CaloUseModBox: true
+  }
+  CorrectSCE: false
+  FlipTrack_dQdx: false
+  SpacePointModuleLabel: "pandora"
+  T0ModuleLabel: ""
+  TrackModuleLabel: "pandoraTrack"
+  UseArea: true
+  module_type: "Calorimetry"
```
etyley, line 17: Maybe more of a question for @gputnam than anything else, but is there a raeson not to use the `Gnocci` calorimetry module in both cases? It seems to me like it would be less confusiong to use that in all cases rather than use different modules for the sce corrected and uncorrected

VCLanNguyen: @etyley I wrote the fcl for sbnd. I put two calo modules just to be safe since I checked the sbnd calorimetry and saw two modules for sce uncorrected and corrected. I can edit the fcl to use the Gnocchi one to keep it consistent.

cjbackhouse:
> Hi Lan! Thanks for the PR. One suggestion -- if both hasSBND and hasICARUS are false, the code should throw an exception. We would want to know if the module can't handle the input configuration.

This is handled by the `hasSBND == hasIcarus` check, which catches both that case and the case where both are somehow detected. I disagree with `throw`. My original code this is copied from has `abort();` here. There's really nothing reasonable you can do with an exception here, you're definitely doomed, and people (and art!) have a bad habit of swallowing exceptions and covering up problems.

cjbackhouse:
NB https://github.com/SBNSoftware/sbncode/pull/165

sbncode/Calibration/TrackCaloSkimmer_module.cc
```diff
@@ -289,7 +307,7 @@ void sbn::TrackCaloSkimmer::analyze(art::Event const& e)
     fWiresToSave.clear();
 
     // Fill the track!
-    FillTrack(*trkPtr, pfp, t0, trkHits, trkHitMetas, calo, rawdigits, track_infos);
+    FillTrack(*trkPtr, pfp, t0, trkHits, trkHitMetas, calo, rawdigits, track_infos, hasSBND, hasIcarus);
```
cjbackhouse, line 310: Seems like it would be neater/clearer to have a single detector enum passed here, rather than two mutually exclusive booleans. Then all the `if(hasIcarus)` become `if(det == kICARUS)`

etyley:
@cjbackhouse Is there a reason you decided to check the gdml name for the detector rather than use the geometry service: https://nusoft.fnal.gov/larsoft/doxsvn/html/classgeo_1_1GeometryCore.html#a42ebac075e48c50ffd72ca2e41b677c0

cjbackhouse:
Which particular function? I am already looking at `DetectorName()` as well.

etyley:
Sorry, I missed that you were checking both. I assume there is a reason for not just relying on `DetectorName()`?

cjbackhouse:
> I assume there is a reason for not just relying on `DetectorName()`?

Paranoia, basically. If I understand it right, the name is just a string set in the fcl, and I could imagine someone overriding the gdml filename but not the detector name. It would be nice if it was guaranteed to be correct. Maybe it's worth opening a new issue somewhere to settle what the canonical source for knowing the detector should be.

etyley:
Okay that makes sense. I think having a single, simple way to do this would certainly be useful as we continue to develop more common code.

sbncode/Calibration/TrackCaloSkimmer_module.cc
```diff
@@ -129,6 +132,22 @@ void sbn::TrackCaloSkimmer::analyze(art::Event const& e)
...
+  for(unsigned int i = 0; i <gdml.size(); ++i) gdml[i] = std::tolower(gdml[i]); 
+
+  const bool hasSBND = ((gdml.find("sbnd") != std::string::npos) ||
+			(geometry->DetectorName().find("sbnd") != std::string::npos));
+
+  const bool hasIcarus = ((gdml.find("icarus") != std::string::npos) ||
+			(geometry->DetectorName().find("icarus") != std::string::npos));
+
+  if(hasSBND == hasIcarus) { 
+    std::cout << "Unable to automatically determine either SBND or ICARUS!" << std::endl;
```
cjbackhouse, line 147: Can we put a little more context in this message so people can figure out what happened if this goes wrong? eg start it with `TrackCaloSkimmer: `

sbncode/Calibration/fcl/sbnd/sbnd_trackcalo_skimmer.fcl
gputnam: yeah probably best to use Gnocchi

sbncode/Calibration/TrackCaloSkimmer_module.cc
VCLanNguyen: I will get this fixed once I'm back from holiday (1st September). Thanks

sbncode/Calibration/TrackCaloSkimmer_module.cc
```diff
@@ -18,6 +18,9 @@
 #include "sbncode/CAFMaker/FillTrue.h"
 #include "sbncode/CAFMaker/RecoUtils/RecoUtils.h"
 
+// C/C++ includes
+#include <filesystem>
```
cjbackhouse, line 22: We found out in the other PR that `<filesystem>` doesn't work properly under clang. I think you will have to use my old `basename()` implementation instead.

cjbackhouse (approved):

wesketchum:
trigger build

PR #167, opened 2021-08-19
Fixing SBND CAF maker fcl 

jzennamo:
Updated things to call the correct SimChannel label to match the LArG4 migration.

### DIFF
--- a/sbncode/PID/Dazzle_module.cc
+++ b/sbncode/PID/Dazzle_module.cc
@@ -89,7 +89,7 @@ class Dazzle : public art::EDProducer {
   art::ServiceHandle<art::TFileService> tfs;
   art::ServiceHandle<cheat::ParticleInventoryService> particleInventory;
 
-  art::InputTag fLArGeantLabel, fPFPLabel, fTrackLabel, fCaloLabel, fMCSLabel, fChi2Label, fRangeLabel, fClosestApproachLabel, fStoppingChi2Label;
+  art::InputTag fLArGeantLabel, fSimChannelLabel, fPFPLabel, fTrackLabel, fCaloLabel, fMCSLabel, fChi2Label, fRangeLabel, fClosestApproachLabel, fStoppingChi2Label;
   const float fMinTrackLength;
   const bool fMakeTree, fRunMVA;
   const std::string fMethodName, fWeightFile;
@@ -157,6 +157,7 @@ class Dazzle : public art::EDProducer {
 Dazzle::Dazzle(fhicl::ParameterSet const& p)
     : EDProducer { p }
     , fLArGeantLabel(p.get<std::string>("LArGeantLabel"))
+    , fSimChannelLabel(p.get<std::string>("SimChannelLabel"))
     , fPFPLabel(p.get<std::string>("PFPLabel"))
     , fTrackLabel(p.get<std::string>("TrackLabel"))
     , fCaloLabel(p.get<std::string>("CaloLabel"))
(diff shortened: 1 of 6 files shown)


sbncode/PID/Dazzle_module.cc
```diff
@@ -89,7 +89,7 @@ class Dazzle : public art::EDProducer {
   art::ServiceHandle<art::TFileService> tfs;
   art::ServiceHandle<cheat::ParticleInventoryService> particleInventory;
 
-  art::InputTag fLArGeantLabel, fPFPLabel, fTrackLabel, fCaloLabel, fMCSLabel, fChi2Label, fRangeLabel, fClosestApproachLabel, fStoppingChi2Label;
+  art::InputTag fLArGeantLabel, fSimChannelLabel, fPFPLabel, fTrackLabel, fCaloLabel, fMCSLabel, fChi2Label, fRangeLabel, fClosestApproachLabel, fStoppingChi2Label;
```
wesketchum, line 92: fLArGeantLabel no longer used and should be removed.

sbncode/PID/Razzle_module.cc
```diff
@@ -83,7 +83,7 @@ class Razzle : public art::EDProducer {
   art::ServiceHandle<art::TFileService> tfs;
   art::ServiceHandle<cheat::ParticleInventoryService> particleInventory;
 
-  art::InputTag fLArGeantLabel, fPFPLabel, fShowerLabel, fShowerSelVarsLabel;
+  art::InputTag fLArGeantLabel, fSimChannelLabel, fPFPLabel, fShowerLabel, fShowerSelVarsLabel;
```
wesketchum, line 86: fLArGeantLabel no longer used and should be removed.

sbncode/PID/sbn_pid.fcl
```diff
@@ -4,6 +4,7 @@ dazzle:
 {
   module_type:          Dazzle
   LArGeantLabel:        "largeant"
```
wesketchum, line 6: remove LArGeantLabel

sbncode/PID/sbn_pid.fcl
```diff
@@ -28,6 +29,7 @@ razzle:
 {
   module_type:         Razzle
   LArGeantLabel:       "largeant"
```
wesketchum, line 31: remove LArGeantLabel

wesketchum (changes requested):

jzennamo:
Here is the output CAF file:
`/sbnd/app/users/jaz8600/AugustPrep/Fixes/Checking/DroppingOptT0Finder/reco1_sce_Reco2-20210819T170607.caf.root`

sbncode/PID/sbn_pid.fcl
jzennamo: done

sbncode/PID/sbn_pid.fcl
jzennamo: done

sbncode/PID/Razzle_module.cc
jzennamo: done

sbncode/PID/Dazzle_module.cc
jzennamo: done

etyley (changes requested):
If we need to point the backtracker to use `simdrift` I think it is much better to change it where it is defined in sbndcode. This will save many headaches down the line if we get the default version working.

jzennamo:
Currently, in SBNDCode it inherits a configuration from LArSim but I have added that configuration in SBNDCode, does this CAF fcl inherit from SBNDCode?

etyley:
The CAF fcl does inherit a bunch of stuff from sbndcode yes, although it looks like it currently uses the standard backtracker rather than the sbnd version which should be changed.

wesketchum (approved):
Approving. Can update later to get the SBND-defined BackTracker when it is located. But this will allow us to move ahead with test production.

jzennamo:
I added two lingering issues and @gputnam reviewed the output CAF files and said they looked good

jzennamo:
This needs to be paired with https://github.com/SBNSoftware/sbnanaobj/pull/25

PR #177, opened 2021-10-13
Fill new genie_mode and genie_inttype variables

cjbacchus:
Keep up with https://github.com/SBNSoftware/sbnanaobj/pull/27


cjbackhouse:
Errors are like 'class caf::SRTrueInteraction' has no member named 'plane', but clearly it does
https://github.com/SBNSoftware/sbnanaobj/blob/develop/sbnanaobj/StandardRecord/SRTrueInteraction.h#L53
merged six days ago in PR https://github.com/SBNSoftware/sbnanaobj/commit/91732e549df0b7b48f2d5d45eb2b0ca29f92b77e
but I think that was after the sbnanaobj branch this depends on was created. I think the solution to make the tests pass may be to merge develop into https://github.com/SBNSoftware/sbnanaobj/pull/27

PR #181, opened 2021-10-15
Use filesystem API

cjbackhouse:
There is no good way to get clang to build this given current mrb/cetbuildtools etc. Leaving this as a draft until that future arrives.


wesketchum:
@cjbackhouse any details on this? Just wondering what prospects for closing it out are.

cjbackhouse:
@PetrilloAtWork filed a redmine issue about this, whose conclusion was a big shrug, so I think we are expecting this to take a long time to resolve.

PetrilloAtWork:
No. An update to the compilers is the most likely solution, which will happen one day (maybe at the time of C++20 adoption).

PR #184, opened 2021-11-03
Make End Median dQ/dx range configurable. Move range to 2cm in ICARUS.

gputnam:
Title.


wesketchum:
@mrmooney or @VCLanNguyen , can one of you review?

VCLanNguyen (approved):

henrylay97:
@miquelnebot these reco1 errors for SBND are consistent with what we're already seeing from a larsoft change. Myself and @pgreen135 are investigating them but they shouldn't hold this PR back :)

PR #188, opened 2021-11-12
feature/cafmaker_flat_option

cjbackhouse:
Adds options to produce any combination of CAFs and FlatCAFs directly from CAFMaker.

We can't merge this right now because it completely ignores the proper dependency order and makes this part of sbncode depend on sbnana, marked by `TODO`s in the patch. Slack discussion is leaning towards moving the SRProxy and FlatCAF parts from sbnana to sbnanaobj.

Tested, and it works, but for some reason you need to use
`LD_LIBRARY_PATH= $MRB_INSTALL/sbnana/$SBNANA_VERSION/slf7.x86_64.e20.prof/lib/:$LD_LIBRARY_PATH lar -c mycafmakerjob.fcl my.file.root`
This will probably be fixed by the rearrangement mooted above.


gputnam:
Just wanted to say this looks great and would be really useful! Also like the idea of moving Proxy stuff to sbnanaobj.

cjbackhouse:
Now depends on https://github.com/SBNSoftware/sbnanaobj/pull/32

wesketchum:
@gputnam was that a review you gave, or just general enthusiasm?

etyley (approved):
Looks good to me

gputnam (approved):

PR #189, opened 2021-11-15
Update flux FHiCL file under SBNEventWeight module

Ooohu:
Correct the path, and flux eventweight module would not give errors if we manually add  the path `/cvmfs/sbn.opensciencegrid.org/products/sbn/sbndata/v01_02/` to the variable `FW_SEARCH_PATH`.

However, we still need a better solution for identifying the path to completely avoid such modification after setting up ups products.


cjbackhouse (approved):

wesketchum:
Assigning @miquelnebot since it should be ready to merge into next release.

cjbackhouse:
We need to bump sbndata up to v01_03 first, right? (I reviewed this before it became clear we were cutting a new sbndata with the ups fix, rather than adjusting the old one)

Ooohu:
Updated the CMakeLists.txt to include sbndata v01_03

wesketchum:
Closing this PR, as it's contained within #192. This should not be merged (so long as #192 is merged).

PR #191, opened 2021-11-15
Update eventweight_genie_sbn.fcl

Ooohu:
Comments out GENIE v3.2 knobs.


jzennamo:
Hi @Ooohu, why did you shift the number of multisims for the flux from 1000 to 2? I think that is too small of a number to adequately cover the correlation space. I would stick with 1000 for the time being

PR #193, opened 2021-11-15
Update ICARUS detector location and detector box

gputnam:
Synchronizes the ICARUS-NuMI location with the neutrino simulation (https://github.com/SBNSoftware/icaruscode/pull/230).

### DIFF
--- a/sbncode/EventGenerator/MeVPrtl/config/numi_kaon_common.fcl
+++ b/sbncode/EventGenerator/MeVPrtl/config/numi_kaon_common.fcl
@@ -5,15 +5,14 @@ empty_kaon: {
 }
 
 # Taken from: https://github.com/SBNSoftware/icaruscode/blob/develop/fcl/gen/numi/GNuMIFlux.xml#L765
-icarus_numi_beam2detector_rotation: [   0.921228671,    0.0226872648,   0.388359401,
-				        0.00136256111,  0.998103714,   -0.061539578,
-				       -0.389019125,    0.0572211871,   0.919450845  ]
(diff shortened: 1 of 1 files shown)


wesketchum (approved):
Would still be better to pull this in from a geo service ... particularly the active bounding box.

gputnam:
If there was an available volume in the ICARUS Geometry then I would be happy to do it. But unfortunately, there is no volume that covers both cryostats without too much space outside of it.

wesketchum:
@gputnam thanks. we should figure out this for the future.

@miquelnebot sorry to confuse: was meant as a note but not needed for now. This looks ok to me to go.

sbncode/EventGenerator/MeVPrtl/config/numi_kaon_common.fcl
```diff
@@ -5,15 +5,14 @@ empty_kaon: {
 }
 
 # Taken from: https://github.com/SBNSoftware/icaruscode/blob/develop/fcl/gen/numi/GNuMIFlux.xml#L765
-icarus_numi_beam2detector_rotation: [   0.921228671,    0.0226872648,   0.388359401,
-				        0.00136256111,  0.998103714,   -0.061539578,
-				       -0.389019125,    0.0572211871,   0.919450845  ]
+icarus_numi_beam2detector_rotation: [ 0.921035925, 0.022715103,  0.388814672,
+                                                0, 0.998297825, -0.058321970
```
henrylay97, line 9: @gputnam @miquelnebot this is where the error crept in, no comma at end of this line

PR #199, opened 2021-11-17
This is a fix for a seg-fault arising from missing MWR devices

jzennamo:
Adding in a quick hot-fix for when the spills aren't matched to any MWR signals, long term this should be looked into but this gets production going. Marking this as a bug fix.

### DIFF
--- a/sbncode/BeamSpillInfoRetriever/BNBRetriever/BNBRetriever_module.cc
+++ b/sbncode/BeamSpillInfoRetriever/BNBRetriever/BNBRetriever_module.cc
@@ -569,9 +569,8 @@ sbn::BNBSpillInfo sbn::BNBRetriever::makeBNBSpillInfo
     beamInfo.M875BB_spill_time_diff = (MWR_times[0][matched_MWR[0]] - time);
   }
 
- if(unpacked_MWR[1].size() == 0){
-    std::vector<int> empty;
-    beamInfo.M876BB = empty;
+ if(unpacked_MWR[1].empty()){
+    beamInfo.M876BB.clear();
     beamInfo.M876BB_spill_time_diff = -999;//units in seconds
  }
  else{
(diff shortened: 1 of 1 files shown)


jzennamo:
trigger build

wesketchum (changes requested):
Likely better to leave `empty` as actually empty, assuming downstream code will not break with that. Will make checking for that in downstream code a bit easier.

jzennamo:
Changes Wes made are done.

jzennamo:
trigger build

sbncode/BeamSpillInfoRetriever/BNBRetriever/BNBRetriever_module.cc
```diff
@@ -560,13 +560,35 @@ sbn::BNBSpillInfo sbn::BNBRetriever::makeBNBSpillInfo
...
+    beamInfo.M876BB_spill_time_diff = -999;//units in seconds
+ }
+ else{
+   beamInfo.M876BB = unpacked_MWR[1][matched_MWR[1]];
+   beamInfo.M876BB_spill_time_diff = (MWR_times[1][matched_MWR[1]] - time);
+ }
+
+ if(unpacked_MWR[2].size() == 0){
+    std::vector<int> empty;
+    beamInfo.MMBTBB = empty;
```
PetrilloAtWork, line 585: And finally:
```suggestion
 if(unpacked_MWR[2].empty()){
    beamInfo.MMBTBB.clear();
```

PetrilloAtWork (commented):
Some programming style changes suggested.
In terms of content, I don't think there is any downstream user yet, so the empty information can become protocol, but it needs *documentation*.

jzennamo:
I addressed @PetrilloAtWork and @wesketchum's comments

jzennamo:
trigger build

PetrilloAtWork (approved):
Looks ok.

miquelnebot (approved):
Changes were adressed.

PR #203, opened 2021-11-19
change genieweight fhicl, universes 100->20

Ooohu:
For current genie weight calculators: it takes 1.5 minute to prepare 50 knob for each universe. This effect is linear such that we can speed up the code by using less universes; say 20, then we only need 30 mins to prepare knobs.


jzennamo (approved):
Excellent, thank you!

jzennamo:
trigger build

cjbackhouse:
I'm suspicious of the utility of 20 universes for analysis purposes. It seems like the real fix has to be to figure out how to speed up the initialization. We had genie knobs before with a reasonable runtime. I understand that might not be possible in time for the production, but you should understand that it's likely to become a "technology demo" rather than anything actually usable.

PR #206, opened 2021-11-30
Add another globalTree bail out condition for the case where the list of labels is empty

cjbackhouse:
This fixes the "multiple globalTrees in output" problem for job configurations that don't include any reweights. I think that problem is harmless, but aesthetically unappealing, so if this simple fix could sneak in somewhere that would be good.


brucehoward-physics (approved):
I confirm that I don't see globalTree in my data test file after this change.

PR #207, opened 2021-11-30
BNBRetriever fix: don't write entries incrementally

PetrilloAtWork:
The issue is that the buffer where `sbn::BNBSpillInfo` entries are kept is never emptied.

~The fix I propose is not fully C++-compliant, since after move a `std::vector` is only guaranteed to be destructible, while we keep reusing it assuming it empty. Adding a `fOutbeamInfos = {};` would make it compliant, but it's just too ugly for me to bear (science!).~ _Edit: checking back the C++17 standard and the common Internet lore, it appears that additional guarantees are provided for STL containers after the move and that the change is actually compliant._

**Important**: this is GitHub coding: I haven't even tried to compile it. C.I. test is imperative here.

**Note**: this pull request is to `develop`; it needs to be merged into the proper 2021C production branch too.

### DIFF
--- a/sbncode/BeamSpillInfoRetriever/BNBRetriever/BNBRetriever_module.cc
+++ b/sbncode/BeamSpillInfoRetriever/BNBRetriever/BNBRetriever_module.cc
@@ -609,7 +609,7 @@ void sbn::BNBRetriever::endSubRun(art::SubRun& sr)
 mf::LogDebug("BNBRetriever")<< "Total number of DAQ Spills : " << TotalBeamSpills << std::endl;
 mf::LogDebug("BNBRetriever")<< "Total number of Selected Spills : " << fOutbeamInfos.size() << std::endl;
 
-  auto p =  std::make_unique< std::vector< sbn::BNBSpillInfo > >(fOutbeamInfos);
+  auto p =  std::make_unique< std::vector< sbn::BNBSpillInfo > >(std::move(fOutbeamInfos));
 
   sr.put(std::move(p), art::subRunFragment());
 
--- a/sbncode/BeamSpillInfoRetriever/BNBRetriever/BNBRetriever_module.cc
+++ b/sbncode/BeamSpillInfoRetriever/BNBRetriever/BNBRetriever_module.cc
@@ -609,7 +609,8 @@ void sbn::BNBRetriever::endSubRun(art::SubRun& sr)
 mf::LogDebug("BNBRetriever")<< "Total number of DAQ Spills : " << TotalBeamSpills << std::endl;
 mf::LogDebug("BNBRetriever")<< "Total number of Selected Spills : " << fOutbeamInfos.size() << std::endl;
 
-  auto p =  std::make_unique< std::vector< sbn::BNBSpillInfo > >(std::move(fOutbeamInfos));
+  auto p =  std::make_unique< std::vector< sbn::BNBSpillInfo > >();
+  std::swap(*p, fOutbeamInfos);
 
   sr.put(std::move(p), art::subRunFragment());
 


sbncode/BeamSpillInfoRetriever/BNBRetriever/BNBRetriever_module.cc
```diff
@@ -609,7 +609,7 @@ void sbn::BNBRetriever::endSubRun(art::SubRun& sr)
 mf::LogDebug("BNBRetriever")<< "Total number of DAQ Spills : " << TotalBeamSpills << std::endl;
 mf::LogDebug("BNBRetriever")<< "Total number of Selected Spills : " << fOutbeamInfos.size() << std::endl;
 
-  auto p =  std::make_unique< std::vector< sbn::BNBSpillInfo > >(fOutbeamInfos);
+  auto p =  std::make_unique< std::vector< sbn::BNBSpillInfo > >(std::move(fOutbeamInfos));
```
cjbackhouse, line 612: Does something like this work?
```suggestion
  auto p =  std::make_unique< std::vector< sbn::BNBSpillInfo > >();
  std::swap(*p, fOutbeamInfos);
```

jzettle: I was going to simply suggest a `clear()` call but further examining the code there's probably not a safe place for that where we'd know it works as intended. It could go in `beginSubRun()` but that assumes for each file the code calls that function and we'd have one vector and entry/file that doesn't include incrementing the same vector for each file. Thanks Gianluca for looking into that and finding this.

PetrilloAtWork: @cjbackhouse: `std::swap` does work, that's how things would have been done before C++11 move semantics.
@jzettle: `clear()` call is semantically correct (I would do it here to compact the code).

I have checked again [`std::vector`](https://en.cppreference.com/w/cpp/container/vector/vector) documentation and C++ standard, and it is actually guaranteed that after the move `vector` is in a valid state with `empty()` `true`, so `clear()` would be redundant and `push_back()` appears not to have any additional requirement, so the simple `move()` should actually be compliant.

So technically the proposed code turns out to be compliant, but if you think a modification is necessary for clarity, I will make it.

jzettle: Sorry, I think I confused things. I was thinking about the possibility of a different solution entirely I think yours is definitely fine and would fix the issue. I was curious if instead of `move` we could just call `clear()` at some point but you clarifying how `move` works that accomplishes the exact same thing in likely a more foolproof way than what I suggested. I'm happy with your solution if that wasn't clear.

PetrilloAtWork: > I was curious if instead of `move` we could just call `clear()` at some point

Ah, ok: the answer is a big fat yes, it could be called instead. `move()` would be still recommended where I added it, to avoid the _copy_ of the vector content.

cjbackhouse: I am sort of nervous that we at some point fall into C++ undefined behaviour, where literally anything can happen and there's no point trying to reason what should happen "in practice".

An implementation of std::vector that had an fHasBeenMovedFrom field, a destructor that is safe when that is set, an empty() that checks for that and returns true, and where all other functions format your hard drive if that flag is set would be technically valid I think. And while that's obviously not how the standard library is written, an optimizing compiler is allowed to optimize code into that sort of form in the presence of undefined behaviour.

Is it really worth reasoning ourselves into knots for the sake of avoiding a copy of a few bytes of data?

jzennamo (approved):

jzennamo:
trigger build

jzettle (approved):

sbncode/BeamSpillInfoRetriever/BNBRetriever/BNBRetriever_module.cc
PetrilloAtWork: I don't see any undefined behaviour here, but I have no issue with being more explicit so I have accepted your suggestion (adding a `clear()` was just as good — and maybe `clear`er).

For sake of discussion: your example of `std::vector` would also need provisions so that the new state after moving is _valid_, that means that all operations must work as prescribed (for a `std::vector`, for example, `clear()` empties the container, `back()` returns the first element unless `empty()`, `push_back()` adds one element, etc.). Thus, there is no _undefined behaviour_ in C++ terms. That valid state is _unspecified_ and in principle the surviving vector might still hold any number of entries with any value, but for move _construction_ in C++14 the [additional constraints prevent anything else than the new state owning no data](https://stackoverflow.com/questions/17730689/is-a-moved-from-vector-always-empty/17735913#17735913), hence empty. Pointing to non-owned data is apparently also not an option for a compliant `std::vector`.

PR #209, opened 2021-12-01
Take the fIsRealData that was previously hard-coded to false and set …

brucehoward-physics:
…it instead based on a fhicl parameter. Move backtracker and particle inventory service to be initialized when needed - this does have the downside of meaning it can be initiated multiple times per event. Move some of the Fill function calls to be inside a check on IsRealData, e.g. where it calls something that needed the backtracker or particle inventory...

**NOTE** This was pretty hastily coded up given the time, so comments/looks are appreciated. Maybe there is a better way even. I am concurrently opening an issue and tagging some folks there.

### DIFF
--- a/sbncode/CAFMaker/CAFMakerParams.h
+++ b/sbncode/CAFMaker/CAFMakerParams.h
@@ -26,8 +26,20 @@ namespace caf
     /*   Comment("true = hide sensitive info, false = include full record") */
     /* }; */
 
+    Atom<bool> CreateCAF { Name("CreateCAF"),
+      Comment("Whether to produce an output file in CAF format"), true
+    };
+
+    Atom<bool> CreateFlatCAF { Name("CreateFlatCAF"),
+      Comment("Whether to produce an output file in FlatCAF format"), false
+    };
+
     Atom<std::string> CAFFilename { Name("CAFFilename"),
-      Comment("Provide a string to override the automatic filename.")
+      Comment("Provide a string to override the automatic filename."), ""
+    };
+
+    Atom<std::string> FlatCAFFilename { Name("FlatCAFFilename"),
+      Comment("Provide a string to override the automatic filename."), ""
     };
 
     Atom<std::string> DetectorOverride { Name("DetectorOverride"),
@@ -36,7 +48,8 @@ namespace caf
     };
 
     Atom<string> DataTier        { Name("DataTier") };
-    Atom<string> FileExtension   { Name("FileExtension") };
+    Atom<string> FileExtension   { Name("FileExtension"), ".caf.root" };
+    Atom<string> FlatCAFFileExtension { Name("FlatCAFFileExtension"), ".flat.caf.root" };
     Atom<string> GeneratorLabel  { Name("GeneratorInput") };
 
     Atom<bool> StrictMode        { Name("StrictMode"),
--- a/sbncode/CAFMaker/CAFMaker_module.cc
+++ b/sbncode/CAFMaker/CAFMaker_module.cc
@@ -105,8 +105,9 @@
 
 // StandardRecord
 #include "sbnanaobj/StandardRecord/StandardRecord.h"
-
 #include "sbnanaobj/StandardRecord/SRGlobal.h"
+
+#include "sbnanaobj/StandardRecord/Flat/FlatRecord.h"
 
 // // CAFMaker
 #include "sbncode/CAFMaker/AssociationUtil.h"
@@ -156,6 +157,7 @@ class CAFMaker : public art::EDProducer {
   CAFMakerParams fParams;
 
   std::string fCafFilename;
+  std::string fFlatCafFilename;
 
   bool fFirstInSubRun;
   bool fFirstInFile;
@@ -170,9 +172,10 @@ class CAFMaker : public art::EDProducer {
   TFile* fFile;
   TTree* fRecTree;
 
-  TH1D* hPOT;
-  TH1D* hSinglePOT;
-  TH1D* hEvents;
+  TFile* fFlatFile;
+  TTree* fFlatTree;
+
+  flat::Flat<caf::StandardRecord>* fFlatRecord;
 
   Det_t fDet;  ///< Detector ID in caf namespace typedef
 
@@ -188,10 +191,16 @@ class CAFMaker : public art::EDProducer {
   /// Map from parameter labels to previously seen parameter set configuration
   std::map<std::string, std::vector<sbn::evwgh::EventWeightParameterSet>> fPrevWeightPSet;
 
-  void AddEnvToFile();
-  void AddMetadataToFile(const std::map<std::string, std::string>& metadata);
+  std::string DeriveFilename(const std::string& inname,
+                             const std::string& ext) const;
 
-  void InitializeOutfile();
+  void AddEnvToFile(TFile* f);
+  void AddMetadataToFile(TFile* f,
+                         const std::map<std::string, std::string>& metadata);
+  void AddGlobalTreeToFile(TFile* outfile, caf::SRGlobal& global) const;
+  void AddHistogramsToFile(TFile* outfile) const;
+
+  void InitializeOutfiles();
 
   void InitVolumes(); ///< Initialize volumes from Gemotry service
 
@@ -255,6 +264,7 @@ class CAFMaker : public art::EDProducer {
 }; //Producer
 
 //.......................................................................
+
   CAFMaker::CAFMaker(const Parameters& params)
   : art::EDProducer{params},
     fParams(params()), fFile(0)
@@ -262,6 +272,7 @@ class CAFMaker : public art::EDProducer {
   // Note: we will define isRealData on a per event basis in produce function [using event.isRealData()], at least for now.
 
   fCafFilename = fParams.CAFFilename();
+  fFlatCafFilename = fParams.FlatCAFFilename();
 
   // Normally CAFMaker is run wit no output ART stream, so these go
   // nowhere, but can be occasionally useful for filtering in ART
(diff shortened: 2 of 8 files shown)


brucehoward-physics:
It seemed to be the creation of the service handle itself.

brucehoward-physics:
For what it's worth, I re-ran cafmaker in my test area using `cafmakerjob_icarus_sce_genie_and_fluxwgt.fcl` and Joseph's workflow test stage1 file in `/icarus/app/users/jaz8600/2021C_Workflow_Prep/NuMI/numi_nu_cosmics/` and then compared to the caf in that area with `diff_cafs`

In my `diff_cafs` (output at `/icarus/app/users/howard/icaruscode_Nov21_Release/test_cafmaker/RUN_DIFF_CAFS.txt`) I'm seeing some differences but they look like they are just null results, e.g. a bunch that relate to CRT-track matching
rec.reco.trk[0].crthit.hit.plane differs: -1626884672 vs -1646519680
and fewer but still non-zero of track vars like
rec.reco.trk[0].ID differs: -1906314998 vs -1488944617
and one instance of shower completeness
rec.reco.shw[139].truth.matches[2].energy_completeness differs: 4.178704321093373e+31 vs 1.0946131984469779e+33


I can repeat the test with one of the data files I was looking at too, using standard v09_37_01 and my test area. I guess because of the difference in filling/not filling certain variables that there might be more "diffs" to parse through.

sbncode/CAFMaker/CAFMaker_module.cc
```diff
@@ -774,9 +775,17 @@ void CAFMaker::produce(art::Event& evt) noexcept {
...
-  std::map<int, caf::HitsEnergy> id_to_hit_energy_map = SetupIDHitEnergyMap(hits, clock_data, *bt_serv.get());
+  std::map<int, std::vector<std::pair<geo::WireID, const sim::IDE*>>> id_to_ide_map;
+  std::map<int, std::vector<art::Ptr<recob::Hit>>> id_to_truehit_map;
+  std::map<int, caf::HitsEnergy> id_to_hit_energy_map;
+
+  if ( !fIsRealData ) {
+    art::ServiceHandle<cheat::BackTrackerService> bt_serv;
+
+    id_to_ide_map = PrepSimChannels(simchannels, *geometry);
+    id_to_truehit_map = PrepTrueHits(hits, clock_data, *bt_serv.get());
```
PetrilloAtWork, line 786: The `.get()` should not be necessary here:
```suggestion
    id_to_truehit_map = PrepTrueHits(hits, clock_data, *bt_serv);
```
(`ServiceHandle::operator *` will return a reference to the service object directly).
I _suggest_ removal.

cjbackhouse:
Looks like there are some more SR initialization cases that need to be patched, for fields that always get filled by CAFMaker for MC, but are left uninitialized in the data mode.

brucehoward-physics:
@cjbackhouse This is a sim file above. I talked with @henrylay97 and CRTHit planes and Trk IDs are in the set of instances he was looking at. I pointed out the shower one to him.

We'll see what the data one turns up to your point about additional SR initialization cases.

sbncode/CAFMaker/CAFMakerParams.h
```diff
@@ -258,6 +258,14 @@ namespace caf
       25.
     };
 
+    // TODO: may need to be thought more about for overlays. For now using this otherwise unused parameter as a hacky way 
+    //       of ignoring some of the truth info for data files (BH).
+    Atom<bool> IsRealData {
+      Name("IsRealData"),
+      Comment("Set true for a data file."),
+      false
```
PetrilloAtWork, line 266: Topic for discussion: do we really want a default value here?
Beside making the transition easier (which may still be a value, but IMO should not make things harder till the end of time), I am not sure there is a good reason to assume the input to be simulation.

PetrilloAtWork (commented):
Three general comments:
1. The presence of a default value for the new parameter is questionable (not wrong: just questionable), and I suggest to think a moment about that and take a pondered decision (which maybe already happened without me knowing, which is also fine); this is the reason why I don't straight approve the pull request, but I let you proceed if you think the current choice is good.
2. Given the usage current pattern, I would have chosen a `fIsSimulation` variable, since all instances I can see use `!fIsRealData`; of course, if in the future data-only features will be added (beam quality? calibrations?), that argument would not stand.
3. As Bruce wrote elsewhere, `ServiceHandle` _construction_ will throw if the service can't be found. To be even more explicit: all configured services are constructed at the beginning of the job and before all modules. Calling `ServiceHandle` only picks the pointer to the service object, which exists already. So the overhead for creating and for dereferencing a service handle is very low, and (a) the first construction of `ServiceHandle` is the one which needs error checking and (b) constructing handles multiple times does not trigger service reinitialization or anything beside some lookup in _art_ service registry.

It might have also had sense to save either a pointer or an `std::optional` to the two backtracking service objects as class data members to be conditionally filled at module construction time, but that's matter of choice.

Finally and foremost, thanks Bruce for taking care of this.

sbncode/CAFMaker/CAFMakerParams.h
brucehoward-physics: Happy to change but this would mean we need to add the `false` setting to the current workflows in ICARUS/SBND for CAFs from simulation. I wasn't sure about tampering with the default workflow, but if the sentiment is to go for it then we can 😅

sbncode/CAFMaker/CAFMaker_module.cc
brucehoward-physics: Sure - I just copied the previous function call into the loop, but I can make this change. Are the other instances of `.get()` similar? (I thought I spotted a few last night...)

brucehoward-physics:
I responded to point 1 in the comments. Ultimately I say it's not my choice, but am happy to participate in discussion! Having the default be `false` means that current workflows remain unchanged, but if it's better to change them for the future then that's fair.

There was already an `fIsRealData` parameter that was sort-of extraneous (it seemed to only be filled as a CAF file variable and was always hard-coded to false) and I have just co-opted it and tried to make it more useful to fit this purpose.

(Tagging @PetrilloAtWork directly because I will shortly be making another post with info from the `diff_cafs` on a data event and don't want this response to be lost...)

brucehoward-physics:
Ran a single data event through a version with default CAFMaker and just a somewhat edited fhicl, meant to be more in-line with the current state of my `icaruscode` PR (e.g. also having the BackTracker and ParticleInventory services). Also ran the event through my test area CAFMaker. Then used `diff_cafs` to compare: `/icarus/app/users/howard/icaruscode_Nov21_Release/test_cafmaker_data/nominal/RUN_DIFF_CAFS.txt`.

I took a very quick look and report a few things below. Are the following sorts of behaviors, or any others that people spot in that file or other testing, considered reasonable enough behavior?

Of course, most of the diffs are related to "truth" information, which makes some sense since in one we moved a lot of fills behind a check on `fIsRealData` meaning that things can be treated differently. From a pretty quick look, it seems they fall into ~3 classes.

1. Seemingly null value vs. `0` or vice-versa, like `rec.reco.shw[0].truth.p.parent differs: 0 vs 2141192192` or `rec.reco.trk[1].truth.p.parent differs: 265537960 vs 0`
2. Similar variable but that has two seemingly null values like `rec.slc[57].reco.shw[1].truth.p.parent differs: 1667330644 vs 2141192192`
3. Default value vs. `nan`, e.g. `rec.reco.shw[4].truth.eff differs: -1.0 vs nan`

The lines not matching to "truth" (`grep -v "truth"`) show mostly the track -> CRTHit plane null differences as before, like
`rec.reco.trk[1].crthit.hit.plane differs: 191717616 vs 185549280`
and a few instances of the track ID type
`rec.reco.trk[0].ID differs: -1426694211 vs 1592542871`
and finally the very much expected
`rec.hdr.ismc differs: 1 vs 0`

sbncode/CAFMaker/CAFMaker_module.cc
brucehoward-physics: Indeed there were multiple and multiple `*pi_serv.get()`. But since they are all service handle `.get()` calls with the `*` operator, I assume they can all be changed in this fashion. If this is not the case let me know!

cjbackhouse:
The most interesting differences would be any you see between two runs over the same mode in isRealData mode. Those will be missing initializations we'd want to clean up for diff_cafs to be properly useful.

brucehoward-physics:
After adding back in the stubs, can see they appear to have similar printouts to the other stuff in terms of before/after changing how this filling happens. They also have some interesting truth parent printouts like
` rec.reco.stub[4].truth.p.parent differs: 64 vs 2`

To @cjbackhouse point in the most recent comment: not a bad idea. I compared the one I made for this test and one I made yesterday and filtered out the lines matching `stub`. The only remaining lines seem to be CRTHit and Trk ID related, so nothing "new".

brucehoward-physics:
**Context**

This comment is in the current CAFMaker around the fIsRealData param:
`use this instead of evt.isRealData(), see init in [...] produce(evt)`

However I believe this is vestigial and left-over from NOvA's CAFMaker (especially given that it was previously just being hard-coded to `false`)? For now, I've removed this comment, added a comment to the constructor, and I left this `fIsRealData` parameter in the code and use it but overwrite it at each new event with `evt.isRealData()`.

These changes nullify Gianluca's question regarding defaulting a fhicl parameter since there is no longer a fhicl parameter for this.

**A remaining point/question**

The filling of true (g4) particles is itself inside of a check on `mc_particles.isValid()` [here](https://github.com/SBNSoftware/sbncode/blob/feature/howard_cafmaker_fixBackTrackParticleInventory/sbncode/CAFMaker/CAFMaker_module.cc#L796) and since the related labels should be set to null for data I think this block is not hit in my tests. I did **not** put this block into an explicit check on `fIsRealData` but can do if that is better. Let me know if you think this would be better?

**For the reviewers**

@gputnam @cjbackhouse @PetrilloAtWork - I believe this is ready to be looked at again? Please let me know if things are left or are newly unsatisfactory or if other comments remain to be addressed (it's possible I'm missing or forgetting something!)? I did ask one question in the above as well. Thanks!

cjbackhouse:
Yeah, I think that's a NOvA comment. We had trouble with overlays. If you start with data and then overlay MC into it, art thinks you have a data event. But you probably do want to be able to query the truth of the stuff you added. So we wound up with some kind of check whether there were any truth products in the record, rather than relying on art's answer.

cjbackhouse (approved):

PetrilloAtWork (commented):
It looks ok.
The only thing I am a bit fought about is that now `fIsRealData` does not really need to be a class data member: it could be a variable local to `produce()`, with no other consequence.
It bothers me a bit because it suggests that we know which kind the data is, which is actually true only after the first event is read.
After some deliberation, I _recommend_ that this variable be removed from the class, and a local one be used. But I will accept a refusal.

brucehoward-physics:
Sure - I'm fine either way, so I've pushed a change. Obviously I can't say if others feel strongly a certain way.

I did not re-run any test as this seems a very minor change...

gputnam (approved):

PR #211, opened 2021-12-02
POT Accounting in CAFs using BNB Retriever module

jzettle:
Adding POT Accounting to CAFs usign output of BNBRetriever module. Fills in SRHeader with one entry per event. More ideal is an entry per subrun.

### DIFF
--- a/sbncode/CAFMaker/CAFMakerParams.h
+++ b/sbncode/CAFMaker/CAFMakerParams.h
@@ -60,8 +60,8 @@ namespace caf
       Comment("List of suffixes to add to TPC reco tag names (e.g. cryo0 cryo1)")
     };
 
-    Atom<string> POTDataLabel {
-      Name("POTDataLabel"),
+    Atom<string> BNBPOTDataLabel {
+      Name("BNBPOTDataLabel"),
       Comment("Label of BNBRetriever module"),
       "bnbinfo"
     };
--- a/sbncode/CAFMaker/CAFMakerParams.h
+++ b/sbncode/CAFMaker/CAFMakerParams.h
@@ -60,6 +60,12 @@ namespace caf
       Comment("List of suffixes to add to TPC reco tag names (e.g. cryo0 cryo1)")
     };
 
+    Atom<string> POTDataLabel {
+      Name("POTDataLabel"),
+      Comment("Label of BNBRetriever module"),
+      "bnbinfo"
+    };
+
     Atom<string> G4Label {
       Name("G4Label"),
       Comment("Label of G4 module."),
--- a/sbncode/CAFMaker/CAFMaker_module.cc
+++ b/sbncode/CAFMaker/CAFMaker_module.cc
@@ -451,9 +451,9 @@ void CAFMaker::beginSubRun(art::SubRun& sr) {
     fBNBInfo.clear();
     fSubRunPOT = 0;
     art::Handle<std::vector<sbn::BNBSpillInfo> > bnb_spill;
-    sr.getByLabel(fParams.POTDataLabel(), bnb_spill);
-    std::vector<art::Ptr<sbn::BNBSpillInfo> >  bnb_spill_info;
+    sr.getByLabel(fParams.BNBPOTDataLabel(), bnb_spill);
     if (bnb_spill.isValid()) {
+      std::vector<art::Ptr<sbn::BNBSpillInfo> >  bnb_spill_info;
       art::fill_ptr_vector(bnb_spill_info, bnb_spill);
       FillExposure(bnb_spill_info, fBNBInfo, fSubRunPOT);
       fTotalPOT += fSubRunPOT;
@@ -1409,10 +1409,10 @@ void CAFMaker::produce(art::Event& evt) noexcept {
   StandardRecord* prec = &rec;
   fRecTree->SetBranchAddress("rec", &prec);
   fRecTree->Fill();
-  fBNBInfo.clear();
-  rec.hdr.pot = 0;
   srcol->push_back(rec);
   evt.put(std::move(srcol));
+  fBNBInfo.clear();
+  rec.hdr.pot = 0;
 }
 
 void CAFMaker::endSubRun(art::SubRun& sr) {
--- a/sbncode/CAFMaker/CAFMaker_module.cc
+++ b/sbncode/CAFMaker/CAFMaker_module.cc
@@ -17,6 +17,7 @@
 #include "sbncode/CAFMaker/FillFlashMatch.h"
 #include "sbncode/CAFMaker/FillTrue.h"
 #include "sbncode/CAFMaker/FillReco.h"
+#include "sbncode/CAFMaker/FillExposure.h"
 #include "sbncode/CAFMaker/Utils.h"
 
 // C/C++ includes
@@ -168,7 +169,7 @@ class CAFMaker : public art::EDProducer {
   double fSubRunPOT;
   double fTotalSinglePOT;
   double fTotalEvents;
-  std::vector<caf::SRBNBInfo> bnb_info_store;
+  std::vector<caf::SRBNBInfo> fBNBInfo; ///< Store detailed BNB info to save into the first StandardRecord of the output file
   // int fCycle;
   // int fBatch;
 
@@ -445,51 +446,23 @@ void CAFMaker::beginRun(art::Run& run) {
 void CAFMaker::beginSubRun(art::SubRun& sr) {
   // get the POT
   // get POT information
-  fIsRealData = true;
   if(fIsRealData)
   {
-    bnb_info_store.clear();
+    fBNBInfo.clear();
     fSubRunPOT = 0;
     art::Handle<std::vector<sbn::BNBSpillInfo> > bnb_spill;
-    sr.getByLabel("bnbinfo", bnb_spill);
+    sr.getByLabel(fParams.POTDataLabel(), bnb_spill);
     std::vector<art::Ptr<sbn::BNBSpillInfo> >  bnb_spill_info;
     if (bnb_spill.isValid()) {
       art::fill_ptr_vector(bnb_spill_info, bnb_spill);
-      for(unsigned int i = 0; i < bnb_spill_info.size(); ++i)
-      {
-	caf::SRBNBInfo single_store;
-	fSubRunPOT += bnb_spill_info[i]->TOR875;
-	fTotalPOT += bnb_spill_info[i]->TOR875;
-	single_store.spill_time_sec = bnb_spill_info[i]->spill_time_s;
-	single_store.spill_time_nsec = bnb_spill_info[i]->spill_time_ns;
-	single_store.event = bnb_spill_info[i]->event;
-	single_store.TOR860 = bnb_spill_info[i]->TOR860;
-	single_store.TOR875 = bnb_spill_info[i]->TOR875;
-	single_store.LM875A = bnb_spill_info[i]->LM875A;
-	single_store.LM875B = bnb_spill_info[i]->LM875B;
-	single_store.LM875C = bnb_spill_info[i]->LM875C;
-	single_store.HP875 = bnb_spill_info[i]->HP875;
-	single_store.VP875 = bnb_spill_info[i]->VP875;
-	single_store.HPTG1 = bnb_spill_info[i]->HPTG1;
-	single_store.VPTG1 = bnb_spill_info[i]->VPTG1;
-	single_store.HPTG2 = bnb_spill_info[i]->HPTG2;
-	single_store.VPTG2 = bnb_spill_info[i]->VPTG2;
-	single_store.BTJT2 = bnb_spill_info[i]->BTJT2;
-	single_store.THCURR = bnb_spill_info[i]->THCURR;
-	single_store.M875BB = bnb_spill_info[i]->M875BB;
-	single_store.M876BB = bnb_spill_info[i]->M876BB;
-	single_store.MMBTBB = bnb_spill_info[i]->MMBTBB;
-	single_store.M875BB_spill_time_diff = bnb_spill_info[i]->M875BB_spill_time_diff;
-	single_store.M876BB_spill_time_diff = bnb_spill_info[i]->M876BB_spill_time_diff;
-	single_store.MMBTBB_spill_time_diff = bnb_spill_info[i]->MMBTBB_spill_time_diff;
-	bnb_info_store.push_back(single_store);
-      }
+      FillExposure(bnb_spill_info, fBNBInfo, fSubRunPOT);
+      fTotalPOT += fSubRunPOT;
     }
   }
   else
   {
     art::Handle<sumdata::POTSummary> pot_handle;
-    sr.getByLabel("generator", pot_handle);
+    sr.getByLabel(fParams.GenLabel(), pot_handle);
     
     if (pot_handle.isValid()) {
       fSubRunPOT = pot_handle->totgoodpot;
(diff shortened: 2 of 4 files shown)


sbncode/CAFMaker/CAFMaker_module.cc
```diff
@@ -443,12 +445,56 @@ void CAFMaker::beginRun(art::Run& run) {
...
   // get POT information
-  art::Handle<sumdata::POTSummary> pot_handle;
-  sr.getByLabel("generator", pot_handle);
-
-  if (pot_handle.isValid()) {
-    fSubRunPOT = pot_handle->totgoodpot;
-    fTotalPOT += fSubRunPOT;
+  fIsRealData = true;
+  if(fIsRealData)
+  {
```
cjbackhouse, line 450: Can we pull this out into a separate `FillBNBPOTInfo` function? Maybe even in a new `FillExposure.cxx`? Hopefully we will soon also have a NuMI version of this, which is going to make beginSubRun very long.

sbncode/CAFMaker/CAFMaker_module.cc
```diff
@@ -1368,6 +1414,8 @@ void CAFMaker::produce(art::Event& evt) noexcept {
   rec.hdr.det     = fDet;
   rec.hdr.fno     = fFileNumber;
   rec.hdr.pot     = fSubRunPOT;
+  if(fIsRealData)
+    rec.hdr.bnbinfo = bnb_info_store;
```
cjbackhouse, line 1418: This is duplicating all the bnbinfo into every event, right? I think we probably only want to write it once. Not sure if you should put an `&& fFirstInFile` here, or just a `.clear()` after you've copied it into the record.

sbncode/CAFMaker/CAFMaker_module.cc
```diff
@@ -167,6 +168,7 @@ class CAFMaker : public art::EDProducer {
   double fSubRunPOT;
   double fTotalSinglePOT;
   double fTotalEvents;
+  std::vector<caf::SRBNBInfo> bnb_info_store;
```
cjbackhouse, line 171: Can we give this a name following the convention? `fBNBInfo` is probably fine. And a comment like `///< Store detailed BNB info to save into the first StandardRecord of the output file`

cjbackhouse (changes requested):
Looks good, but I have a bunch of small comments about reorganizing this.

sbncode/CAFMaker/CAFMaker_module.cc
```diff
@@ -443,12 +445,56 @@ void CAFMaker::beginRun(art::Run& run) {
 void CAFMaker::beginSubRun(art::SubRun& sr) {
   // get the POT
   // get POT information
-  art::Handle<sumdata::POTSummary> pot_handle;
-  sr.getByLabel("generator", pot_handle);
-
-  if (pot_handle.isValid()) {
-    fSubRunPOT = pot_handle->totgoodpot;
-    fTotalPOT += fSubRunPOT;
+  fIsRealData = true;
```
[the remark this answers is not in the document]
jzettle: I'm sorry I did this in somewhat in a rush and forgot to clean this up. I added this to test things as I thought this was set somewhere when I first wrote this in but I found it wasn't. I will fix this by removing that forced set.

sbncode/CAFMaker/CAFMaker_module.cc
jzettle: Sure, that was actually my first thought. I will work towards that

sbncode/CAFMaker/CAFMaker_module.cc
jzettle: Yes, this is what we discussed before. I pushed this before resolving that. Will add that in too. I can try a `fFirstInFile` and see if that works that's really what we want

sbncode/CAFMaker/CAFMaker_module.cc
jzettle: Yep, my bad I forgot to follow the convention here

sbncode/CAFMaker/CAFMaker_module.cc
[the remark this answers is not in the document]
cjbackhouse: Ah, the other PR is not merged yet, but you'll want to make sure you don't clash too badly with it https://github.com/SBNSoftware/sbncode/pull/209

jzettle:
Okay, update in latest commit that I think addresses all comments from @cjbackhouse. Tested and works as I think intended where there's only one non-zero entry at the firstInFile. Thanks for the review and any other comments I'm happy to address.

sbncode/CAFMaker/CAFMaker_module.cc
```diff
@@ -443,12 +446,28 @@ void CAFMaker::beginRun(art::Run& run) {
...
-    fTotalPOT += fSubRunPOT;
+  if(fIsRealData)
+  {
+    fBNBInfo.clear();
+    fSubRunPOT = 0;
+    art::Handle<std::vector<sbn::BNBSpillInfo> > bnb_spill;
+    sr.getByLabel(fParams.POTDataLabel(), bnb_spill);
+    std::vector<art::Ptr<sbn::BNBSpillInfo> >  bnb_spill_info;
+    if (bnb_spill.isValid()) {
+      art::fill_ptr_vector(bnb_spill_info, bnb_spill);
```
cjbackhouse, line 457:
```suggestion
    if (bnb_spill.isValid()) {
      std::vector<art::Ptr<sbn::BNBSpillInfo> >  bnb_spill_info;
      art::fill_ptr_vector(bnb_spill_info, bnb_spill);
```

sbncode/CAFMaker/CAFMaker_module.cc
```diff
@@ -1385,6 +1409,8 @@ void CAFMaker::produce(art::Event& evt) noexcept {
   StandardRecord* prec = &rec;
   fRecTree->SetBranchAddress("rec", &prec);
   fRecTree->Fill();
+  fBNBInfo.clear();
+  rec.hdr.pot = 0;
```
cjbackhouse, line 1413: Can you move these down two lines, below the srcol stuff? Otherwise you are wiping out the POT information from the version of the StandardRecord that gets saved into the art file. We don't use that for much/anything yet, but that will be an unpleasant surprise one day.

sbncode/CAFMaker/FillExposure.cxx
```diff
@@ -0,0 +1,39 @@
+
+#include "FillExposure.h"
```
cjbackhouse, line 2:
```suggestion
#include "sbncode/CAFMaker/FillExposure.h"
```

sbncode/CAFMaker/FillExposure.h
```diff
@@ -0,0 +1,25 @@
...
+#include "art/Framework/Core/ModuleMacros.h"
+#include "art/Framework/Principal/Event.h"
+#include "art/Framework/Principal/Handle.h"
+#include "art/Framework/Principal/SubRun.h"
+
+#include "sbnanaobj/StandardRecord/StandardRecord.h"
+#include "sbnanaobj/StandardRecord/SRHeader.h"
+#include "sbnanaobj/StandardRecord/SRBNBInfo.h"
+#include "sbnobj/Common/POTAccounting/BNBSpillInfo.h"
+#include <vector>
```
cjbackhouse, line 15: The majority of these includes are unnecessary here.

sbncode/CAFMaker/CAFMakerParams.h
```diff
@@ -60,6 +60,12 @@ namespace caf
       Comment("List of suffixes to add to TPC reco tag names (e.g. cryo0 cryo1)")
     };
 
+    Atom<string> POTDataLabel {
```
cjbackhouse, line 63: Can we call this `BNBPOTInfoLabel` or similar, to allow ourselves room for a `NuMIPOTInfoLabel` in future?

cjbackhouse (approved):
Looking good. A few minor nits.

jzettle:
Those comments resolved as well. Still working as before

sbncode/CAFMaker/CAFMaker_module.cc
```diff
@@ -443,12 +446,28 @@ void CAFMaker::beginRun(art::Run& run) {
 void CAFMaker::beginSubRun(art::SubRun& sr) {
   // get the POT
   // get POT information
-  art::Handle<sumdata::POTSummary> pot_handle;
-  sr.getByLabel("generator", pot_handle);
-
-  if (pot_handle.isValid()) {
-    fSubRunPOT = pot_handle->totgoodpot;
-    fTotalPOT += fSubRunPOT;
+  if(fIsRealData)
```
PetrilloAtWork, line 449: We don't know this yet.

sbncode/CAFMaker/CAFMaker_module.cc
```diff
@@ -443,12 +446,28 @@ void CAFMaker::beginRun(art::Run& run) {
...
+      FillExposure(bnb_spill_info, fBNBInfo, fSubRunPOT);
+      fTotalPOT += fSubRunPOT;
+    }
+  }
+  else
+  {
+    art::Handle<sumdata::POTSummary> pot_handle;
+    sr.getByLabel(fParams.GenLabel(), pot_handle);
+    
+    if (pot_handle.isValid()) {
```
PetrilloAtWork, line 467: Fun with C++ (not particularly endorsing this change):
```suggestion
    if (auto pot_handle = sr.getHandle<sumdata::POTSummary>(fParams.GenLabel())) {
```

sbncode/CAFMaker/CAFMaker_module.cc
PetrilloAtWork: I suggest a duck-typing work-around (which still sucks because it may be NuMI instead of BNB etc.):
```suggestion
  auto const bnb_spill = sr.getHandle<std::vector<sbn::BNBSpillInfo>>(fParams.BNBPOTDataLabel(), bnb_spill);
  fIsRealData = bnb_spill.isValid();
  if(fIsRealData)
```
(of course then you don't need to retrieve `bnb_spill` again in the `if` branch).

Unfortunately, for some mysterious reason `art::SubRun::isRealData()` is not exposed (thus, I have no idea if it is correctly set, either).

PetrilloAtWork (changes requested):
There is a problem in the logic of `beginSubRun()` that needs to be addressed.
For the rest, C++ suggestions here and there.

sbncode/CAFMaker/CAFMaker_module.cc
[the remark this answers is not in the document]
jzettle: I reviewed that PR after the fact and I think everything is fine. As long as fIsRealData gets set via the fcl before beginSubRun is called then it should work. It looks like that is the case. Would want to confirm after this happens. I've made sure to remove the hardcode I used for testing

sbncode/CAFMaker/CAFMaker_module.cc
jzettle: There's a PR that Chris mentioned #209 that puts this in the fcl so I'm inclined to let it be set there? If not we'll probably want to coordinate that. I'm not sure where else this fIsRealData value may be used in the future.

sbncode/CAFMaker/CAFMaker_module.cc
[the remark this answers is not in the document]
brucehoward-physics: Heads up: That PR no longer uses a fhicl parameter - it was changed yesterday to determine it from the `art` data memeber (`evt.isRealData()`). *It is no longer set by beginSubrun but in the process step itself* 😬

sbncode/CAFMaker/FillExposure.cxx
PetrilloAtWork: Idle comment about this: I would argue that `FillExposure.h` and `FillExposure.cxx` living in the same directory is a requirement. If that's true, isn't actually a better choice to go Bruce's way?

Currently, we use to officially recommend Chris's approach anyway.

sbncode/CAFMaker/CAFMaker_module.cc
brucehoward-physics: (commented elsewhere in this PR just a moment ago that this is not the case any more after discussion at yesterdays SBN Ana Framework meeting. Note I'm not really following this PR that much but see my PR getting mentioned so wanted to clarify)

jzettle: Never mind, apparently this is not the case anymore (wow this stuff is hard to follow sometimes...). This is likely a good way to do this in light of `fIsRealData` being set in the process step so I need to set it in the subRun elsewhere to make this work for POT accounting...everything else should be self-consistent even with that change

sbncode/CAFMaker/FillExposure.cxx
jzettle: I'm not sure what Bruce is doing, but I just followed the convention I see with all the other "Fill" files. I agree Chris's approach seems fine too.

sbncode/CAFMaker/CAFMaker_module.cc
PetrilloAtWork: Actually in that pull request I was suggesting that `fIsRealData` be _removed_, exactly to avoid bugs like this.
The workaround I suggest here may affect that too:
1. we can decide to leave `fIsRealData` data member be, but then the use in `beginRun()` will suffer the same issue as here
2. we can decide to drop it, and then `fIsRealData` here becomes a local `isRealData` variable.

brucehoward-physics: I haven't fully looked at the code (or indeed most of the comments in this PR review outside of the few items I've been following that involved my PR 😅 ), but just throwing in my 2 cents with my current thoughts:

1) `fIsRealData` doesn't seem be being used in `beginRun()` at the moment though? Is that going to change with other suggestions or just a general comment? Unless you (@PetrilloAtWork ) are maybe considering how one might use such a data member consistently throughout the module (e.g. later when I have things in the `produce` step) or something?

2) Why not just get rid of a variable altogether (unless the answer to point 1 demands it) and just do `if ( bnb_spill.isValid() )`?

3) Could you then also do something similar for NuMI and do `if ( bnb_spill.isValid() || numi_spill.isValid() ) `

PetrilloAtWork: > fIsRealData doesn't seem be being used in beginRun() at the moment though? Is that going to change with other suggestions or just a general comment?

General comment. Until one week ago, `fIsRealData` was not used even in `beginSubRun()` and we were happy.
But then we forget the quirks, and six months from now we start using it, and much later we realize that we are missing information for one file per job (the issue is only for the first call of `begin[Sub]Run()`, after which `analyze()` kicks in; and those methods are in our situation called on each new input file).

cjbackhouse: > Unfortunately, for some mysterious reason art::SubRun::isRealData() is not exposed

I made an art issue about this years ago, and it was closed WONTFIX. They say that in principle you could mix data and MC events into a single file. Not sure how this is consistent with including it on Event given the possibility of overlays...

sbncode/CAFMaker/FillExposure.cxx
cjbackhouse: All of our code uses full paths, except for a few places that don't (lots of places in StandardRecord :cry:). If I was in charge I would require full paths everywhere. Been bitten too many times by things being picked up from unexpected places.

jzettle:
I think I've addressed everything.

sbncode/CAFMaker/FillExposure.cxx
```diff
@@ -3,37 +3,43 @@
...
-        single_store.BTJT2 = info->BTJT2;
-        single_store.THCURR = info->THCURR;
-        single_store.M875BB = info->M875BB;
-        single_store.M876BB = info->M876BB;
-        single_store.MMBTBB = info->MMBTBB;
-        single_store.M875BB_spill_time_diff = info->M875BB_spill_time_diff;
-	single_store.M876BB_spill_time_diff = info->M876BB_spill_time_diff;
-        single_store.MMBTBB_spill_time_diff = info->MMBTBB_spill_time_diff;
-        BNBInfo.push_back(single_store);
+	subRunPOT += info.TOR875;
```
PetrilloAtWork, line 41: I assume that you considered and declined the suggestion of making this
```suggestion
	subRunPOT += info.POT();
```
which is fine.

PetrilloAtWork (approved):
I confirm all the necessary and some of the optional suggestions have been implemented.
I have left one comment on [one suggestion that hasn't been received](https://github.com/SBNSoftware/sbncode/pull/211#discussion_r762234436), which is fine but I don't know whether it was a deliberate choice or just missed do I pedantically point that out.
Approved regardless of how that part goes.

sbncode/CAFMaker/FillExposure.cxx
jzettle: I did make that change, then I tried to put everything in a separate function and forgot to keep that change once I figured everything else out. Have recovered that change again.

jzettle:
Comments addressed

PR #212, opened 2021-12-03
SBNEventWeight_fix bugs that slow down the code & of incorrect looping.

Ooohu:
Fix bugs of looping universes & knobs incorrectly. 
When setting up knob values, I should have looped knobs first, then universes. I was looping universes first then knobs. The consequence from the incorrect loop is that swapping knobs order will contribute different resulting weights.

Correct FHiCL file to correctly apply all knobs are in the same universe.


jzennamo:
Does this bug mean we would be getting incorrect weights or just it would be slow?

Ooohu:
> Does this bug mean we would be getting incorrect weights or just it would be slow?

Both.

jzennamo:
OK, sounds like we'll want this in for this production release patch

Ooohu:
Yes, please!

mroda88 (approved):
good for me

miquelnebot (approved):

cjbackhouse:
Does the speedup coming from this allow us to re-increase the number of universes?

PR #216, opened 2021-12-08
SBNEventWeight_ Add more universes for the genie module.

Ooohu:
After fixing the bug of incorrect looping in the Genie Weight module, it is expected to take 2 seconds per universe to prepare the genie weight calculator. For 1000 universes, it takes about 34 minutes to prepare the calculators before looping events.

Also remove two knobs `FrPiProd_pi`, `FrPiProd_N` to avoid cushion error form the code. This knobs are implied by other `Fr*` knobs in the current GENIE version.


Ooohu:
Have merged it to the latest develop branch, should it be able to compile using `v09_37_01`?

cjbackhouse:
There seem to be a lot of unrelated changes mixed in here, most of which seem to not have any real effects. It would be much easier to approve if it was just the `n_universes` line, plus maybe removing those `FrPiProd` knobs.

jzennamo:
I second @cjbackhouse's comment, I had expected this to be only a change in the number of universes, are there other changes included in here? @Ooohu could you update this PR to contain only a change to the number of universes?

Ooohu:
Ok! I will save others for next PR.

jzennamo (approved):
Sweet! Thanks!

wesketchum:
I'm closing this PR, as this is now included in `feature/wketchum_2021C_patch_9Dec`, and this contains a number of the other commits that we don't want yet.

PR #223, opened 2021-12-21
Correctly zero-initialize CAFMaker output files and trees in all cases

cjbackhouse:


henrylay97 (approved):
Yeah as discussed this looks great (and is nice and neat) and I've tested it and happy with it :D

Issue #221, opened 2021-12-17
CAF-precursor to CAF, including additional weights


cjbacchus:
Once we have the ability to create CAF-precursor files (https://github.com/SBNSoftware/sbncode/issues/220), we will want a way to CAF them. That much is trivial, just a fcl that runs the CAFMaker module alone, making use of the StandardRecord objects in the precursor file.

The reason to want all this, though, is the ability to add new weights to the CAF. So the fcl would additionally run `genieweight` (and save the art-level output so we can incrementally add yet more weights in a later pass). And then there needs to be some kind of logic to merge the new weights with the existing StandardRecord objects. This may mean splitting the weights-to-SR-representation-of-weights logic out from StandardRecordMaker to a new SRWeightsMaker.

Issue #220, opened 2021-12-17
Ability to create CAF-precursor art files


cjbackhouse:
Once CAFMaker is split in 2 (https://github.com/SBNSoftware/sbncode/issues/219), it will be reasonable to create "CAF precursor" files (better name/metaphor required) which are art files containing nothing but StandardRecord objects. These are logically equivalent to CAFs, but stored in art format. One would simply run the "StandardRecordMaker" but not "CAFMaker" step, and use an `outputCommands` to drop essentially everything but the StandardRecords.

The motivation to do this is to allow re-creating the CAFs, particularly with new systematic weights. What we are looking for here, then, is a fcl that does the above, and additionally allows whatever genie-level objects are necessary for computation of the genie weights to survice the outputCommands. As I write this, I realize we will also need some mechanism to understand which genie objects are associated to which reconstructed slice. Perhaps this information is already fully-retrievable from the StandardRecord structure. Perhaps we will need to add something.

wesketchum:
Can this be closed @cjbackhouse ?

cjbackhouse:
No, this is part of a series of issues detailing a longer-term plan. None of this has happened. Same for https://github.com/SBNSoftware/sbncode/issues/219 and https://github.com/SBNSoftware/sbncode/issues/221

Issue #219, opened 2021-12-17
Split CAFMaker into record-making and CAF-making parts


cjbackhouse:
CAFMaker does (at least) two things

1. Summarizes art level information into StandardRecord objects
2. Writes those objects to an output TFile (the CAF)

This issue is to split the two parts into separate modules (StandardRecordMaker and CAFMaker?) that would usually run one after the other, reproducing the current behaviour, but also enabling some other modes of operation.

The records from 1. would be communicated to 2. by writing them into the art record. That is already happening, they are simply immediately dropped because we don't configure an art output file.

Probably this is a little more complicated than I have made it sound, due to things done at the run/subrun level (I think 1. should write an SRGlobal product to be written by 2. for example), metadata, etc. The only way to find all these intertwined places is to try and separate them. The sooner this is done, the less untangling there will be to do.

PR #225, opened 2022-01-11
CAF fix merge into production branch

miquelnebot:
This is meant to include CAF fix needed for the production patch, but also some other later developments (BNB retriever, flux reader, fcl changes...)
Could people involved confirm what is needed/wanted for the production patch? or it's just SBNSoftware/sbncode#223 ?


henrylay97:
I can't comment on whether we/need want all of this. I can't see any glaring reasons why not! My only comment is on the version numbers & dependencies to make sure they are reverted back to the ones we want in the production branch.

cjbackhouse:
> Also, what's the relation with the issues #219 #220 #221

These are a longer-term plan. Nothing in any of these commits implements any of this, but nor do we need to right now.

cjbackhouse:
Do we need the FluxReader stuff? Isn't that independent of CAF making?

Otherwise, I guess we want all this (some of it I know we definitely do, other parts seem fine, and we probably want to have the CAFs produced be as close as possible to the state in develop).

I think the volume of changes here tells us something about the timing of the production branches i.e. that we should probably have chosen a more recent branch-off point for the release to be used for CAF making.

PR #227, opened 2022-01-17
OpT0Finder with new LArG4

marcodeltutto:
This PS switches from using the `OpFastScintillation` class in `LegacyLArG4` to using the new `SemiAnalyticalModel` class. The `SemiAnalyticalModel` class, introduced with PR LArSoft/larsim#84, returns visibilities instead of number of photons, and the code has been adjusted accordingly. This allows addressing SBND issue SBNSoftware/sbndcode#163.

While working on this, I noticed that the `GeoAlgo` library used by this flash matching and located at `sbncode/OpT0Finder/flashmatch/GeoAlgo` duplicates the library in `larcorealg/GeoAlgo`. This was causing problems when running the `OpT0Finder` module with other modules that use `GeoAlgo` in `larcorealg`. We should change the namespace of the one in `sbncode/OpT0Finder/flashmatch`, but for now I am simply switching to the `larcorealg` one.

### DIFF
--- a/sbncode/OpT0Finder/flashmatch/Algorithms/CMakeLists.txt
+++ b/sbncode/OpT0Finder/flashmatch/Algorithms/CMakeLists.txt
@@ -7,7 +7,8 @@ art_make(
         larsim_PhotonPropagation
         larsim_PhotonPropagation_PhotonVisibilityService_service
         larsim_LegacyLArG4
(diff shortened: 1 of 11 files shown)


pgreen135 (approved):
Thanks Marco. @miquelnebot I've run the CI tests from the corresponding sbndcode PR https://github.com/SBNSoftware/sbndcode/pull/247 and the changes are as expected.

PR #236, opened 2022-02-11
Feature/howard v09370103p01 nuid features

brucehoward-physics:
Updated PR that uses v09_37 series as the base. Requires changes in larpandoracontent to be useful.

@cjbackhouse is [this](https://github.com/SBNSoftware/sbncode/blob/4cb002fa8ec92408c82aab92fbb8152a1482d583/sbncode/CAFMaker/FillReco.cxx#L764) basically what you had in mind for the helper function?

### DIFF
--- a/sbncode/CAFMaker/FillReco.cxx
+++ b/sbncode/CAFMaker/FillReco.cxx
@@ -264,7 +264,7 @@ namespace caf
     srslice.nuid_crLongestTrackDirY =-9999.f;
     srslice.nuid_crLongestTrackDeflection = -9999.f;
     srslice.nuid_crFracHitsInLongestTrack = -9999.f;
-    srslice.nuid_nCRHitsMax = 0;
+    srslice.nuid_crNHitsMax = 0;
 
     // collect the properties
     if (primary_meta != NULL) {
@@ -312,7 +312,7 @@ namespace caf
         srslice.nuid_crFracHitsInLongestTrack = properties.at("CRFracHitsInLongestTrack");
       }
       if (properties.count("CRNHitsMax")) {
-        srslice.nuid_nCRHitsMax = properties.at("CRNHitsMax");
+        srslice.nuid_crNHitsMax = properties.at("CRNHitsMax");
       }
 
     }
--- a/sbncode/CAFMaker/FillReco.cxx
+++ b/sbncode/CAFMaker/FillReco.cxx
@@ -761,9 +761,10 @@ namespace caf
 
   //......................................................................
   template<class T, class U>
-  void CopyPropertyIfSet( const std::map<std::string, T> map, const std::string search, U &value )
+  void CopyPropertyIfSet( const std::map<std::string, T>& props, const std::string& search, U& value )
   {
-    if (map.count(search)) value = map.at(search);
+    auto it = props.find(search);
+    if ( it != props.end() ) value = it->second;
   }
 
 } // end namespace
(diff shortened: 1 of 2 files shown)


sbncode/CAFMaker/FillReco.cxx
```diff
@@ -738,5 +759,11 @@ namespace caf
     return;
   }
 
+  //......................................................................
+  template<class T, class U>
+  void CopyPropertyIfSet( const std::map<std::string, T> map, const std::string search, U &value )
+  {
+    if (map.count(search)) value = map.at(search);
+  }
```
cjbackhouse, line 767: `search` should definitely be a const reference. I think `map` (bad name because it clashes with the type, maybe `props`?) should be too, but I think you will need to change the implementation to not use `at()`. You can use `find()` to get an iterator and dereference that.

Something like this
```suggestion
  //......................................................................
  template<class T, class U>
  void CopyPropertyIfSet( const std::map<std::string, T>& props, const std::string& search, U& value )
  {
    auto it = props.find(search);
    if(it != props.end()) value = *it;
  }
```

sbncode/CAFMaker/FillReco.h
```diff
@@ -35,6 +35,7 @@
 #include "nusimdata/SimulationBase/MCTruth.h"
 
 #include "sbnanaobj/StandardRecord/SRSlice.h"
+#include "sbnanaobj/StandardRecord/SRNuID.h"
```
cjbackhouse, line 38: Do you need this include?

cjbackhouse (approved):

sbncode/CAFMaker/FillReco.cxx
brucehoward-physics: The map iterator returns a pair, so I think I want `it->second` instead of `*it` right? Update to method and removing the include noted below now committed. Does this seem better?

sbncode/CAFMaker/FillReco.cxx
```diff
@@ -255,6 +255,16 @@ namespace caf
...
+    srslice.nuid.nufspfos = -9999.f;
+    srslice.nuid.nutothits = -1;
+    srslice.nuid.nuvtxy = -9999.f;
+    srslice.nuid.nuwgtdirz = -9999.f;
+    srslice.nuid.nusps = -9999.f;
+    srslice.nuid.nueigen = -9999.f;
+    srslice.nuid.crlongtrkdiry =-9999.f;
+    srslice.nuid.crlongtrkdef = -9999.f;
+    srslice.nuid.crlongtrkhitfrac = -9999.f;
+    srslice.nuid.crmaxhits = -1;
```
PetrilloAtWork, line 267: Isn't this what `srslice.nuid.setDefault()` is for?

PetrilloAtWork (approved):
One comment about duplication of values in the code.
But in general good to go (although I would still like to see the names camel-cased).

henrylay97 (approved):
Looks good for me. I appreciate the non-physical default values as requested in the previous iteration of this PR :D

sbncode/CAFMaker/FillReco.cxx
brucehoward-physics: Yeah I suppose I could just run `setDefault()` here 😅 I can make this change when I make the Standard Record changes.

brucehoward-physics: ✅

PR #237, opened 2022-02-14
Update eventweight_genie_sbn.fcl

Ooohu:
Remove problematic knobs.


jedori0228 (approved):
Looks good to me. We might want to remove `FormZone` from `multisigma` mode too, but not as critical as in `multisim`.

marcodeltutto (approved):
Thanks for putting this together!

PR #239, opened 2022-02-17
Feature/jdyer_KDIFonlyOption

jamiecd:
I made changes to enable generation of Higgs Portal Scalar events only from kaons decaying in flight.


gputnam:
Just one comment on this PR -- I implemented the KDAR option as only selecting for KDAR at the absorber (hence the z position cut). There will also be a fraction of KDAR from the decay pipe, which this implementation will allow through. But that shouldn't stop approval of this PR -- it's up to Jamie what she wants the behavior of the KDIF option to be.

jamiecd:
Thanks Gray, good point! I'll exclude the decay-pipe KDAR. I just made the needed change.

jamiecd:
@miquelnebot I have made a few more commits - some that update the code and one that I believe will resolve the conflict above. Can you try to merge the code again?

(@grayputnam - unless this is not the right branch to be merging changes to?)

gputnam:
There was a git conflict induced by this PR alongside https://github.com/SBNSoftware/sbncode/pull/243/files. I've resolved the conflict to include the changes from both PR's. This should be good to go now.

gputnam (approved):

PR #243, opened 2022-03-03
Change Nominal Higgs Portal Mixing Angle to 5e-5

gputnam:
Fcl change to the nominal Higgs Portal mixing angle.


gputnam:
Added another commit -- also add in more configuration options to the Higgs-Portal generator.

mroda88 (approved):
It's ok. It's just a change of a parameters that is supposed to be changed many times in a single analysis.

PR #248, opened 2022-03-11
Port file_size_ana from NOvA

cjbackhouse:
Can make numerical printouts or ring or bar plots of CAF/flatcaf/art file sizes.


cjbackhouse:
Filed https://github.com/SBNSoftware/sbncode/issues/249 about where this should live long-term

henrylay97 (approved):
Very useful script, have tested it locally with some modern artroot, caf and flatcaf files successfully

cjbackhouse:
No, #249 is about potentially having these scripts live in different locations (to be decided).

PR #250, opened 2022-03-16
CRUMBS

henrylay97:
This is the major PR to implement the CRUMBS tool. Small PRs are also appearing in sbnobj sbnanaobj and sbndcode which I will reference to this one.

This PR can be split into two elements:

The first four commits (0adb103 ea7187b b3130f8 c2ba354) implement the refactoring of the StoppingChi2Fittter module into a separate Alg that is run by the producer (rather than all living in the producer). This is so that I can run it in a different configuration for a handful of tracks within the CRUMBS module.

The other five commits (8c8ba02 3489c87 30a5339 c3fd9ae fa4bbc2) create the CRUMBS module, run it as a producer during CAFMaker and then add the result (and input variables) to the caf file.

I will follow this PR up in future with some examples of how to use CRUMBS outside of the CAF files.

### DIFF
--- a/sbncode/CAFMaker/CAFMaker_module.cc
+++ b/sbncode/CAFMaker/CAFMaker_module.cc
@@ -113,6 +113,8 @@
 #include "sbnanaobj/StandardRecord/SRGlobal.h"
 
 #include "sbnanaobj/StandardRecord/Flat/FlatRecord.h"
+#include "lardataobj/RawData/ExternalTrigger.h"
+#include "lardataobj/RawData/TriggerData.h"
 
 // // CAFMaker
 #include "sbncode/CAFMaker/AssociationUtil.h"
@@ -1010,10 +1012,39 @@ void CAFMaker::produce(art::Event& evt) noexcept {
   GetByLabelStrict(evt, fParams.CRTHitLabel(), crthits_handle);
   // fill into event
   if (crthits_handle.isValid()) {
+
+    //==== gate start time
+    //==== 03/31/22 : 1600000 ns = 1.6 ms is the default T0Offset in MC
+    //==== https://github.com/SBNSoftware/icaruscode/blob/v09_37_02_01/icaruscode/CRT/crtsimmodules_icarus.fcl#L11
+    uint64_t m_gate_start_timestamp = fParams.CRTSimT0Offset(); // ns
+    if(isRealData){
+
+      art::Handle< std::vector<raw::ExternalTrigger> > externalTrigger_handle;
+      evt.getByLabel( fParams.TriggerLabel(), externalTrigger_handle );
+      const std::vector<raw::ExternalTrigger> &externalTrgs = *externalTrigger_handle;
+
+      art::Handle< std::vector<raw::Trigger> > trigger_handle;
+      evt.getByLabel( fParams.TriggerLabel(), trigger_handle );
+      const std::vector<raw::Trigger> &trgs = *trigger_handle;
+
+      if(externalTrgs.size()==1 && trgs.size()==1){
+        long long TriggerAbsoluteTime = externalTrgs[0].GetTrigTime(); // Absolute time of trigger
+        double BeamGateRelativeTime = trgs[0].BeamGateTime(); // BeamGate time w.r.t. electronics clock T0 in us
+        double TriggerRelativeTime = trgs[0].TriggerTime(); // Trigger time w.r.t. electronics clock T0 in us
+        m_gate_start_timestamp = TriggerAbsoluteTime + (int)(BeamGateRelativeTime*1000-TriggerRelativeTime*1000);
+      }
+      else{
+        std::cout << "Unexpected in " << evt.id() << ": there are " << trgs.size()
+          << " triggers in '" << fParams.TriggerLabel().encode() << "' data product."
+          << " Please contact CAFmaker maintainer." << std::endl;
+        abort();
+      }
+    }
+
     const std::vector<sbn::crt::CRTHit> &crthits = *crthits_handle;
     for (unsigned i = 0; i < crthits.size(); i++) {
       srcrthits.emplace_back();
-      FillCRTHit(crthits[i], fParams.CRTUseTS0(), srcrthits.back());
+      FillCRTHit(crthits[i], m_gate_start_timestamp, fParams.CRTUseTS0(), srcrthits.back());
     }
   }
 
--- a/sbncode/CAFMaker/FillReco.cxx
+++ b/sbncode/CAFMaker/FillReco.cxx
@@ -61,10 +61,14 @@ namespace caf
   }
 
   void FillCRTHit(const sbn::crt::CRTHit &hit,
+                  uint64_t gate_start_timestamp,
                   bool use_ts0,
                   caf::SRCRTHit &srhit,
                   bool allowEmpty) {
+
     srhit.time = (use_ts0 ? (float)hit.ts0_ns : hit.ts1_ns) / 1000.;
+    srhit.t0 = ((long long)(hit.ts0_ns)-(long long)(gate_start_timestamp))/1000.;
+    srhit.t1 = hit.ts1_ns/1000.;
 
     srhit.position.x = hit.x_pos;
     srhit.position.y = hit.y_pos;
--- a/sbncode/CAFMaker/FillReco.cxx
+++ b/sbncode/CAFMaker/FillReco.cxx
@@ -302,7 +302,8 @@ namespace caf
 
 
   void FillSliceCRUMBS(const sbn::CRUMBSResult *crumbs,
-                       caf::SRSlice& slice) {
+                       caf::SRSlice& slice,
+                       bool allowEmpty) {
     if (crumbs != nullptr) {
       slice.crumbs_result.score = crumbs->score;
       slice.crumbs_result.tpc.crlongtrackhitfrac = crumbs->tpc_CRFracHitsInLongestTrack;
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,7 +15,7 @@
 
 cmake_minimum_required(VERSION 3.19 FATAL_ERROR)
 
-project(sbncode VERSION 09.37.02 LANGUAGES CXX)
+project(sbncode VERSION 09.37.02.01 LANGUAGES CXX)
 
 message(STATUS "\n\n  ==========================   ${PROJECT_NAME}   ==========================")
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,7 +15,8 @@
 
 cmake_minimum_required(VERSION 3.19 FATAL_ERROR)
 
-project(sbncode VERSION 09.37.02.01 LANGUAGES CXX)
+
+project(sbncode VERSION 09.37.02.02 LANGUAGES CXX)
 
 message(STATUS "\n\n  ==========================   ${PROJECT_NAME}   ==========================")
 
--- a/sbncode/CAFMaker/CAFMakerParams.h
+++ b/sbncode/CAFMaker/CAFMakerParams.h
@@ -234,7 +234,19 @@ namespace caf
       Comment("Label of sbn CRT tracks."),
       "crttrack" // same for icarus and sbnd
     };
-    
+
+    Atom<long long> CRTSimT0Offset {
+      Name("CRTSimT0Offset"),
+      Comment("start of beam gate/simulation time in the simulated CRT clock"),
+      0,
+    };
+
+    Atom<art::InputTag> TriggerLabel {
+      Name("TriggerLabel"),
+      Comment("Label of trigger."),
+      "daqTrigger"
+    };
+
     Atom<string> FlashTrigLabel {
       Name("FlashTrigLabel"),
       Comment("Label of bool of passing flash trigger."),
--- a/sbncode/CAFMaker/FillReco.h
+++ b/sbncode/CAFMaker/FillReco.h
@@ -86,7 +86,8 @@ namespace caf
                        bool allowEmpty = false);
 
   void FillSliceCRUMBS(const sbn::CRUMBSResult *crumbs,
-                       caf::SRSlice& slice);
+                       caf::SRSlice& slice,
+                       bool allowEmpty = false);
 
   bool SelectSlice(const caf::SRSlice &slice, bool cut_clear_cosmic);
 
--- a/sbncode/CAFMaker/FillReco.h
+++ b/sbncode/CAFMaker/FillReco.h
@@ -158,6 +158,7 @@ namespace caf
                            unsigned truth_ind);
 
   void FillCRTHit(const sbn::crt::CRTHit &hit,
+                  uint64_t gate_start_timestamp,
                   bool use_ts0,
                   caf::SRCRTHit &srhit,
                   bool allowEmpty = false);
--- a/ups/product_deps
+++ b/ups/product_deps
@@ -36,7 +36,7 @@ fwdir   product_dir scripts
 product             version
 larsoft             v09_37_01_02
 sbnobj              v09_12_12
-sbnanaobj           v09_17_06_03
+sbnanaobj           v09_17_06_04
 sbndaq_artdaq_core  v1_00_00of2
 genie_xsec          v3_00_04a
 sbndata             v01_03
--- a/ups/product_deps
+++ b/ups/product_deps
@@ -36,7 +36,7 @@ fwdir   product_dir scripts
 product             version
 larsoft             v09_37_01_02
 sbnobj              v09_12_12
-sbnanaobj           v09_17_06_04
+sbnanaobj           v09_17_06_05
 sbndaq_artdaq_core  v1_00_00of2
 genie_xsec          v3_00_04a
 sbndata             v01_03


etyley (commented):
Looks good to me, a few minor comments about making code more efficient but nothing major. 

I do wonder if the `CRUMBS_module` would be better in its own `CosmicID` directory to separate it from the more generic producers. 

Additionally, a README (and ideally a root macro) on how to retrain the BDT would ensure reproducibility going forwards

sbncode/CAFMaker/CAFMaker_module.cc
```diff
@@ -1046,6 +1073,14 @@ void CAFMaker::produce(art::Event& evt) noexcept {
       slcHits = fmSlcHits.at(0);
     }
 
+    art::FindOneP<sbn::CRUMBSResult> foSlcCRUMBS =
+      FindOnePStrict<sbn::CRUMBSResult>(sliceList, evt,
+          fParams.CRUMBSLabel() + slice_tag_suff);
+    const sbn::CRUMBSResult *slcCRUMBS = NULL;
```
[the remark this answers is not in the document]
henrylay97: Both seem to be used within this file? I'm happy to change this if that's the right way to go. Shall I change them all while I'm at it? @cjbackhouse might have opinions?

sbncode/CAFMaker/FillReco.cxx
```diff
@@ -289,6 +289,33 @@ namespace caf
   }
 
 
+  void FillSliceCRUMBS(const sbn::CRUMBSResult *crumbs,
+                       caf::SRSlice& slice,
+                       bool allowEmpty) {
+    if (crumbs != NULL) {
```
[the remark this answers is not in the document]
henrylay97: Likewise

sbncode/LArRecoProducer/CRUMBS_module.cc
```diff
@@ -0,0 +1,681 @@
...
+	this->FillPandoraNuScoreVars(propertiesMap);
+
+	tpc_StoppingChi2CosmicRatio = this->GetLongestTrackStoppingChi2Ratio(e, slice, handlePFPs, handleSlices);
+    
+	const art::Ptr<sbn::SimpleFlashMatch> flashmatch = pfpFMVec.front();
+	pds_FMTotalScore = flashmatch->score.total;
+	pds_FMPE = flashmatch->light.pe;
+	pds_FMTime = flashmatch->time;
+      
+	if(pds_FMTime == -9999)
```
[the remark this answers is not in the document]
henrylay97: So this is due to the magic value assigned to the time in the original producer [here](https://github.com/SBNSoftware/sbncode/blob/develop/sbncode/FlashMatch/FlashPredict_module.cc#L154) for the BDT I wanted to bring its default value closer to the actual distribution.

sbncode/LArRecoProducer/CRUMBS_module.cc
```diff
@@ -0,0 +1,681 @@
...
+      }
+
+    e.put(std::move(resultsVec));
+    e.put(std::move(sliceAssns));
+  }
+
+  void CRUMBS::FillCRTVars(std::vector<art::Ptr<anab::T0> > trackT0s, std::vector<art::Ptr<anab::T0> > hitT0s)
+  {
+    if (trackT0s.size() != 0){
+      crt_TrackScore = 999999;
```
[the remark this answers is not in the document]
henrylay97: Yep fair point :+1:

sbncode/LArRecoProducer/TrackStoppingChi2Alg.cc
```diff
@@ -0,0 +1,83 @@
...
+  if(calo.XYZ().size() == 0)
+    return StoppingChi2Fit();
+
+  geo::Point_t start(calo.XYZ().front());
+  geo::Point_t end(calo.XYZ().back());
+
+  std::vector<float> dEdxVec, resRangeVec;
+  // Fill the dEdx vs res range vectors, ignoring the first/last points
+  for (size_t i = 1; i < calo.dEdx().size() - 1; i++) {
+    const float thisdEdx(calo.dEdx()[i]);
```
absolution1, line 67: Is it possible that this can segfault because of where the loop starts and ends

sbncode/CAFMaker/CAFMaker_module.cc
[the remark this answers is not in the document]
cjbackhouse: I'm a dinosaur who usually uses `0`, and would probably argue that corner in a pub discussion. But I agree that `nullptr` is probably preferred over `NULL`.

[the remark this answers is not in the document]
cjbackhouse: Also, yes, feel free to do a bulk find/replace in a separate PR.

sbncode/LArRecoProducer/CRUMBS_module.cc
```diff
@@ -0,0 +1,681 @@
...
+    runID = e.run();
+  }
+
+  void CRUMBS::produce(art::Event& e)
+  {
+    if(fTrainingMode)
+      {
+	this->ClearMaps();
+	this->SetupMaps(e);
+      }
```
cjbackhouse, line 292: These member variables, meanwhile, are not necessary. You could have one `GetMaps()`-like function and keep the maps as local variables in this function, and pass them into users. Makes the intent clearer, and prevents any possibility of accidental contamination from one event to the next.

sbncode/LArRecoProducer/CRUMBS_module.cc
```diff
@@ -0,0 +1,681 @@
...
+    auto const handleHits(e.getValidHandle<std::vector<recob::Hit>>(fHitModuleLabel));
+    std::vector<art::Ptr<recob::Hit>> allHits;
+    art::fill_ptr_vector(allHits, handleHits);
+
+    art::FindManyP<larpandoraobj::PFParticleMetadata> pfpMetadataAssoc(handlePFPs, e, fPFParticleModuleLabel);
+    art::FindManyP<sbn::SimpleFlashMatch> pfpFMAssoc(handlePFPs, e, fFlashMatchModuleLabel);
+
+    for(auto const &slice : slices)
+      {
+	this->ResetVars();
```
cjbackhouse, line 315: Likewise, the vars should be scoped to within this loop.

sbncode/LArRecoProducer/CRUMBS_module.cc
```diff
@@ -0,0 +1,681 @@
...
+	  pds_FMTime = -100;
+
+	const float score = fMVAReader->EvaluateMVA(fMVAName);
+
+	CRUMBSResult thisResult(score, tpc_CRFracHitsInLongestTrack, tpc_CRLongestTrackDeflection, tpc_CRLongestTrackDirY, tpc_CRNHitsMax,
+				tpc_NuEigenRatioInSphere, tpc_NuNFinalStatePfos, tpc_NuNHitsTotal, tpc_NuNSpacePointsInSphere, tpc_NuVertexY, 
+				tpc_NuWeightedDirZ, tpc_StoppingChi2CosmicRatio, pds_FMTotalScore, pds_FMPE, pds_FMTime, crt_TrackScore, crt_HitScore, 
+				crt_TrackTime, crt_HitTime);
+
+	resultsVec->push_back(thisResult);
```
cjbackhouse, line 354:
```suggestion
	resultsVec->emplace_back(score, tpc_CRFracHitsInLongestTrack, tpc_CRLongestTrackDeflection, tpc_CRLongestTrackDirY, tpc_CRNHitsMax,
				tpc_NuEigenRatioInSphere, tpc_NuNFinalStatePfos, tpc_NuNHitsTotal, tpc_NuNSpacePointsInSphere, tpc_NuVertexY, 
				tpc_NuWeightedDirZ, tpc_StoppingChi2CosmicRatio, pds_FMTotalScore, pds_FMPE, pds_FMTime, crt_TrackScore, crt_HitScore, 
				crt_TrackTime, crt_HitTime);
```

sbncode/LArRecoProducer/CRUMBS_module.cc
```diff
@@ -0,0 +1,681 @@
...
+	  }
+      }
+
+    e.put(std::move(resultsVec));
+    e.put(std::move(sliceAssns));
+  }
+
+  void CRUMBS::FillCRTVars(std::vector<art::Ptr<anab::T0> > trackT0s, std::vector<art::Ptr<anab::T0> > hitT0s)
+  {
+    if (trackT0s.size() != 0){
```
cjbackhouse, line 378:
```suggestion
    if (!trackT0s.empty()){
```

sbncode/LArRecoProducer/CRUMBS_module.cc
```diff
@@ -0,0 +1,681 @@
...
+	    }
+	}
+    }
+  }
+
+  void CRUMBS::FillPandoraNuScoreVars(std::map<std::string, float> propertiesMap)
+  {
+    auto propertiesMapIter = propertiesMap.find("NuScore");
+    if (propertiesMapIter == propertiesMap.end()){
+      std::cout << "Error finding variable -- NuScore" << std::endl;
```
cjbackhouse, line 407: Preface these messages with `CRUMBS_module:` so that the poor user can find where they come from.

Personally I would `abort()` rather than simply `return` here. The user who intended to run CRUMBS is not getting what they wanted, and will probably be happier that you make sure they notice now, rather than digging through their grid logs to find out why things aren't filled right.

sbncode/LArRecoProducer/CRUMBS_module.cc
```diff
@@ -0,0 +1,681 @@
...
+    }
+    tpc_NuWeightedDirZ = propertiesMapIter->second;
+  }
+
+  std::vector<art::Ptr<recob::Hit> > CRUMBS::GetAllSliceHits(art::Event const& e, const art::Ptr<recob::Slice> slice, art::ValidHandle<std::vector<recob::Slice> > handleSlices)
+  {
+    art::FindManyP<recob::Hit> sliceHitAssn(handleSlices,e,fSliceModuleLabel);
+    const std::vector<art::Ptr<recob::Hit> > sliceHits = sliceHitAssn.at(slice.key());
+
+    return sliceHits;
```
cjbackhouse, line 488:
```suggestion
    return sliceHitAssn.at(slice.key());
```

sbncode/LArRecoProducer/CRUMBS_module.cc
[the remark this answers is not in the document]
etyley: In that case it is unavoidable here. Floating point comparisons like this always make me nervous though, so you could always use something like ```pds_FMTime = std::max(pds_FMTime, -100)```

sbncode/LArRecoProducer/TrackStoppingChi2Alg.cc
```diff
@@ -0,0 +1,83 @@
...
+  // Try and fit a flat polynomial
+  graph->Fit("pol0", "Q");
+  const TF1* polFit = graph->GetFunction("pol0");
+  const float pol0Chi2(polFit ? polFit->GetChisquare() : -5.f);
+  const float pol0Fit(polFit ? polFit->GetParameter(0) : -5.f);
+
+  // Try to fit an exponential
+  graph->Fit("expo", "Q");
+  const TF1* expFit = graph->GetFunction("expo");
+  const float expChi2(expFit ? expFit->GetChisquare() : -5.f);
```
cjbackhouse, line 33: Perosnally I don't trust TGraph/TF1 etc and would just code this manually. `pol0` fit at least is simply a mean. And isn't expo just a linear fit of the log values? I guess if it works it works...

sbncode/LArRecoProducer/CRUMBS_module.cc
```diff
@@ -0,0 +1,681 @@
...
+						  art::ValidHandle<std::vector<recob::Slice> > handleSlices);
+    float GetLongestTrackStoppingChi2Ratio(art::Event const& e, const art::Ptr<recob::Slice> slice, art::ValidHandle<std::vector<recob::PFParticle> > handlePFPs,
+					   art::ValidHandle<std::vector<recob::Slice> > handleSlices);
+
+    void FillCRTVars(std::vector<art::Ptr<anab::T0> > trackT0s, std::vector<art::Ptr<anab::T0> > hitT0s);
+    void FillPandoraNuScoreVars(std::map<std::string, float> propertiesMap);
+
+    std::vector<art::Ptr<recob::Hit> > GetAllSliceHits(art::Event const& e, const art::Ptr<recob::Slice> slice, art::ValidHandle<std::vector<recob::Slice> > handleSlices);
+    std::map<int,float> SlicePurity(art::Event const& e, std::vector<art::Ptr<recob::Hit> > sliceHits);
+    float SliceCompleteness(art::Event const& e, std::vector<art::Ptr<recob::Hit> > sliceHits, std::vector<art::Ptr<recob::Hit> > allHits, const int matchedGenID);
```
cjbackhouse, line 84: I haven't checked in detail, but are all these `Ptr`s and `Handle`s really necessary? Could most of these arguments just be simple references?

sbncode/LArRecoProducer/CRUMBS_module.cc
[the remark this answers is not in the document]
henrylay97: Yep, I like that, thanks!

sbncode/LArRecoProducer/TrackStoppingChi2Alg.cc
henrylay97: No I don't think so? It starts at 1 and finishes one before the end?

sbncode/LArRecoProducer/CRUMBS_module.cc
henrylay97: Okay will do so!

sbncode/LArRecoProducer/CRUMBS_module.cc
henrylay97: I think this is a required artefact of TMVA / TTree though right? These variables need to be members and so need resetting at the start of each unit, in this case a slice.

sbncode/LArRecoProducer/TrackStoppingChi2Alg.cc
henrylay97: So this is actually from the original tool, I have just moved it into the alg so its highlighting as added lines. It seems to work for now. I think this would be a separate PR to amend this :)

sbncode/LArRecoProducer/CRUMBS_module.cc
henrylay97: Is there a good reason for changing them? Clearly this is my inexperience showing but I've tended to pass ptrs and handles in other analyzers and producers I've written before?

cjbackhouse: I guess mostly to reduce clutter, so you can see what actually matters. You only need a Ptr (say) if you want to look it up in an Assn, or other art-like thing. Normally I think it's supposed to be good practice to keep the stuff for dealing with the art interfaces, and the stuff for actually doing the physics, as separated as possible.

sbncode/LArRecoProducer/CRUMBS_module.cc
```diff
@@ -0,0 +1,681 @@
...
+
+    return slicePurityMap;
+  }
+
+  float CRUMBS::SliceCompleteness(art::Event const& e, std::vector<art::Ptr<recob::Hit> > sliceHits, std::vector<art::Ptr<recob::Hit> > allHits, const int matchedGenID)
+  {
+    int nSliceHits(0), nHits(0);
+    auto clockData = art::ServiceHandle<detinfo::DetectorClocksService>()->DataFor(e);
+
+    for (auto const& hit : sliceHits)
```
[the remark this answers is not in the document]
henrylay97: Yep, done :)

sbncode/LArRecoProducer/CRUMBS_module.cc
henrylay97: This is another one where I'm going to opt for leaving it as is unless anyone feels its a merge-stopping case. I think the code is followable (sometimes more so) and seems to be relatively common place.

henrylay97:
Will just comment on this PRs but applies to all. Very very grateful to you all for reviewing so quickly and thoroughly, I owe you all one @cjbackhouse @etyley @absolution1. 

The other 3 PRs are now approved and all conversations are resolved. This still has a few outstanding. From my point of view I'm not planning on doing anything other than moving CRUMBS to a new folder and sorting the indentation out. This will mess with the diff so I'll wait until you guys have had a chance to have a final look.

Thanks again!

etyley (approved):
Looks good to me, although I still think version controlling the root macro to run the training would be beneficial.

henrylay97:
Thanks Ed, yeah I do agree, I'm working on a cleaned up version of the macro with some documentation to go with it and will push these later / in a separate PR depending on the production schedule.

wesketchum:
@cjbackhouse since you said you wanted to approve when indentation fixed, I'll ping you on doing that :)

sbncode/CAFMaker/FillReco.cxx
```diff
@@ -289,6 +289,33 @@ namespace caf
   }
 
 
+  void FillSliceCRUMBS(const sbn::CRUMBSResult *crumbs,
+                       caf::SRSlice& slice,
+                       bool allowEmpty) {
```
cjbackhouse, line 294: allowEmpty is now unused. Either add an `if(!crumbs && !allowEmpty) abort("no crumbs! (better text please...)");` or remove the argument.

cjbackhouse (approved):
Indentation looks right now. Just one more comment on the response to a previous one.

sbncode/CAFMaker/FillReco.cxx
henrylay97: I have removed it for now. I noted, however, whilst doing this that for almost every function in `FillReco` this argument exists and is unused... something for a separate PR!

PR #255, opened 2022-04-13
Light sim+reco updates for SBN2022A release

miquelnebot:
Copy of SBNSofware/sbncode#230 to develop


pgreen135:
the diff from this just includes a lot of things I don't think we want -- this is going to be a problem for all of these branches now I suspect as 2022A has significantly diverged from develop. This PR though I suggest can be closed.

PR #258, opened 2022-04-13
SBN2022A to develop Feature/jskim genie weight module

miquelnebot:
Copy of SBNSoftware/sbncode#236 to develop


jedori0228 (changes requested):
There was a request to remove or comment-out https://github.com/SBNSoftware/sbncode/blob/0237c80fedced8d8ce252965018a6b4a449bb953/sbncode/SBNEventWeight/Calculators/CrossSections/GenieWeightCalc.cxx#L472 which prints probably ~40k lines. These information can be retrieved  from globalTree later.

PR #261, opened 2022-04-14
Change variables to pick up performance improvements

ikatza:
This PR depends on https://github.com/LArSoft/larsim/pull/90 where the big changes are happening.

Besides the performance improvements on the g4 stage advertised in that PR, the performance of the OpT0Finder module in here will also improve significantly.


marcodeltutto (approved):
Thank you, @ikatza!

ikatza:
I've been reviewing the warnings.

I think it's too few events to be able to see the difference in performance. Whilst the performance per event does improve, I also moved the initialisation of some parameters to happen all at the beginning. The increase in overhead is granted when running several events in a single job, but is likely hindering these test that only use 2 or 10 events.

With that in mind, @marcodeltutto can you confirm that `_semi_model` is initialised only once per job?

In principle there could be some difference in data products as I changed the random number generator to use the non-static type.

Are the `c7` warnings sensible? I feel like it's failing in places were it shouldn't at all. 

The single caf is mostly unrelated, I suspect it's similar for the cosmics caf.

marcodeltutto:
Hi @ikatza, yes, the `_semi_model` is set in the constructor of the LArSoft plugin, so it's initialized only once per job.

https://github.com/SBNSoftware/sbndcode/blob/94d8ee5025b8763da95270f437f27cb8d8ed9eea/sbndcode/OpT0Finder/SBNDOpT0Finder_module.cc#L175

ikatza:
The validation software has been overhauled by @chilge to be able to test changes at the g4 stage. However there remains some problems that need to be fixed in order to run a test. Chris is working on that, I can't say how long it will take.

ikatza:
I finally have some validation results. My impression is that the variations are expected, and that my original PR https://github.com/LArSoft/larsim/pull/90 doesn't introduce any errors or biases.


TPC Simulation
https://dbweb9.fnal.gov:8443/LarCI/app/ns:sbnd/build_detail/phase_details?build_id=sbnd_ci/9615&platform=Linux%203.10.0-1160.59.1.el7.x86_64&phase=ci_validation&buildtype=slf7%20e20:prof

PDS Simulation
https://dbweb9.fnal.gov:8443/LarCI/app/ns:sbnd/build_detail/phase_details?build_id=sbnd_ci/9602&platform=Linux%203.10.0-1160.59.1.el7.x86_64&phase=ci_validation&buildtype=slf7%20e20:prof

ikatza:
The change of data product sizes of `std::vector<sim::OpDetBacktrackerRecord>`  are expected, they use to be bigger because they would store some instances of `sim::OpDetBacktrackerRecord` with no detected photons.

PR #262, opened 2022-04-29
Fix use of same name algorithm

henrylay97:
The TrajectoryMCSFitter algorithm has an identical name to one that resides in [larreco](https://github.com/LArSoft/larreco/blob/develop/larreco/RecoAlg/TrajectoryMCSFitter.h). This leads to undefined behaviour which showed up in the CI when testing (unrelated) SBNSoftware/sbndcode#278. By changing the namespace in which our version of the alg resides we can fix this issue.

Kyle from SciSoft has confirmed that my understanding of this is correct and the solution has been tested with the CI system.

### DIFF
--- a/sbncode/LArRecoProducer/LArReco/TrajectoryMCSFitter.cxx
+++ b/sbncode/LArRecoProducer/LArReco/TrajectoryMCSFitter.cxx
@@ -5,7 +5,7 @@
 #include "TMatrixDSymEigen.h"
 
 using namespace std;
-using namespace trkf;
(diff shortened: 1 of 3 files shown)


pgreen135 (approved):
Thanks @henrylay97 for fixing this. This obviously isn't for you to fix as its not your code, but I will say that `using namespace` especially `using namespace std` should probably be avoided. It's in danger of leading to ambiguities / unexpected behaviour that can be trivially avoided by explicitly stating the namespaces when using functions/classes from them.

PR #266, opened 2022-05-16
Add example training script for CRUMBS tool

henrylay97:
I have finally been creating some _how to_ resources to go with a technote-style document I am making for CRUMBS. This is an example training script which I feel would be useful to have in sbncode for future retraining.


etyley (approved):
Looks good to me

PR #268, opened 2022-05-26
Updates to POT accounting codes to account for new trigger fragment

jzettle:
Emergency PR in order to restart POT accounting efforts after updates to trigger decoding.

When doing the updates to the trigger fragment structure we missed the need to update what is referenced in the POT accounting codes. The fragments pulled for the POT accounting are now updated to match the new trigger system. Only will work with new data with new trigger system (>run 8354). Will need to expand this to create a separate path (via fcl or run number flags) to handle the data with the old trigger fragment. This is easier to handle as we don't need a sophisticated decoder for this so it might just have a pre-run 8354 and post-run 8354. Will need to examine this.

Tested with run 8413 which Maya was reporting issues processing before and both BNB and NuMI are now working and run 8353 (which I think is one of first runs with new trigger fragment).

### DIFF
--- a/sbncode/BeamSpillInfoRetriever/BNBRetriever/BNBRetriever_module.cc
+++ b/sbncode/BeamSpillInfoRetriever/BNBRetriever/BNBRetriever_module.cc
@@ -20,7 +20,7 @@
 #include "larcorealg/CoreUtils/counter.h"
 
 #include "artdaq-core/Data/Fragment.hh"
-#include "sbndaq-artdaq-core/Overlays/ICARUS/ICARUSTriggerUDPFragment.hh"
+#include "sbndaq-artdaq-core/Overlays/ICARUS/ICARUSTriggerV2Fragment.hh"
 
 #include "sbnobj/Common/POTAccounting/BNBSpillInfo.h"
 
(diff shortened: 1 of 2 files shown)


PetrilloAtWork (approved):
It looks fine to me, with the caveat that Jacob already described in the pull request description.

jzennamo (approved):

PetrilloAtWork:
@jzettle this is certainly the fastest way to fix the issue.
The information that this module needs should be contained in `sbn::ExtraTriggerInfo`. Is this the case with the current decoder in the production branch? Using the information in there is the _definitive_ solution (that is because the actual work is delegated to the decoder, so if the decoder does not support a certain run, this definitive solution will still not work).

jzettle:
@PetrilloAtWork I would think this is the case, but I am not sure what data products are available when this runs this is more a question for the production team or @jzennamo. If this module runs on the pre-decoded data, then I don't think that will work as the data products produced by the decoder itself (so sbn::ExtraTriggerInfo) will not be available to my knowledge. It's possible this can be adjusted to run at a different step to use that but I do not have the ability to solve that myself before mid-next week. If it can pull information from the decoded data product it may remove the necessity for separate use cases for this module

jzettle:
So I did develop here quickly...this is actually more needed on the production branch. Will do same thing there

jzettle:
Sounds good, thanks @miquelnebot for doing this. Now there is also https://github.com/SBNSoftware/sbncode/pull/269 which is the exact same thing but on the production branch when that is created.

PR #271, opened 2022-06-04
use ts0() instead od ts0_ns in CAFMaker

jedori0228:
Followed by https://github.com/SBNSoftware/sbnobj/pull/54, `CRTHit::ts0_ns` now only takes the sub-second part of the original t0. We now should use `CRTHit::ts0()` when we fill the timing information of CRTHit in CAFMaker.

### DIFF
--- a/sbncode/CAFMaker/FillReco.cxx
+++ b/sbncode/CAFMaker/FillReco.cxx
@@ -66,9 +66,9 @@ namespace caf
                   caf::SRCRTHit &srhit,
                   bool allowEmpty) {
 
-    srhit.time = (use_ts0 ? (float)hit.ts0_ns : hit.ts1_ns) / 1000.;
-    srhit.t0 = ((long long)(hit.ts0_ns)-(long long)(gate_start_timestamp))/1000.;
-    srhit.t1 = hit.ts1_ns/1000.;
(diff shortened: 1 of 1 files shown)


biswaranjan91 (approved):

jedori0228:
> @jedori0228 we may need some more info for this PR, as is this strictly necessary for the SBN2022A production and so a new patch would be needed? Or is this that can be fixed at run time, then the PR to develop enough (and of course will be included in the next production)?

Before https://github.com/SBNSoftware/sbnobj/pull/54, `ts0_ns` returns the full timestamp in nanosecond (say `1 111 111 111 222 222 222`, a 19-digit integer), and that can be subtracted by `gate_start_timestamp` which is also the a timestamp in nanosecond. But now, it only contains the sub-second part, i.e., `222 222 222`. So wrong value will be saved in `srhit.t0`, which is used in the analysis (for CRTVeto). So I believe this should be included in the production..

jedori0228:
> So, in MC this doesn't matter (or at least can be worked around) because the gate_start_timestamp is just a constant. Right?
> But in data (i.e. only in icarus) all the CRT times in a particular event are offset by some constant that is impossible to determine from the CAFs themselves?

Yes, but even if we have `gate_start_timestamp` in our caf files, now `ts0_ns` only has sub-second time, while we need full time stamp to subtract the offset; so `ts0_ns` is changed to `ts0()`

PR #275, opened 2022-07-12
Updates to Calibration NTuple-r to Support Summer Production

gputnam:
Various configuration updates. Enables running the Calbiration ntuple-r with multiple T0 inputs, and with not requiring a T0 at all. Changes waveform window to +/-200 ticks. 

Depends on https://github.com/SBNSoftware/sbnobj/pull/58.


gputnam:
@miquelnebot this PR is intended for use in processing the data produced by ICARUS over the summer. So my understanding is that it needs to go on the production branch.

mrmooney (approved):
Looks good to me, from what I can tell.  The added functionality is indeed what we want, and it appears to be implemented correctly from a quick look.

PR #278, opened 2022-08-11
Updates to MeVPrtl Generator

gputnam:
Fix phi computation in RayTracer. Add in more variables, including equivalent neutrino energy. Depends on https://github.com/SBNSoftware/sbnanaobj/pull/71 and https://github.com/SBNSoftware/sbnobj/pull/60.


VCLanNguyen:
image
Plots after phi computation fix for on-axis beam.

VCLanNguyen (approved):

PR #281, opened 2022-08-24
Adding trigger information to CAF files to compute time within beam gate

jzettle:
Adding trigger information to CAF files in order to compute the time within the beam gate. Requires changes to create a new place for trigger information in order move the `sbn::ExtraTriggerInfo` into sbncode.

Tested on ICARUS data from raw -> CAF stage and successfully extracted the correct trigger information at each stage.

Depends on equivalent PRs in icaruscode, sbnanaobj, and sbnobj

### DIFF
--- a/sbncode/CAFMaker/CAFMaker_module.cc
+++ b/sbncode/CAFMaker/CAFMaker_module.cc
@@ -1003,12 +1003,8 @@ void CAFMaker::produce(art::Event& evt) noexcept {
   fSRTrigger.clear();
   if(isRealData)
   {
-    art::Handle<sbn::ExtraTriggerInfo> addltrig_handle;
-    evt.getByLabel(fParams.TriggerLabel(), addltrig_handle);
-    const sbn::ExtraTriggerInfo& addltrig = *addltrig_handle;
-    art::Handle<std::vector<raw::Trigger> > trig_handle;
-    evt.getByLabel(fParams.TriggerLabel(), trig_handle);
-    const std::vector<raw::Trigger>& trig = *trig_handle;
+    const auto& addltrig = evt.getProduct<sbn::ExtraTriggerInfo>(fParams.TriggerLabel());
+    const auto& trig = evt.getProduct<raw::Trigger>(fParams.TriggerLabel());
     if(trig.size()==1)
     {
       FillTrigger(addltrig, trig, fSRTrigger);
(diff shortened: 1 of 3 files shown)


jzettle:
Depends on icaruscode https://github.com/SBNSoftware/icaruscode/pull/435, sbnanaobj https://github.com/SBNSoftware/sbnanaobj/pull/73, sbnobj https://github.com/SBNSoftware/sbnobj/pull/61

jzettle:
Also if someone can test this set of PRs on an SBND workflow that would be great! I'm not sure who is the right person to direct that ask to.

sbncode/CAFMaker/CAFMaker_module.cc
```diff
@@ -996,6 +999,22 @@ void CAFMaker::produce(art::Event& evt) noexcept {
   // Fill detector & reco
   //#######################################################
 
+  //Beam gate and Trigger info
+  fSRTrigger.clear();
```
cjbackhouse, line 1003: Why does this need to be a member variable rather than just creating an SRTrigger object within the produce() function here?

cjbackhouse: Or indeed just fill straight into `rec.hdr.triggerinfo`?

cjbackhouse (approved):

PetrilloAtWork (approved):
* As it is written, `sbn::ExtraTriggerInfo` is a mandatory piece of information for data.
Is SBND ok with that? ICARUS tried to shape `sbn::ExtraTriggerInfo` so that the quantities more or less make sense for SBND too, but it still clearly shows where it comes from. Anyway this is not urgent, since it just needs to be addressed before SBND starts getting data.
* Do you or @cjbackhouse think it would be useful to add also the beam gate duration? Or is it too uninteresting to have event by event?

Neither of these questions are stopping the merge of this code, though.

fjnicolas:
> @fjnicolas could you look for someone to test this as @jzettle suggested,please?

Tha CAF making stage looks good to me in the SBND CI. I'm happy with the merge from the SBND side.

jzettle:
@miquelnebot Yes, sorry for the delay I will try to do this today. I am at a conference this week

sbncode/CAFMaker/CAFMaker_module.cc
```diff
@@ -996,6 +999,22 @@ void CAFMaker::produce(art::Event& evt) noexcept {
...
+  if(isRealData)
+  {
+    art::Handle<sbn::ExtraTriggerInfo> addltrig_handle;
+    evt.getByLabel(fParams.TriggerLabel(), addltrig_handle);
+    const sbn::ExtraTriggerInfo& addltrig = *addltrig_handle;
```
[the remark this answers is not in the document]
jzettle: Done

jzettle:
I have addressed the comments on the PR

PR #283, opened 2022-09-06
Feature/jlarkin nuefakereco

jacoblarkin:
Add fake reco for beam induced nue events (no dirt or cosmics) following the selection in section 4.3 of the oscillation tech note (docdb 27037) and the energy smearing in the [code used for the old oscillation samples](https://github.com/SBNSoftware/sbncode/tree/feature/DomBarkerNueSelection/sbnanalysis/ana/SBNOsc) (specifically [here](https://github.com/SBNSoftware/sbncode/blob/d23369eab5567d1cec7b642f99ddce9cf6bd1245/sbnanalysis/ana/SBNOsc/Utilities.cxx#L554) for the lepton and [here](https://github.com/SBNSoftware/sbncode/blob/d23369eab5567d1cec7b642f99ddce9cf6bd1245/sbnanalysis/ana/SBNOsc/Utilities.cxx#L241) for hadrons).
I decided to rewrite the selection instead of copying from that old branch since it was difficult to follow what was going on there, but the tech note didn't specify an energy smearing so I did copy that from the old branch.

This is needed for the oscillation analysis until we have actual reco for nue events.

Not sure who the right people are to review this, so I'm just putting Chris for now.
Chris or others should feel free to add reviewers if there are other people who should look at this.

### DIFF
--- a/sbncode/CAFMaker/CAFMaker_module.cc
+++ b/sbncode/CAFMaker/CAFMaker_module.cc
@@ -983,7 +983,7 @@ void CAFMaker::produce(art::Event& evt) noexcept {
   }
 
   std::vector<caf::SRFakeReco> srfakereco;
-  FillFakeReco(mctruths, mctracks, fActiveVolumes, *fFakeRecoTRandom, srfakereco);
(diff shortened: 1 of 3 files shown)


cjbackhouse (approved):
Looks fine to me.

PR #287, opened 2022-09-09
bug fixes so it works with blinding off

etworcester:
Needed to be more tidy with file handling in case of blinding off


jacoblarkin (approved):
I also confirmed this works by running a test job with blinding turned off.

PR #295, opened 2022-09-21
CRUMBS Multi Signal

henrylay97:
Retains previous CRUMBS functionality but adds separate scores trained specifically for CCNuMu, CCNuE and NC signals. Also includes a `bestscore` and `bestid` field to use the score that gave the most neutrino-like result of the three.

This is the central PR but it also relies on:
- SBNSoftware/sbnanaobj#76
- SBNSoftware/sbnobj#66
- SBNSoftware/sbndcode#302

Have left all in draft mode until we have a new version of `sbnd_data` containing the new weight files (otherwise the CI will scream) but it is otherwise ready for review.


henrylay97:
trigger build SBNSoftware/sbnanaobj#76 SBNSoftware/sbnobj#66 SBNSoftware/sbndcode#302

etyley (approved):
Looks good to me generally. The three new training scripts are almost identical so I would be tempted to make a single script where one can specify the name and cut to be placed for the specific signal to reduce code duplication. This would also have the advantage of being able to retrain all BDTs at the same time more easily, as I can't imagine a scenario in which only one would be updated.

PR #297, opened 2022-09-29
Feature/etw icarusblinding

etworcester:
Two things - this fixes yet another bug in the output files - this time the flat cafs were corrupted, which has been fixed with this update. The second thing is to fill a flag in the header which labels whether a file is blinded/prescaled or not. The second thing depends on sbnanaobj PR #78, which adds that flag to SRHeader.


jacoblarkin (approved):
Look good. Tested on ICARUS data and it appears to be working.

wesketchum:
@cerati and I marked ready for release on this, but we should verify if these CI tests should be passing/what the issue is there

etworcester:
trigger build https://github.com/SBNSoftware/sbnanaobj/pull/78

etworcester:
@miquelnebot the CI test doesn't seem to be picking up the sbnanaobj PR correctly - both Jacob and I confirmed that we can build on gpvms, but the CI build fails complaining about a variable that doesn't exist, which is exactly the variable added in the sbnanobj PR. Any suggestions?

PR #298, opened 2022-09-30
Adding offbeam gate counting into the CAFs

jzettle:
This PR adds offbeam gate counting into the CAFs. The number of gates is computed as a difference with the gate counting performed within the event to compute the number of offbeam gates for BNB or NuMI seen by the trigger hardware on an event-by-event basis.


jzettle:
This is work largely done a month ago but I missed making the PR then. This is a set of PRs that spans https://github.com/SBNSoftware/sbnana/pull/89, https://github.com/SBNSoftware/sbnobj/pull/67, and https://github.com/SBNSoftware/sbnanaobj/pull/79

jzennamo:
Big picture, it looks like you have added the EXT accounting to `sbncode/BeamSpillInfoRetriever/BNBRetriever/BNBRetriever_module.cc` is that true? If not could you describe what you're changing in each file? Thanks!

jzettle:
@jzennamo Sorry for the delay. Yeah the latest commit from me cleans that up that is the way I had it initially and then realized is was better to use a separate module. That should not be the case and is not in my working version. Most of the details here are in CAFMaker that adds a connection to the EXTRetriever module which fills an object like the BNB/NuMIRetriever which the number of offbeam gates. The other PR in the set of 4 add the infrastructure needed to support this. I've tested this on the data and it appears to work.

jzettle:
@miquelnebot Will do ASAP. Took a quick look now and I need to understand better what the error means

jzettle:
@miquelnebot I see you merged the branch in this repo (sbncode) with develop before running the trigger build command. I do not see that done in the other repos. This branch at the time you performed that action is based on an older version than develop (v09_59_00) Is that causing a mismatch somewhere when building?

jzettle:
I honestly have no idea, I thought I resolved this with all the commits I am able to build and run this in my own area with no problems and git did not tell me I had anymore conflicts after I did this last round of commits. So I don't understand where the issue is with git here.

jzennamo (approved):

jzennamo:
Looks good!

sbncode/CAFMaker/CAFMaker_module.cc
```diff
@@ -1818,10 +1834,14 @@ void CAFMaker::produce(art::Event& evt) noexcept {
   rec.hdr.fno     = fFileNumber;
   if(fFirstInFile)
   {
+    rec.hdr.pot   = fSubRunPOT;
```
PetrilloAtWork, line 1837: Already present at the end of this scope.
```suggestion
```

PetrilloAtWork (commented):
Some comments documenting the meaning of the variables would help.
No shining issue, but for once I yield to @jzennamo (and maybe also Gray?) for design comments.

jzettle:
@miquelnebot Sorry a lot going on, will try to do today. Watch this space

jzettle:
Hi Miquel,

I can update this I am sorry I had a lot of other things to do in the past week or tow. There was a thought to overhaul this to use the decoded data products instead of the raw data. I want to understand if this is an option that is realizable on a short timescale or whether it is best to implement this later.

Thank you,

Jacob
________________________________

PR #305, opened 2022-10-19
Fix NuMI Spill accounting

gputnam:
Fixes a bug where the BeamFolder cache would not be properly filled with spills between two triggers. The old code was:

```
    try{auto cur_vec_temp = bfp->GetNamedVector((t_previous_event)-fTimePad,"E:HP121[]");} catch (WebAPIException &we) {}
    try{auto cur_vec_temp_2 = bfp->GetNamedVector((t_current_event)+fTimePad,"E:VP121[]");} catch (WebAPIException &we) {}
    try{auto packed_MTGTDS_temp = bfp->GetNamedVector((t_current_event)+fTimePad, "E:MTGTDS[]");} catch(WebAPIException &we) {}
```

The first line of code would invalidate the cache if it did not include the time of the most recent spill. Then the second line of code would fill the cache from the previous event to the current event. The final line of code created the bug: it would sometimes re-set the cache to include spills after the current event. Note that the purpose of these function calls (resetting the cache) was only a side effect of the actual function. Also, this would fail if the time between events was larger than the fcl configured `TimeWindow` (although in practice this seems to not have happened in Run 1).

I've changed this code to:
```
    bfp->FillCache(t_current_event + fTimePad);
    bfp->FillCache(t_previous_event - fTimePad);
```

This fixes the bug. I've also directly switched to just calling the `FillCache` function instead of using the `GetNamedVector` function to call `FillCache` as a side effect. This code is still a little obscure, but it seems like the best we can do with the provided `BeamFolder` interface.

I've also added in some code which adaptively resets the `BeamFolder` if the time between two events is larger than the time between triggers.

### DIFF
--- a/sbncode/BeamSpillInfoRetriever/NuMIRetriever/NuMIRetriever_module.cc
+++ b/sbncode/BeamSpillInfoRetriever/NuMIRetriever/NuMIRetriever_module.cc
@@ -55,14 +55,27 @@ private:
   std::vector< sbn::NuMISpillInfo > fOutbeamInfos;
   double fTimePad;
   double fBFPEpsilion;
-  std::string fURL;
   //MWRData mwrdata;
   std::string raw_data_label_;
   std::string fDeviceUsedForTiming;
   int TotalBeamSpills;
   art::ServiceHandle<ifbeam_ns::IFBeam> ifbeam_handle;
   std::unique_ptr<ifbeam_ns::BeamFolder> bfp;
+
+  // Beam Folder config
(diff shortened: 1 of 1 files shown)


gputnam:
@SFBayLaser @rennney this PR is necessary for Run 2 data processing.

gputnam:
@brucehoward-physics this is the bugfix -- following up on today's NuMI meeting.

PetrilloAtWork (approved):
Good to go as is.

Kind of a prime for me.

PR #308, opened 2022-10-24
Fix Calo Point Time and add it PFP T0

gputnam:
Depends on https://github.com/SBNSoftware/sbnanaobj/pull/84.

### DIFF
--- a/sbncode/CAFMaker/CAFMaker_module.cc
+++ b/sbncode/CAFMaker/CAFMaker_module.cc
@@ -1423,6 +1423,10 @@ void CAFMaker::produce(art::Event& evt) noexcept {
       FindManyPStrict<recob::Track>(fmPFPart, evt,
             fParams.RecoTrackLabel() + slice_tag_suff);
 
+    art::FindManyP<anab::T0> fmPFPT0 =
+      FindManyPStrict<anab::T0>(fmPFPart, evt,
+            fParams.PFParticleLabel() + slice_tag_suff);
+
     // make Ptr's to tracks for track -> other object associations
     std::vector<art::Ptr<recob::Track>> slcTracks;
     if (fmTrack.isValid()) {
@@ -1673,8 +1677,14 @@ void CAFMaker::produce(art::Event& evt) noexcept {
         FillTrackMCS(*thisTrack[0], trajectoryMCS, rec.reco.trk.back());
         FillTrackRangeP(*thisTrack[0], rangePs, rec.reco.trk.back());
 
+        art::Ptr<anab::T0> thisPFPT0;
+        if (fmPFPT0.isValid()) {
+          if (!fmPFPT0.at(iPart).empty()) {
+            thisPFPT0 = fmPFPT0.at(iPart).at(0);
+          }
+        }
         const larpandoraobj::PFParticleMetadata *pfpMeta = (fmPFPMeta.at(iPart).empty()) ? NULL : fmPFPMeta.at(iPart).at(0).get();
-        FillPFPVars(thisParticle, primary, pfpMeta, rec.reco.trk.back().pfp);
+        FillPFPVars(thisParticle, primary, pfpMeta, thisPFPT0, rec.reco.trk.back().pfp);
 
         if (fmChi2PID.isValid()) {
            FillTrackChi2PID(fmChi2PID.at(iPart), lar::providerFrom<geo::Geometry>(), rec.reco.trk.back());
@@ -1716,8 +1726,14 @@ void CAFMaker::produce(art::Event& evt) noexcept {
         rec.reco.shw.push_back(SRShower());
         FillShowerVars(*thisShower[0], vertex, fmShowerHit.at(iPart), lar::providerFrom<geo::Geometry>(), producer, rec.reco.shw.back());
 
+        art::Ptr<anab::T0> thisPFPT0;
+        if (fmPFPT0.isValid()) {
+          if (!fmPFPT0.at(iPart).empty()) {
+            thisPFPT0 = fmPFPT0.at(iPart).at(0);
+          }
+        }
         const larpandoraobj::PFParticleMetadata *pfpMeta = (iPart == fmPFPart.size()) ? NULL : fmPFPMeta.at(iPart).at(0).get();
-        FillPFPVars(thisParticle, primary, pfpMeta, rec.reco.shw.back().pfp);
+        FillPFPVars(thisParticle, primary, pfpMeta, thisPFPT0, rec.reco.shw.back().pfp);
 
         // We may have many residuals per shower depending on how many showers ar in the slice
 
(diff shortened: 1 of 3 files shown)


gputnam:
Added in another commit to make the CaloPoints usable.

sbncode/CAFMaker/FillReco.cxx
```diff
@@ -558,18 +558,19 @@ namespace caf
...
 
         // lookup the wire -- the Calorimery object makes this
         // __way__ harder than it should be
         for (const art::Ptr<recob::Hit> &h: hits) {
           if (h.key() == tps[i]) {
             p.wire = h->WireID().Wire;
+            p.tpc = h->WireID().TPC;
             p.sumadc = h->SummedADC();
             p.integral = h->Integral();
+            p.t = h->PeakTime();
```
brucehoward-physics, line 573: @grayputnam do we understand why the previous version wasn't looking as intended? I think it seems ok to use the hit `PeakTime()` but I guess I'm not sure *why* the previous version was not right.

(I guess this for loop is looping through all the hits to find the one which matches the calo point presumably because there is no art::Assn between calo point and hit? -- Just double checking I understand the purpose of this loop.)

gputnam: Yes there is no association.

And I'm not sure why the x point method wasn't working... presumably Pandora is doing stuff with x position that I do not understand.

brucehoward-physics: Pandora can change the X position for example if it stitches the track across the cathode, but I don't see why it would impact things in the way that Maria was seeing.

sbncode/CAFMaker/CAFMaker_module.cc
```diff
@@ -1673,8 +1677,14 @@ void CAFMaker::produce(art::Event& evt) noexcept {
         FillTrackMCS(*thisTrack[0], trajectoryMCS, rec.reco.trk.back());
         FillTrackRangeP(*thisTrack[0], rangePs, rec.reco.trk.back());
 
+        art::Ptr<anab::T0> thisPFPT0;
+        if (fmPFPT0.isValid()) {
+          if (!fmPFPT0.at(iPart).empty()) {
+            thisPFPT0 = fmPFPT0.at(iPart).at(0);
```
brucehoward-physics, line 1683: From the code conventions (and if I'm properly understanding what is returned here) I think this should be using a FindOneP or something that explicitly assumes there will only be one associated T0?

sbncode/CAFMaker/CAFMaker_module.cc
```diff
@@ -1716,8 +1726,14 @@ void CAFMaker::produce(art::Event& evt) noexcept {
         rec.reco.shw.push_back(SRShower());
         FillShowerVars(*thisShower[0], vertex, fmShowerHit.at(iPart), lar::providerFrom<geo::Geometry>(), producer, rec.reco.shw.back());
 
+        art::Ptr<anab::T0> thisPFPT0;
+        if (fmPFPT0.isValid()) {
+          if (!fmPFPT0.at(iPart).empty()) {
+            thisPFPT0 = fmPFPT0.at(iPart).at(0);
```
brucehoward-physics, line 1732: As commented above.

gputnam: Fixed.

sbncode/CAFMaker/CAFMaker_module.cc
gputnam: Fixed.

sbncode/CAFMaker/FillReco.cxx
gputnam: I don't either. But the time variable was always meant to encode the hit time. I was trying to be cute and get it out of Pandora. Regardless of whether that "should" work, directly accessing the hit time is correct.

sbncode/CAFMaker/FillReco.cxx
```diff
@@ -733,6 +735,9 @@ namespace caf
       CopyPropertyIfSet(propertiesMap, "LArThreeDPCAFeatureTool_TertiaryPCARatio",           srpfp.pfochar.pca3ratio);
       CopyPropertyIfSet(propertiesMap, "LArThreeDVertexDistanceFeatureTool_VertexDistance",  srpfp.pfochar.vtxdist);
     }
+    if (t0) {
```
brucehoward-physics, line 738: Do we want to set some value for if !t0 case? If I'm not mistaken it will otherwise be nan in the CAF? Maybe I'm missing where some default is set.

sbncode/CAFMaker/CAFMaker_module.cc
```diff
@@ -1713,8 +1721,12 @@ void CAFMaker::produce(art::Event& evt) noexcept {
         rec.reco.shw.push_back(SRShower());
         FillShowerVars(*thisShower[0], vertex, fmShowerHit.at(iPart), lar::providerFrom<geo::Geometry>(), producer, rec.reco.shw.back());
 
+        art::Ptr<anab::T0> thisPFPT0;
+        if (f1PFPT0.isValid()) {
+          thisPFPT0 = f1PFPT0.at(iPart);
+        }
```
brucehoward-physics, line 1727: I defer to others here because I'm definitely not proclaiming to be an expert in efficiency, but do we want to initialize this variable similarly to the pfpMeta below? (I.e. const and with a ternary.) Maybe the answer is no. Does @PetrilloAtWork have wisdom here?

sbncode/CAFMaker/CAFMaker_module.cc
```diff
@@ -1670,8 +1674,12 @@ void CAFMaker::produce(art::Event& evt) noexcept {
         FillTrackMCS(*thisTrack[0], trajectoryMCS, rec.reco.trk.back());
         FillTrackRangeP(*thisTrack[0], rangePs, rec.reco.trk.back());
 
+        art::Ptr<anab::T0> thisPFPT0;
+        if (f1PFPT0.isValid()) {
+          thisPFPT0 = f1PFPT0.at(iPart);
+        }
```
brucehoward-physics, line 1680: Same as other instance of this.

brucehoward-physics (approved):
I include a possible suggestion and a question:

1) I defer to others here because I'm not proclaiming to be an expert in efficiency, but do we want to initialize this variable similarly to the pfpMeta below? (I.e. const and with a ternary.) Maybe the answer is no. Does @PetrilloAtWork have wisdom here?

2) If I'm reading correctly I think that the t0 variable is left nan in the case that t0 is NULL? Not sure if this is intended or if we want to set some default (or if I've misread). (For neutrino candidate slices it may not create any t0? Or for other things that don't stitch. And then we'd have lots of nans.)

However, not sure any of these should hold up me hitting approve so I'll hit approve. I assume that @grayputnam has done any necessary CI/validation? I have not done this.

PR #321, opened 2022-12-01
Updated POT accounting scripts for new trigger fragment definition

jzettle:
This PR updates the POT accounting scripts to handle new information from an updated trigger fragment definition ICARUSTriggerV3 that was requested to split the trigger gate counting for majority and minbias streams. This code may get updated as it gets fully vetted and tested which I hope to complete tonight or at the latest tomorrow morning. The idea is this code is backwards compatible and should work with any data taken with ICARUSTriggerV3 or ICARUSTriggerV2 (the current implementation) but this piece need to be confirmed working. This PR is combined with ones on icaruscode and the offline branch of sbndaq-artdaq-core and I will update this once everything is in place and link to those. 

I understand this is important for Run2 data taking and I may need emergency help dealing with any version update issues as this is related to an older version of sbncode at this point but this has been a long project where I did not have a good opportunity to update this until it was done.


wesketchum (approved):
looks good to me

jzettle:
Okay @miquelnebot I have rebased the icaruscode PR also to the SBN2022B branch and resolved the conflict it had there. Thanks for your patience and your help

sbncode/BeamSpillInfoRetriever/NuMIRetriever/NuMIRetriever_module.cc
```diff
@@ -61,14 +63,21 @@ class sbn::NuMIRetriever : public art::EDProducer {
   int TotalBeamSpills;
   art::ServiceHandle<ifbeam_ns::IFBeam> ifbeam_handle;
   std::unique_ptr<ifbeam_ns::BeamFolder> bfp;
-
+  
+  struct TriggerInfo_t {
+    int gate_type = 0; ///< Source of the spill: `1`: BNB, `2`: NuMI
+    double t_current_event  = 0;
+    double t_previous_event = 0;
+    unsigned int number_of_gates_since_previous_event = 0; // FIXME needs to be integral type
```
PetrilloAtWork, line 71:
```cpp
    unsigned int number_of_gates_since_previous_event = 0; // FIXME needs to be integral type                        
```
What does the comment mean? this variable is already an integral type.

sbncode/BeamSpillInfoRetriever/NuMIRetriever/NuMIRetriever_module.cc
```diff
@@ -104,36 +162,51 @@ void sbn::NuMIRetriever::produce(art::Event &e)
...
+  {
+    auto const & raw_data = (*raw_data_ptr);
+    if(raw_data.size() > 1)
+    {
+      std::cout<<"More than one trigger fragment in this event, this should not happen! Skipping this event!"<<std::endl;
+	return;
+    }
+    for(auto raw_datum : raw_data){
+      uint64_t artdaq_ts = raw_datum.timestamp();
+      triggerInfo.t_current_event = static_cast<double>(artdaq_ts)/(1000000000.0);
```
PetrilloAtWork, line 182: Same comment as above, here and in the next `if` statement.

PetrilloAtWork (changes requested):
I think there is an issue with `t_current_event` in the code: please take a look at it.
I am not going to hammer on the form of the code because this is supposed to be temporary, while the final form should use `sbn::ExtraTriggerInfo`.

PR #323, opened 2022-12-13
Shifting CRT T0 and T1 properly in CAFMaker

jedori0228:
CRT T1 is a time reference to the trigger unlike T0 which is an absolute timestamp. When `CRTUseTS0` is set to false and we are saving `T1` in `CRTHit::time`, it should be shift by `+srtrigger.trigger_within_gate`

### DIFF
--- a/sbncode/CAFMaker/CAFMakerParams.h
+++ b/sbncode/CAFMaker/CAFMakerParams.h
@@ -329,11 +329,18 @@ namespace caf
       true
     };
 
-    Atom<bool> ReferenceCRTToBeam {
-      Name("ReferenceCRTToBeam"),
-      Comment("Whether to switch the reference time of CRT reco to the 'beam spill' time."),
+    Atom<bool> ReferenceCRTT0ToBeam {
+      Name("ReferenceCRTT0ToBeam"),
+      Comment("Whether to switch the reference time of CRT T0 reco to the 'beam spill' time."),
       true
     };
+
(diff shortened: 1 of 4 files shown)


francescopoppi (approved):

henrylay97:
Of course. It's not an area of the code I'm familiar with so apologies if my question is naive. 

Principle looks fine to me :heavy_check_mark: 

I appreciate that this PR purely allows for a different reference time. Its not clear to me where its ensured that MC doesn't have this correction applied? Given this function is applied regardless of whether the event is MC or data? Is the value carefully set elsewhere to avoid this problem?

jedori0228:
> @ jedori0228 I assume this goes with [SBNSoftware/sbnobj#72](https://github.com/SBNSoftware/sbnobj/pull/72)
> 
> And these are targeting production but not critical right now for a patch so they will be deferred to next production release early next year. Please say otherwise.

Yes, it only affects CAFMaker, and not critical for now.

jedori0228:
> 

Timings in MC is not understood fully yet (at least for me), and we are not doing this shifts for it; `ReferenceCRTToBeam()` can/should be set to false.

henrylay97:
So my worry is that it currently defaults to true, so we would need to actively set it to false in icaruscode & sbndcode fcls

https://github.com/SBNSoftware/sbncode/blob/develop/sbncode/CAFMaker/CAFMakerParams.h#L332-L335

PR #310, opened 2022-10-24
Feature/jskim dune syst fromv09 59 00

jedori0228:
(Reopening https://github.com/SBNSoftware/sbncode/pull/299, which was close by accident..)

This PR adds a module (named as `SystToolsEventWeight`) that retrieves systematic reweights from https://github.com/larsoft/systematicstools/ and https://github.com/larsoft/nusystematics/. Details can be found [doc-26425](https://sbn-docdb.fnal.gov/cgi-bin/sso/RetrieveFile?docid=26425&filename=220531__JaesungKim__SBN_Analysis_Infrastructure_Group_Meeting__Nusystematics.pdf&version=1) and [doc-28096](https://sbn-docdb.fnal.gov/cgi-bin/sso/RetrieveFile?docid=28096&filename=220928__JaesungKim__Event%20Selection%20and%20Analysis%20Task%20Force__Baseline%20Model.pdf&version=1).   
This PR requires https://github.com/SBNSoftware/sbnobj/pull/68.

Currently build fails due to the dependency of larcv in develop. Waiting for the solution.

PR #327, opened 2023-01-01
CMakeLists.txt fixed for SystToolsEventWeight

jedori0228:
CMakeLists.txt was not updated properly in https://github.com/SBNSoftware/sbncode/pull/310. Leave the `SBNEventWeight` as it was before, and adding SystToolsEventWeight to the cet plugin.


mroda88 (approved):
LGTM

PR #330, opened 2023-01-24
Save truth information in CaloPoints and add a few var

gputnam:
Depends on https://github.com/SBNSoftware/sbnanaobj/pull/90.


brucehoward-physics:
I don’t think I’m necessarily the best reviewer here… for one, I’m not really familiar with the SCE functions, nor am I expert on everything being done here. In lieu of a review. I offer a few thoughts/questions I had while giving it a read, but again I’m not sure I’m really the best person to give the ultimate go ahead… 

* These are some fairly substantive additions. First, thanks for them! Second, curious what level(s) of validation has been done? 

* I think this is grabbing all the IDEs which match to a given readout channel for the given true particle? How is this affected by particle trajectory itself, e.g. say the trajectory is fairly into (or out of) the plane such that you end up making a number of hits on the same wire from the same track? Maybe not a concern? 

* How will looping through trajectory points be affected by things like particles being rolled up and such [maybe namely as an example, do deltas get actually saved as separate particles or are they tacked on in some way to the muon?]. Again maybe this is not a concern especially if everything is separated, but figured worth asking. 

* Looking at the comment on line 190 (https://github.com/SBNSoftware/sbncode/blob/3df2f0399c2a8a7308157b76e2ae8f63c08f2487/sbncode/CAFMaker/FillTrue.cxx#L190): a doxygen page I was looking at claimed it was the ionization position. Is it a quirk of how SCE is added within the LArG4 step that this comment is true? 

* Line 232 (https://github.com/SBNSoftware/sbncode/blob/3df2f0399c2a8a7308157b76e2ae8f63c08f2487/sbncode/CAFMaker/FillTrue.cxx#L232) - I guess the 1e-4 is just meant to be some arbitrarily small number to check that it’s actually been assigned something?

gputnam:
To @brucehoward-physics 's comments:

1) This code is mostly taken from the truth matching in the Calibration N-tupler. So the validation basically comes from the operation of that module. 

2) Yes -- I think you'd just need to be careful about what the truth matching is telling you in that case.

3) Deltas would get rolled up into the main particle. I don't think though that the trajectory would usually be long enough for it to mess up the matching to the main particle trajectory.

4) The IDE's are after SCE

5) Yes i think so

Thanks for taking a look!

gputnam:
@miquelnebot it looks like the build failure is due to a versioning issue not related to this PR.

gputnam:
I've done a little more development on this code since I made the PR, so I just pushed another update. 

Still -- this PR is ready to go and should build. It works locally. When I checked the build failures here it looked like an issue unrelated to this PR.

jacoblarkin (approved):
Similar to Bruce I'm not sure I'm the best person to review the physics of these changes.

But the code looks good and I'll approve with the caveat that I'm not the right person to review the physics. If this was validated in the calibration ntuples maybe that doesn't matter, but that's not my call. It would be useful to get Gianluca's review as well.

gputnam:
@miquelnebot on the latest commit I forgot to push a corresponding commit to sbnanaobj -- that's my bad.

I just fixed this, it should build now.

gputnam:
trigger build https://github.com/SBNSoftware/sbnanaobj/pull/90.

gputnam:
trigger build https://github.com/SBNSoftware/sbnanaobj/pull/90

gputnam:
@miquelnebot when I check the build errors it doesn't seemed to have picked up the sbnanaobj PR -- did I do something wrong?

I also found an error in `mrbsetenv` in the build:

```125: ----------- check this block for errors -----------------------
126: Error encountered when setting up product: lardataobj
127: ERROR: Version conflict -- dependency tree requires versions conflicting with current setup of product lardataobj: version v09_13_04 vs v09_13_05
128: ERROR: setup -B lardataobj v09_13_04 -q c7:+prof failed
129: ERROR: setup of required products has failed```

File: sbncode/CAFMaker/FillTrue.cxx [hunk 1]
```diff
@@ -127,6 +130,190 @@ namespace caf {
...
+            traj_index = i_traj;
+          }
+        }
+
+        // residual range
+        truep.rr = 0;
+        if (traj_index >= 0) {
+          for (int i_traj = traj_index+1; i_traj < (int)particle.NumberTrajectoryPoints(); i_traj++) {
+            truep.rr += (particle.Position(i_traj).Vect() - particle.Position(i_traj-1).Vect()).Mag();
+          }
```
PetrilloAtWork, line 249: It may be guessed from my comments that I am a bit concerned of the time it takes to run this.
Here we are computing a residual range for each point of the track on each of the planes.
One thing that can be done to speed up _this_ part is to pre-compute the range as function of the trajectory point.
Something like:
```cpp
std::vector<double> getPartialRange(simb::MCTrajectory const& track) {
  if (track.empty()) return {};
  std::vector<double> ranges(track.size());
  std::size_t iPoint = track.size();
  ranges[--iPoint] = 0.0;
  while(iPoint-- > 0) {
    ranges[iPoint] = ranges[iPoint + 1] + (track.Position(iPoint+1) - track.Position(iPoint)).Vect().Mag();
  }
  return ranges;
} // getPartialRange()
```
should do (then `truep.rr = partialRanges[traj_index];`).

sbncode/CAFMaker/FillTrue.cxx
```diff
@@ -127,6 +130,190 @@ namespace caf {
...
+        }
+        else truep.rr = -1;
+
+        // pitch
+        //
+        // The true pitch is affected by the presence of space charge, so we cannot
+        // directly apply the true direction. We include the effect of space charge
+        // on the true pitch here
+        if (closest_dist >= 0. && direction.Mag() > 1e-4) {
+          geo::PlaneID plane(srtrack.producer, p.tpc, iplane);
```
PetrilloAtWork, line 263: Please move this at the beginning of the plane loop, but like the full object:
```cpp
geo::PlaneGeo const& plane = geo->Plane({ srtrack.producer, p.tpc, iplane });
```
then we can avoid all that nonsense about views.

sbncode/CAFMaker/FillTrue.cxx
```diff
@@ -127,6 +130,190 @@ namespace caf {
...
+          float angletovert = geo->WireAngleToVertical(geo->View(plane), plane) - 0.5*::util::pi<>();
+          TVector3 loc_mdx_v = loc_nosce_v - direction * (geo->WirePitch(geo->View(plane)) / 2.);
+          TVector3 loc_pdx_v = loc_nosce_v + direction * (geo->WirePitch(geo->View(plane)) / 2.);
+
+          // Convert types for helper functions
+          geo::Point_t loc_mdx(loc_mdx_v.X(), loc_mdx_v.Y(), loc_mdx_v.Z());
+          geo::Point_t loc_pdx(loc_pdx_v.X(), loc_pdx_v.Y(), loc_pdx_v.Z());
```
PetrilloAtWork, line 270: Assuming `direction` to be defined as `geo::Vector_t` instead of `TVector3` (you can use a `geo::vect::toVector()` on line 239, for example):
```suggestion
          geo::Point_t loc_mdx = loc_nosce - direction * plane.WirePitch() / 2.;
          geo::Point_t loc_pdx = loc_nosce + direction * plane.WirePitch() / 2.;
```
I pruned `angletovert` (it would have been `double const angletovert = plane.PhiZ();`) because it won't be needed, see below.

sbncode/CAFMaker/FillTrue.cxx
```diff
@@ -127,6 +130,190 @@ namespace caf {
...
+
+          // Map to wires
+          if (sce && sce->EnableSimSpatialSCE()) { 
+            int corr = geo->TPC(plane.TPC).DriftDir()[0];
```
PetrilloAtWork, line 274:
```suggestion
            int corr = -plane.GetNormalDirection().X();
```
(note that the normal of the plane points to the centre of the TPC, so it's opposite to the drift direction)

sbncode/CAFMaker/FillTrue.cxx
```diff
@@ -127,6 +130,190 @@ namespace caf {
...
+            offset_m.SetX(offset_m.X()*corr); // convert from drift direction to detector direction
+            loc_mdx = loc_mdx + offset_m;
+
+            geo::Vector_t offset_p = sce->GetPosOffsets(loc_pdx);
+            offset_p.SetX(offset_p.X()*corr); // convert from drift direction to detector direction
+            loc_pdx = loc_pdx + offset_p;
+          }
+
+          // Direction at wires
+          geo::Vector_t dir = (loc_pdx - loc_mdx) /  (loc_mdx - loc_pdx).r();
```
PetrilloAtWork, line 286:
```suggestion
          geo::Vector_t dir = (loc_pdx - loc_mdx).unit();
```
I recommend a name that does not get confused with the other variable, `direction`.

sbncode/CAFMaker/FillTrue.cxx
```diff
@@ -127,6 +130,190 @@ namespace caf {
...
+
+          // Pitch at wires
+          double cosgamma = std::abs(std::sin(angletovert)*dir.Y() + std::cos(angletovert)*dir.Z());
+          double pitch;
+          if (cosgamma) {
+            pitch = geo->WirePitch(geo->View(plane))/cosgamma;
```
PetrilloAtWork, line 292:
```suggestion
            pitch = plane.WirePitch()/cosgamma;
```

sbncode/CAFMaker/FillTrue.cxx
```diff
@@ -127,6 +130,190 @@ namespace caf {
...
+            geo::Vector_t offset_p = sce->GetPosOffsets(loc_pdx);
+            offset_p.SetX(offset_p.X()*corr); // convert from drift direction to detector direction
+            loc_pdx = loc_pdx + offset_p;
+          }
+
+          // Direction at wires
+          geo::Vector_t dir = (loc_pdx - loc_mdx) /  (loc_mdx - loc_pdx).r();
+
+          // Pitch at wires
+          double cosgamma = std::abs(std::sin(angletovert)*dir.Y() + std::cos(angletovert)*dir.Z());
```
PetrilloAtWork, line 289:
```suggestion
          double cosgamma = std::abs(plane.SinPhiZ()*dir.Y() + plane.CosPhiZ()*dir.Z());
```
Short explanation: `geo::WireAngleToVertical()` [is actually not the wire angle to vertical](https://cdcvs.fnal.gov/redmine/issues/27790), but it is the wire angle (equivalent to `geo::PlaneGeo::ThetaZ()`), and that is the reason while in the definition a few lines above in your code its value was rotated by a right angle. `geo::PlaneGeo` also provides the direction orthogonal to the wires (i.e. the direction that the wire plane measures), calling it `PhiZ()`, and since _z_ is "the horizontal" and _y_ is "the vertical", `PhiZ()` happens to be also the angle _of the wire_ from the vertical (barred possibly a sign, which in `PhiZ()` is defined as the direction where the coordinate increases).
Finally, the plane provides pre-computed sine and cosine.

sbncode/CAFMaker/FillTrue.cxx
```diff
@@ -127,6 +130,190 @@ namespace caf {
...
+          }
+          else {
+            pitch = 0.;
+          }
+
+          // Bring it back to particle trajectory
+          geo::Point_t loc_atwires_alongpitch = loc_sce + dir*pitch; 
+          geo::Point_t loc_alongpitch;
+          if (sce && sce->EnableSimSpatialSCE()) { 
+            geo::Vector_t offset = sce->GetCalPosOffsets(loc_atwires_alongpitch, plane.TPC);
```
PetrilloAtWork, line 302: With the changes above, this may become:
```suggestion
            geo::Vector_t offset = sce->GetCalPosOffsets(loc_atwires_alongpitch, p.TPC);
```
(or `plane.ID().TPC`)

PetrilloAtWork (commented):
My concern is how much computing time this adds.
I have recommended some changes that include:
* small optimisations to speed code up
* straighter use of the `Geometry`

Please take a look at them and consider whether to adopt them.
Also, please report on whether the CPU time increase is noticeable/negligible/dramatic.

sbncode/CAFMaker/FillTrue.cxx
gputnam: This can't be moved to the Plane loop because the TPC depends on the particle position.

sbncode/CAFMaker/FillTrue.cxx
gputnam: This code copies the interfaces used in the Calorimetry code for calculating pitch. I'd rather keep the use of interfaces consistent. I've been burned before by code behavior changing in LArSoft services and I'd rather reduce the number of different interfaces I use overall.

sbncode/CAFMaker/FillTrue.cxx
gputnam: See above comment.

sbncode/CAFMaker/FillTrue.cxx
```diff
@@ -127,6 +130,190 @@ namespace caf {
...
+        truep.z = 0.;
+
+        for (const sim::IDE *ide: ides) {
+          truep.x = (truep.x*truep.p_nelec + ide->x*ide->numElectrons) / (truep.p_nelec + ide->numElectrons);
+          truep.y = (truep.y*truep.p_nelec + ide->y*ide->numElectrons) / (truep.p_nelec + ide->numElectrons);
+          truep.z = (truep.z*truep.p_nelec + ide->z*ide->numElectrons) / (truep.p_nelec + ide->numElectrons);
+
+          truep.p_e += ide->energy;
+          truep.p_nelec += ide->numElectrons;
+        }
```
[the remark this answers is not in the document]
gputnam: I think the existing averaging code is fine. This doesn't change behavior and to me is a less intuitive way of computing an averge.

sbncode/CAFMaker/FillTrue.cxx
gputnam: See above comment. I agree that the existing code is kind of confusing. But it gives the correct result and is the same interface that the Calorimetry uses.

sbncode/CAFMaker/FillTrue.cxx
```diff
@@ -127,6 +130,190 @@ namespace caf {
...
+    // Get service
+    auto const* sce = lar::providerFrom<spacecharge::SpaceChargeService>();
+
+    // Load the hits
+    // match on the channel, which is unique
+    const std::vector<std::pair<geo::WireID, const sim::IDE*>> match_ides = id_to_ide_map.at(srtrack.truth.p.G4ID);
+    std::map<unsigned, std::vector<const sim::IDE*>> chan_2_ides;
+    for (auto const &ide_pair: match_ides) {
+      chan_2_ides[geo->PlaneWireToChannel(ide_pair.first)].push_back(ide_pair.second);
+    }
```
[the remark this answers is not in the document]
gputnam: ide_pair.second is a sim::IDE pointer, not a vector of pointers. So this wouldn't work.

gputnam:
I've addressed the comments above. Also, regarding performance concerns. I've added in a fcl parameter to save or not save the CaloTruth information, so it can be tuned by analyzers as necessary. 

Also, I've done a quick performance test on 15 events (nu+cosmic) with the latest updates:

Without saving truth information

```
TimeTracker printout (sec)                                       Min           Avg           Max         Median          RMS         nEvts
runprod:cafmaker:CAFMaker                                     0.364164       1.06376       2.15046       1.08088      0.408604        15
```

With Saving truth information:

```
TimeTracker printout (sec)                                       Min           Avg           Max         Median          RMS         nEvts
runprod:cafmaker:CAFMaker                                     0.710364       2.18523       4.16181       2.1704       0.834769        15
```

So, it doubles the time for CAFMaker. This is significant, but it's still only 2 seconds and is smaller than other processes run at the same stage (mcreco).

sbncode/CAFMaker/FillTrue.cxx
```diff
@@ -127,6 +130,190 @@ namespace caf {
...
+      if (p.TrackId() == srtrack.truth.p.G4ID) {
+        match = &p;
+        break;
+      }
+    }
+    if (!match) return;
+    const simb::MCParticle &particle = *match;
+
+    // Get service
+    auto const* sce = lar::providerFrom<spacecharge::SpaceChargeService>();
```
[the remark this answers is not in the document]
gputnam: Fixed.

sbncode/CAFMaker/FillTrue.cxx
```diff
@@ -127,6 +130,190 @@ namespace caf {
...
+
+    // Load the hits
+    // match on the channel, which is unique
+    const std::vector<std::pair<geo::WireID, const sim::IDE*>> match_ides = id_to_ide_map.at(srtrack.truth.p.G4ID);
```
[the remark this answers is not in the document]
gputnam: Fixed.

sbncode/CAFMaker/FillTrue.cxx
```diff
@@ -127,6 +130,190 @@ namespace caf {
...
+        // the Gaussian-fit-width does not replicate the width of the pulse. 
+        //
+        // Instead, we use the Hit (start, end) time range. This is also more relevant
+        // for (e.g.) the SummedADC charge extraction method.
+        //
+        // Don't use this:
+        // std::vector<sim::TrackIDE> ides = bt_serv->HitToTrackIDEs(clockData, hit);
+        //
+        // Use this:
+        art::ServiceHandle<cheat::BackTrackerService> bt_serv;
```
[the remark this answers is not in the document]
gputnam: Fixed.

sbncode/CAFMaker/FillTrue.cxx
```diff
@@ -127,6 +130,190 @@ namespace caf {
...
+        }
+
+        // Particle based truth matching
+
+        if (!chan_2_ides.count(p.channel)) { 
+          p.truth = truep;
+          continue; 
+        }
+
+        const std::vector<const sim::IDE *> &ides = chan_2_ides.at(p.channel);
```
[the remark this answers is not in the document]
gputnam: Can't do this

sbncode/CAFMaker/FillTrue.cxx
```diff
@@ -127,6 +130,190 @@ namespace caf {
...
+        TVector3 direction;
+        float closest_dist = -1.;
+        int traj_index = -1;
+        for (unsigned i_traj = 0; i_traj < particle.NumberTrajectoryPoints(); i_traj++) {
+          if (closest_dist < 0. || (particle.Position(i_traj).Vect() - loc_nosce_v).Mag() < closest_dist) {
+            direction = particle.Momentum(i_traj).Vect().Unit();
+            closest_dist = (particle.Position(i_traj).Vect() - loc_nosce_v).Mag();
+            traj_index = i_traj;
+          }
+        }
```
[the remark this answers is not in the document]
gputnam: I'd rather use the TVector's -- also see the performance results below.

sbncode/CAFMaker/FillTrue.cxx
gputnam: Fixed -- thanks for this!

sbncode/CAFMaker/FillTrue.cxx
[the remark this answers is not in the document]
PetrilloAtWork: It's also three times slower (counting multiplications).
I am a bit surprised you find that more intuitive, but I won't certainly argue about that.

sbncode/CAFMaker/FillTrue.cxx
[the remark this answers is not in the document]
PetrilloAtWork: True, my bad (I).

PetrilloAtWork (commented):
I have left a few follow up comments, but it's up to you now.
I appreciate the addition of the option not to add this truth information.
Whatever you decide, I would be interested in trying to profile this code after it's merged — so if you can paste the command line you used to test it, it will make profiling easier for me. While in absolute terms it's indeed invisible in comparison with other stages, if there is an easy way to halve the execution time, I would go for it (that's what I heard being called a "low-hanging fruit").

gputnam:
@PetrilloAtWork I fixed the Mag2 thing. 

On the interface comment, I appreciate the issue you are raising. I agree that the code you suggest would likely be an improvement. However, I don't think that the performance issues are a bottleneck and I have worries about the stability of the interface that you are proposing to use. I think that a possible issue to raise would be to switch over all of the relevant code in sbncode to using the improved interfaces for calculating the pitch. I would want that to be done all at once if we decided to.

PetrilloAtWork (approved):
Official Chrism of Approval.

gputnam:
Ok pushed an update that should fix the remaining build issues.

gputnam:
trigger build https://github.com/SBNSoftware/sbnanaobj/pull/90

gputnam:
oh looks like I did the build command wrong

gputnam:
trigger build SBNSoftware/sbnanaobj#90

PR #333, opened 2023-02-01
Add addtional 2-body and 3-body final states for HNL MeVPrtl

VCLanNguyen:


VCLanNguyen:
trigger build LArSoft/lar*@LARSOFT_SUITE_v09_66_00

rodralva (approved):
Great work @VCLanNguyen. Looks good to me too. As a side note, I think we can manage the couts  with mf::LogInfo instead (similar to the verbose idea).

PR #335, opened 2023-02-17
Luis bugfix

LuisPelegrina:
Fixed calculation of Branchig ratios and added nul1l2 decay width to the calculation of total decay width


gputnam (approved):
Thanks for these bugfixes!

PR #336, opened 2023-02-22
Update CAF Blinding Interface to Allow for Multiple Selections

gputnam:
Note -- this is not needed for the upcoming MCP. Depends on https://github.com/SBNSoftware/icaruscode/pull/537.


etworcester (approved):
Looks good to me. Did you check to make sure the output is identical?

gputnam:
I think because of the use of random numbers I can't check if the files are identical. But I have checked that the energy is blinded as expected, that the POT is offset as expected, and that the prescaled files contain a subset of the total outputs.

gputnam:
Found an issue with the POT -- please do not merge this PR quite yet.

gputnam:
I don't think this is actually a good idea now -- closing.

PR #332, opened 2023-01-27
Feature/gputnam mevprtl morehnl

gputnam:
Copy of https://github.com/SBNSoftware/sbncode/pull/307, now for develop.
Depends on SBNSoftware/sbnobj#75 SBNSoftware/sbnanaobj#92

PR #339, opened 2023-02-27
ALP Implementation

gputnam:
Implements an ALP simulation for SBN. Includes production from mixing with psuedoscalar muons and decays to 2gamma, 2muon, 2electron, 3pion, and 2pion + 1gamma. Includes a jupyter notebook for producing psuedoscalar meson fluxes using pythia.

Also -- includes a couple updates to resolve outstanding requests from https://github.com/SBNSoftware/sbncode/pull/332.


VCLanNguyen:
Looks good to me

VCLanNguyen (approved):

sbncode/EventGenerator/MeVPrtl/Tools/ALP/ThreeBodyIntegrator.h
```diff
@@ -0,0 +1,34 @@
...
+Class for conducting three-body integration
+
+Joshua Berger 01/22/2023
+
+*/
+
+#include <array>
+#include <functional>
+
+class ThreeBodyIntegrator {
```
mroda88, line 14: This interface is a bit dry. Can we add some comments? What are the masses? Why m122 (which I'm guess is m_{12}^2 (a la latex) is singled out?

sbncode/EventGenerator/MeVPrtl/Tools/TxtFileGen_tool.cc
```diff
@@ -251,19 +244,20 @@ simb::MCFlux TxtFileGen::GetNext() {
...
   // read the 4-vector off the line
-  double E, px, py, pz;
+  double E, px, py, pz, npot, wgt;
+  int pdgcode;
   std::stringstream l4vec(line);
-  l4vec >> px >> py >> pz >> E;
+  l4vec >> px >> py >> pz >> E >> pdgcode >> wgt >> npot;
 
-  std::cout << "Values: " << px << " " << py << " " << pz << " " << E << std::endl;
+  std::cout << "Values: " << px << " " << py << " " << pz << " " << E << " " << pdgcode << " " << wgt << std::endl;
```
mroda88, line 255: I see in some other cases we are getting rid of the cout. 
Shall we do the same here?

sbncode/EventGenerator/PythiaGen/setup.sh
```diff
@@ -0,0 +1,4 @@
+source /cvmfs/icarus.opensciencegrid.org/products/icarus/setup_icarus.sh
+setup pythia8 v8_3_05 -q e20:p392:prof
```
mroda88, line 2: This kind of follows an anti-pattern with the respect to other packages. 
Should we not add pythia in the UPS dependencies?

mroda88 (commented):
This is mostly an ICARUS development, so I don't know what I can do. 

I made a few comments that might improve the readability of the code.

sbncode/EventGenerator/PythiaGen/setup.sh
gputnam: I don't think so -- pythia is only needed to run this one notebook and is not needed to build the repository.

I would think it's better to avoid this coupling to an external package unless we need it.

sbncode/EventGenerator/MeVPrtl/Tools/TxtFileGen_tool.cc
gputnam: Fixed.

sbncode/EventGenerator/MeVPrtl/Tools/ALP/ThreeBodyIntegrator.h
gputnam: Fixed.

gputnam:
trigger build

PR #342, opened 2023-03-27
Update Supera for MCParticleLite fix

Temigo:
As title says, Supera has an update to fix a bug related to MCParticleLite that we would like to be merged quickly for Icarus' MC production release. Thank you!


SFBayLaser (approved):
I confirm that we need this update for the pending ICARUS production runs. As I read this, the changes here are simply to pick up the correct updated packages. Thanks!

PR #344, opened 2023-04-07
Update Supera for MCTruth

Temigo:
This updates Supera config to allow them to 

1. expect a list of MCTruth producers which is `[generator, cosmgen]` 
2. not crash if one of the producers is not found

This avoids having to swap the supera config depending on the input generators. It does assume the naming of `generator` and `cosmgen` which seems generic (in Icarus at least).

This is a last minute merge request for Icarus production... again... thank you!!

@SFBayLaser


SFBayLaser (approved):
Interesting... thought I just submitted a review but it seems to have vanished...
Anyway...
This update is good since it decreases the number of special case fhicl files needed to run the supera stage of the reconstruction chain. I am a bit worried about what actually happens when a data object is not in the input file, I'm not sure if you get an exception during look up -or- you can an invalid handle which can be tested as in the code. The latter makes sense as the approach to follow so I am approving.

SFBayLaser:
trigger build

SFBayLaser:
Ok, I have confirmed with the Oracle (sometimes known as Gianluca), this is the right approach so good to go.

PR #346, opened 2023-04-13
Added setup stage where normalization tools are used.

PetrilloAtWork:
Following PR LArSoft/larreco#54, the normalisation tools are supposed to undergo a "setup" stage, which one of ICARUS tools uses to extract the detector clocks data.

This stage was added in `GnocchiCalorimetry` module but not in two modules from `sbncode`.
Without it, for `NormalizeDriftSQLite` tool:
 * `debug` code will complain about a failed assertion (verified)
 * `prof` code will likely break with a memory access violation

Note that `NormalizeDrift` tool is not affected because it was not optimised like the SQLite twin was (probably just slipped under the radar because it was not included in the workflow).
I expect this issue to have manifested from `v09_71_00` on (when the tools were changed), but I am not aware of reports of failure with that version.

Recommended reviewers: @grayputnam, the original author of the code.

This is a commit on top of `develop` branch, but it needs to be merged in the production branch too.


gputnam (approved):
Looks good! Thanks for the fix.

PR #347, opened 2023-04-13
CAFMaker adopting the standard NuRandomService protocol.

PetrilloAtWork:
`CAFMaker` uses random number streams for some aspects of the truth-based reconstruction (random smearing) and to drive blinding.
These changes allow the module to transparently take advantage of all `NurandomService` features, and especially the `perEvent` policy. This implies that processing the same input file (or in fact, the same event, regardless where it comes from) will always use the same random seed, e.g. blinding the same events; that is true as long as the module label does not change, since [the seed algorithm](https://nusoft.fnal.gov/larsoft/doxsvn/html/classrndm_1_1details_1_1PerEventPolicy.html#a34b15d34515ed1f62e9ae42aa06d697c) uses it to create the seed.
Configuration parameters are available to allow to override the seed of each of the two random streams.

The ICARUS standard workflow already uses `perEvent` random policy, so it will immediately take advantage of the change.

### Suggested reviewers

* @etworcester, author of at least part of the code being hooked;
* @brucehoward-physics, who is hunting some random changes in CAF;
* @gputnam, who was helping in the hunt and anyway he must be bored so here is something to do (You are welcome.).


brucehoward-physics:
Hi - are there specific things it would be useful for me to check @PetrilloAtWork? I don't think it's expected that this PR would fix the randomness I was seeing before, right?

That said, running cafmaker on a stage1 data file I found from production (an off-beam minbias file), I did notice different events getting passed to the `Blind` cafs running the module twice. I didn't observe different events in two outputs of the same module with the PR.

In fact, `diff_cafs` points to just instances of this sort of difference between instances:
```rec.slc[0].reco.pfp[0].shw.producer differs: 367079456 vs 349336224```
 (for a majority of the records)

So I think from that standpoint things seem working. I didn't check any fake reco...
 
One side note, when testing, I had to bump up sbnanaobj to `v09_20_07`

PetrilloAtWork:
I thought I had replied, but obviously I did not...
That `producer` difference smells of unassigned variable, both before and after the change. It may be an issue, but one unrelated to this PR. The differences being limited to that variable is to me a test strong enough, and I appreciate it a lot. Thank you!

PetrilloAtWork:
I would encourage the reviewers to check this change: @etworcester, @brucehoward-physics, @gputnam.

gputnam (approved):
Hi @PetrilloAtWork, thanks for this! Has this been tested that it isn't just using the same random seed in every job? If that's the case, then I'm happy with this.

PetrilloAtWork:
I have updated the branch to cope with the new `NuRandomService` (sad) interface.
I have also taken a chance to verify (but on detector data only) that:
1. the only difference of output from the same input is the same as observed by @brucehoward-physics:
    ```
    rec.slc[0].reco.pfp[0].shw.producer differs: 2943979264 vs 3543318272
    rec.slc[0].reco.pfp[0].trk.bestplane differs: 86074688 vs 584688256
    ```
    always only the first slice, first particle. Given that the random streams are used to decide which event to send into the three blinding-related streams, this is a success;
2. running on the same input files (as above), the random seeds are also the same (request by @gputnam, checked from `randomseeds.log` output log)
3. running on different input files, the random seeds are different (same as previous)

I think this is ready for merge.

PR #354, opened 2023-06-16
OAA variable in FluxReaderAna

marcodeltutto:
This fixes a bug in calculating the off-axis angle in FluxReaderAna. Thanks @henrylay97 for spotting this!


henrylay97 (approved):
Looks good, thanks Marco!

PR #355, opened 2023-06-16
Add slice charge center and width to CAFMaker for production release

JackSmedley:
This is a duplicate of a PR #348, which was merged into v09_73_00, for the production branch. There are changes to the CAFMaker, and I update the sbnanaobj dependency from `v09_20_06` to `v09_20_07`.  Note that anything newer than `v09_20_06` will do.

### DIFF
--- a/sbncode/CAFMaker/CAFMaker_module.cc
+++ b/sbncode/CAFMaker/CAFMaker_module.cc
@@ -1593,6 +1593,7 @@ void CAFMaker::produce(art::Event& evt) noexcept {
     FillSliceFlashMatchA(fmatch, recslc);
     FillSliceVertex(vertex, recslc);
     FillSliceCRUMBS(slcCRUMBS, recslc);
+    FillSliceBarycenter(slcHits, slcSpacePoints, recslc);
 
     // select slice
     if (!SelectSlice(recslc, fParams.CutClearCosmic())) continue;
--- a/sbncode/CAFMaker/FillReco.cxx
+++ b/sbncode/CAFMaker/FillReco.cxx
@@ -384,6 +384,48 @@ namespace caf
     }
   }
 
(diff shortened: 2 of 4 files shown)


JackSmedley:
I won't pretend to have a good understanding of how the CI testing works, but am I understanding the logs and stuff correctly when I see that it's trying build `icaruscode v09_75_01` against this `sbncode`? It's not surprising to me that that would fail.

JackSmedley:
trigger build LArSoft/lar*@LARSOFT_SUITE_v09_72_00 SBNSoftware/sbndaq-artdaq-core@v1_06_00of0 SBNSoftware/icarus_signal_processing@v09_63_00_02 SBNSoftware/icarusalg@v09_72_00 SBNSoftware/icarusutil@v09_66_02 SBNSoftware/sbnanaobj@v09_20_07

PetrilloAtWork (changes requested):
@JackSmedley, please add the commits from the development branch here too.

JackSmedley:
@miquelnebot We do really want to get this into the next production patch. This builds just fine for me, on its own and when built with `icaruscode v09_72_00_03p02`, so I am not sure what the issue is. @PetrilloAtWork pointed out that my equivalent [PR in develop](https://github.com/SBNSoftware/sbncode/pull/348) had more commits, but the code here is actually identical. I still don't understand what the issue is, but looking at the logs it maybe has something to do with dependencies?

JackSmedley:
I guess this might be outdated now though and maybe I should make a new PR in 72_00_04? Looks like the newer patch has an updated sbnanaobj to `v09_20_06_01`, so maybe that would mean we need to make a new sbnanaobj patch as well.

etworcester:
@chilge do you have any insight here on the CI failures? Thanks!

PR #358, opened 2023-06-17
Fix CRT time references in CAFs for data and MC 

aheggest:
This PR fixes the CRT time references for both data and MC files. In general, we use the CRT Hit T0 variable for MC and CRT Hit T1 variable for data files. 
`CRT Hit T0: CRT Hit timestamp wrt beamgate timestamp`
`CRT Hit T1: CRT Hit timestamp wrt global trigger timestamp` 

For the CAFs, we can use the same time reference (wrt beamgate) for both data and MC CRT Times by shifting the data CRT Hit timestamp to be wrt the beamgate, this is done [here](https://github.com/SBNSoftware/sbncode/blob/develop/sbncode/CAFMaker/FillReco.cxx#L71), `where trigger_within_gate` is the difference between the global trigger and beam gate timestamps.

This PR: 
* Makes sure to use the value stored in fParams.CRTSimT0Offset() for the time reference for MC files. after adding the trigger emulation,  [these lines](https://github.com/SBNSoftware/sbncode/blob/develop/sbncode/CAFMaker/CAFMaker_module.cc#LL1268C3-L1273C1) don't seem to be doing whats intended. (comment on this offset needed)
   * Note: [ReferenceCRTT0ToBeam](https://github.com/SBNSoftware/sbncode/blob/develop/sbncode/CAFMaker/CAFMakerParams.h#LL344C16-L344C36) and [ReferenceCRTT1FromTriggerToBeam](https://github.com/SBNSoftware/sbncode/blob/develop/sbncode/CAFMaker/CAFMakerParams.h#LL350C16-L350C47) may no longer be needed? 
* Fixes a sign error when shifting the CRT Hit T1 time for data to be wrt the beamgate 

We should now have the `CRT Hit timestamp wrt beamgate timestamp` saved correctly in both data and MC files

### DIFF
--- a/sbncode/CAFMaker/CAFMaker_module.cc
+++ b/sbncode/CAFMaker/CAFMaker_module.cc
@@ -1288,8 +1288,11 @@ void CAFMaker::produce(art::Event& evt) noexcept {
   art::Handle<std::vector<sbn::crt::CRTHit>> crthits_handle;
   GetByLabelStrict(evt, fParams.CRTHitLabel(), crthits_handle);
   // fill into event
-  int64_t CRT_T0_reference_time = fParams.ReferenceCRTT0ToBeam() ? -srtrigger.beam_gate_time_abs : 0; // ns, signed
-  double CRT_T1_reference_time = fParams.ReferenceCRTT1FromTriggerToBeam() ? srtrigger.trigger_within_gate : 0.;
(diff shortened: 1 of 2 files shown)


jedori0228 (approved):
Looks good to me.

PR #361, opened 2023-06-21
CAFMaker fills the source information

PetrilloAtWork:
This branch fills the new data members of standard record header introduced by SBNSoftware/sbnanaobj#98.
The catch is that the module does not really know the position of the event in the input file, so it has to keep a local count, which will be skewed if a filter upstream is not feeding all the events of the file. Tough luck.

Matter of discussion and review!!

PR for ICARUS production branch coming after discussion.

### DIFF
--- a/sbncode/CAFMaker/CAFMaker_module.cc
+++ b/sbncode/CAFMaker/CAFMaker_module.cc
@@ -176,9 +176,11 @@ class CAFMaker : public art::EDProducer {
   std::string fFlatCafFilename;
   std::string fFlatCafBlindFilename;
   std::string fFlatCafPrescaleFilename;
+  
+  std::string fSourceFile;
 
   bool fFirstInSubRun;
-  bool fFirstInFile;
+  unsigned int fIndexInFile = SRHeader::NoSourceIndex;
   bool fFirstBlindInFile;
   bool fFirstPrescaleInFile;
   int fFileNumber;
@@ -233,6 +235,8 @@ class CAFMaker : public art::EDProducer {
 
   std::string DeriveFilename(const std::string& inname,
                              const std::string& ext) const;
+
+  static std::string Basename(const std::string& path);
 
   void AddEnvToFile(TFile* f);
   void AddMetadataToFile(TFile* f,
@@ -528,6 +532,15 @@ CAFMaker::~CAFMaker()
 }
 
 //......................................................................
+std::string CAFMaker::Basename(const std::string& path)
+{
+  // C++17: use filesystem library (Clang 7 still not compliant)
+  constexpr char sep = '/';
+  std::size_t const iSep = path.rfind(sep);
+  return (iSep == std::string::npos)? path: path.substr(iSep + 1);
+}
+
+//......................................................................
 std::string CAFMaker::DeriveFilename(const std::string& inname,
                                      const std::string& ext) const
 {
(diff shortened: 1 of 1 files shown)


etworcester (approved):
This will be helpful!

PetrilloAtWork:
I have run a single 150 event job and saw that the ROOT branches were correctly filled.
I did not try to exercise them into `cafe`.

jzennamo:
According to @jacoblarkin says this doesn't compile, I think that that needs to be fixed before we can move forward?

PetrilloAtWork:
The C.I. failed because it did not pick up SBNSoftware/sbnanaobj#98 (maybe a typo in the C.I. trigger line?).
@jzennamo: I may not be aware of the compilation issue you refer to.
The report I have received from @jacoblarkin is that structured CAF contain the new information, while flat ones appear not to, which is a mystery to me.

PetrilloAtWork:
Still a problem with flat CAF not picking the new data entries, I feel like I may be missing some magic.

etworcester:
sbnanaobj/StandardRecord/Flat/FlatRecord.h is autogenerated by sbnanaobj/StandardRecord/Flat/CMakeLists.txt - you should be able to check include/sbnanaobj/StandardRecord/Flat/FlatRecord.h in your local products area to make sure the changes are getting picked up into FlatRecord.h. Sorry if this is obvious - just my thought of the first thing to check.

PetrilloAtWork:
I am offended and outraged about being pointed out such an _obvious_ thing as `FlatRecord.h` being generated by Cmake.

On a different topic: I have discovered that Cmake generates some code for the flat trees (`FlatRecord.h` and others)!!
:flushed:
Well, the problem with Cmake generating code is that sometimes it does not go so smooth.
It all boils down to `pygccxml`, the Python module that `gen_srproxy` uses to generate the flat tree C++ code, really can't suffer GCC `string` header. If it finds it, it yields... exactly the error messages that are shown in the C.I. test build stage with GCC (not with Clang, which `pygccxml` is `string`-happy with).
I must have seen this error message, and probably I went back to the main branch, compiled that one (expecting an error), and after it mysteriously worked well going back to my branch (expecting an error) and it also mysteriously compiled correctly. The mystery is probably Cmake not attempting to recreate `FlatRecord.h` & co. — I have seen trouble with autogenerating code other times, so I just shrug.

So I need to find a solution. At the first order, that may even be just to have the source information just on the structured CAF and not on the flat ones. At this point, I don't even know how to do _that_.

PetrilloAtWork:
trigger build SBNSoftware/sbnanaobj#98

PetrilloAtWork:
Reviwers (@etworcester , @jacoblarkin , @jzennamo): there is a problem with using strings in flat CAF.
I don't know how to solve it, and I don't know if it _can_ be solved. The details are [in a previous post](https://github.com/SBNSoftware/sbncode/pull/361#issuecomment-1629866074).
I have a couple of workarounds in my mind, and they are both so horrible that I am not sure I would accept them myself in a review. So far I am biting it and excluding the offending code, and the result is that only structured CAF files will benefit from this new feature, while **flat CAF will not have the complete source information**.
I propose we move on like this. I have thrown in the meanwhile the question to some CAF experts who might have some good suggestion.

----

For the record, the two workarounds that I have in mind are:
1. Fill a hash of the name in the CAF tree, and have a separate tree/map/whatever which associates each hash to a full file name.
2. Graft a branch to the flat trees in `CAFMaker` module, managed separately from the `flat::Flat` infrastructure which fills the rest of the tree.

Either cases, I am not sure that the branch would be accessible with the cafe framework.

PetrilloAtWork:
I have implemented a workaround suggested by other CAFana users. Ugly, but temporary because apparently newer versions of Clang will solve the issue (reminder: we use Clang 7.0, the current version is Clang 16.0).
With this, flat CAF files get their source file string too.
I consider this feature complete and ready for merge.

PetrilloAtWork:
@jzennamo , @jacoblarkin: please take a look again and update your review accordingly.

jacoblarkin (approved):
Tested and it now fills in the information for flatcafs. Thanks Gianluca!

PR #362, opened 2023-06-21
POT and Gate Accounting for Updated ICARUS Trigger V3

jzennamo:


jzettle (commented):
From a quick read I don't see anything that is clearly incorrect. It sounds like this has been validated at some level.

jacoblarkin (approved):
Looks good. Tested with ICARUS run 2 BNB data and it seems to be working.

PetrilloAtWork:
Can you explain the workflow you envision to use this feature? And the one to make it into CAF, too?

PetrilloAtWork (commented):
Can you explain the workflow you envision to use this feature? And the one to make it into CAF, too?

jzennamo:
The expected workflow is currently envisioned to be the following:
0. On one grid node
1. Start with a stage0 file
2. Run one of the POT accounting fcls (run_numiinfo_sbn.fcl, run_bnbinfo_sbn.fcl, etc.) based on the stream
3. Run the standard Stage1 fcl 

This requires that each stream has one of these fcls run over it (which was true in the past too)

jzettle:
@jzennamo So this requires the raw data product to be present in stage0 which is currently is. If that ever changes (which maybe it never will at this point) then we may need to revisit that workflow which means in this method this module would run first as it did before the changes close to the new year. Not suggesting a change just pointing this out

PetrilloAtWork:
The problem is that this requires four different workflows for data. As far as I know, right now there is just one.

I am going to make a counter proposal.

PetrilloAtWork:
It turns out that the minimally invasive workaround I was thinking won't work.
My attempt was to have a single workflow including all four the PoT accounting modules, and a filter in front of each to actually execute each one only on events from its own trigger stream. Unfortunately _art_ appears not to support the concept of filtering on subrun end (only on events).

PetrilloAtWork:
> @jzennamo So this requires the raw data product to be present in stage0 which is currently is. If that ever changes (which maybe it never will at this point) then we may need to revisit that workflow which means in this method this module would run first as it did before the changes close to the new year. Not suggesting a change just pointing this out

This implementation is expected to be transitional. There is a data product that is conveniently packaged with all the information these modules need, which will make them robust to low level changes.
Also, I hope that we can merge these four modules into one producing the correct data each time.

jzennamo:
@jacoblarkin did some very nice validation of this module here:
https://sbn-docdb.fnal.gov/cgi-bin/sso/ShowDocument?docid=32106

PetrilloAtWork (approved):
Production team (and the Ivan who is not apparently in SBNSoftware) have been informed of the modifications required to the workflows and have confirmed this is feasible.

PR #363, opened 2023-06-23
Offbeam gate counting for CAFMaker

jacoblarkin:
Fill new variables in the CAFs which count the offbeam gates.

Depends on https://github.com/SBNSoftware/sbnanaobj/pull/99

Does not depend on Joseph's PR (https://github.com/SBNSoftware/sbncode/pull/362) explicitly, but will only fill the CAF variables for input files made using that code.

Tested on ICARUS run 2 numi and bnb offbeam data.

### DIFF
--- a/sbncode/CAFMaker/CAFMakerParams.h
+++ b/sbncode/CAFMaker/CAFMakerParams.h
@@ -95,6 +95,18 @@ namespace caf
       "numiinfo"
     };
 
+    Atom<string> OffbeamBNBCountDataLabel {
+      Name("OffbeamBNBCountDataLabel"),
+      Comment("Label of offbeam BNB EXT producer"),
+      "extinfo"
+    };
+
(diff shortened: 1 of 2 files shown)


PetrilloAtWork (approved):
Good enough for now.

I have left one comment for your consideration, plus the following: I would encapsulate the PoT code in its own function rather than leaving it loose in `beginSubRun()`.

I also don't particularly appreciate this pattern (which you are expanding rather than creating with this PR, so I don't charge it on you :smile:).
I think it would be better rather than trying all the possible data products, to pick just the right one. The only tricky part is that the information on which is the right one is in the event rather than the subrun (one can find it either in `sbn::ExtraTriggerInfo` or, more encrypted, in `std::vector<raw::Trigger>`). To simplify the workflow it is then also possible to have a single label for these data products (that's just an idea; I am not sure that's the best choice).

jzennamo (approved):

PR #364, opened 2023-06-28
Updating SystTool module for Dependent dials

jedori0228:
This PR will handle "dependent" dials properly.
For instance, now we can group dependent dials (e.g., FSI knobs) and throw multisim universes. An example output from UniverseOracle looks like below:
```
 72: GENIEReWeight_ICARUS_v1_multisim_FSI_pi_VariationResponse, type 0, 100 universes, adjusted parameters:
    FrAbs_pi
    FrCEx_pi
    FrInel_pi
    MFP_pi
```

### DIFF
--- a/sbncode/SBNEventWeight/App/SystToolsEventWeight_module.cc
+++ b/sbncode/SBNEventWeight/App/SystToolsEventWeight_module.cc
@@ -149,6 +149,8 @@ void SystToolsEventWeight::produce(art::Event& e) {
           systtools::SystParamHeader const &sph = fParamHeaderHelper.GetHeader( r.pid );
           std::string prettyName = sph.prettyName;
 
+          if(sph.isResponselessParam) continue;
+
           if(fDebugMode){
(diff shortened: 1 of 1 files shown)


jedori0228:
To handle "single universe" parameters introduced in https://github.com/SBNSoftware/icaruscode/pull/612, additional updates are made.
When we write single element in `<param_name>_variation_descriptor` in the nusyst configuration file
```
    VecFFCCQEshape_central_value: 0
    VecFFCCQEshape_variation_descriptor: "[1.0]"
```
`GenerateSystProviderConfig` will make `SystParamHeader:: paramVariations` empty, but `SystParamHeader::centralParamValue` is updated to `1.0` and `SystParamHeader::isCorrection` is set to `true`.

jedori0228:
@miquelnebot This PR should be merged to follow the updates made in https://github.com/SBNSoftware/icaruscode/pull/612.

PR #367, opened 2023-07-12
opt0finder objects in cafmaker

lynnt-uchicago:
Moving OpT0 variables into the cafs. There are _two_ caf variables created:  `slc.opt0` and `slc.opt0_sec`, where the latter is the "secondary" OpT0 object. 

OpT0Finder (in SBND for one-to-many matching) can output more than one match per slice if the slice is split up between two TPCs (matches are TPC specific) or if there are multiple in-time flashes. The default `slc.opt0` is filled with the OpT0 object with the **highest** OpT0 score (best match), and `slc.opt0_sec` is filled with the OpT0 object with the **second highest** OpT0 score. 

For flash-matching, looking at `slc.opt0` would be sufficient. For Q->L charge-likelihood studies, one would likely want information from both TPCs (so look at both `slc.opt0` and `slc.opt0_sec`). 

Dependent on accompanying PR in sbnanaobj [#103](https://github.com/SBNSoftware/sbnanaobj/pull/103) and sbnobj [#90](https://github.com/SBNSoftware/sbnobj/pull/90).

### DIFF
--- a/sbncode/CAFMaker/FillReco.cxx
+++ b/sbncode/CAFMaker/FillReco.cxx
@@ -419,6 +419,54 @@ namespace caf
     }
   }
 
+  void FillSliceOpT0Finder(const std::vector<art::Ptr<sbn::OpT0Finder>> &opt0_v,
+                           caf::SRSlice &slice)
+  {
+    if (opt0_v.empty()==false){
+      unsigned int nopt0 = opt0_v.size();
+      double max_score=-1.; // score of the opt0 object with the highest score 
+      double sec_score=-1.; // score of the opt0 object with the 2nd highest score 
+
+      unsigned int max_idx = 0;
+      unsigned int sec_idx = 0;
+
+      // fill the default, which is the maximum 
+      for (unsigned int i = 0; i < nopt0; i++ ) {
+        const sbn::OpT0Finder &thisOpT0 = *opt0_v[i];
+        if (thisOpT0.score > max_score){
+          max_score = thisOpT0.score;
+          max_idx = i;
+        }
+      }
+
+      const sbn::OpT0Finder &maxOpT0 = *opt0_v[max_idx];
+      slice.opt0.tpc    = maxOpT0.tpc;
+      slice.opt0.time   = maxOpT0.time;
+      slice.opt0.score  = maxOpT0.score;
+      slice.opt0.measPE = maxOpT0.measPE;
+      slice.opt0.hypoPE = maxOpT0.hypoPE;
+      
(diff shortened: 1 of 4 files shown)


sbncode/CAFMaker/FillReco.cxx
```diff
@@ -419,6 +419,54 @@ namespace caf
...
+      unsigned int sec_idx = 0;
+
+      // fill the default, which is the maximum 
+      for (unsigned int i = 0; i < nopt0; i++ ) {
+        const sbn::OpT0Finder &thisOpT0 = *opt0_v[i];
+        if (thisOpT0.score > max_score){
+          max_score = thisOpT0.score;
+          max_idx = i;
+        }
+      }
```
henrylay97, line 440: Could do a `std::sort` function here just directly on the vector? Then use entries 0 and 1?

sbncode/CAFMaker/FillReco.cxx
```diff
@@ -419,6 +419,54 @@ namespace caf
...
+      // usually this is filled for a slice that is split across two tpcs
+      if (nopt0>1){    
+        for (unsigned int i = 0; i < nopt0; i++ ) {
+          if (i == max_idx) continue;
+          const sbn::OpT0Finder &thisOpT0 = *opt0_v[i];
+          if (thisOpT0.score > sec_score){
+            sec_score = thisOpT0.score;
+            sec_idx = i;
+          }
+        }
```
henrylay97, line 459: Previous comment would also avoid doing multiple loops of the vector.

henrylay97 (approved):
Generally looks great. Have added one comment which would slimline the filling procedure a little but its picky, you can freely chose to ignore it.

sbncode/CAFMaker/FillReco.cxx
lynnt-uchicago: this may be a stupid c++ question, but it's a `std::vector<sbn::OpT0Finder>` objects, rather than a vector of OpT0 scores... so I think that I can't directly sort the vector?

lynnt-uchicago: i forgot you can give a function!! I will try this!

henrylay97: Yeah exactly, something along the lines of 

`std::sort(opt0_v.begin(), opt0_v.end(),`
`          [](auto const &a, auto const &b)`
`          { return a->score > b->score; });`

You'll need to make the vector not `const` but then you can just query entry 0 and then entry 1 in the next if statement.

PR #368, opened 2023-07-18
Fix segfault in sbn::TrackStoppingChi2Alg::RunFit()

sjgardiner:
This occurs when the dE/dx information is empty. The solution implements proper handling of conditions where there are two or fewer points.


henrylay97 (approved):
Looks good to me @sjgardiner , thanks for fixing this.

I assume the RunFit function knows how to cope with empty vectors right? We're not just passing the problem?

henrylay97:
> I assume the RunFit function knows how to cope with empty vectors right? We're not just passing the problem?
I stopped being lazy and answered my own question.

Yes this is fine the size is checked [here](https://github.com/SBNSoftware/sbncode/blob/develop/sbncode/LArRecoProducer/TrackStoppingChi2Alg.cc#L19-L20) and is compared to a value set [here](https://github.com/SBNSoftware/sbncode/blob/develop/sbncode/LArRecoProducer/trackstoppingchi2fitter.fcl#L10) (currently has to be size 10+).

PR #369, opened 2023-07-19
Catch Bad Slices

henrylay97:
Occasionally one of the variables pandora calculates to describe slices for the SliceID BDT fails. In this scenario most of the metadata isn't filled and the default value (-1 for SBND) is set for the slice ID score. We can use this as a flag for "bad slices" and we should be treating these as clear cosmics anyway (more accurately _not treating them as neutrino slices_).

This PR ensures we check for that default value before this labelling the slices.


sjgardiner (approved):
Looks great, thanks for the quick work!

PR #377, opened 2023-07-31
Fix units for flashGateTime to be in microseconds

aheggest:
I was accidentally diving by 1e3 twice trying to convert from ns to µs, so flashGateTime was accidentally being saved in ms, but it should be in µs for the CAFs.


jedori0228 (approved):
I have checked `sbn::crt::CRTPMTMatching::flashGateTime` is already in µs, so no need to convert it again in FillReco. Looks good to me.

PR #378, opened 2023-08-01
Fix information saved in SRCRTPMTMatch 

aheggest:
This pull requests updates the information that is saved into the SRCRTPMTMatch object in the CAFs.
* Fix flashGateTime to be in µs (before I was accidentally diving by 1e3 twice trying to convert from ns to µs, so the resulting flashGateTime saved was in ms)
* Fill the Matched CRT Hit X,Y,Z positions in the CAFs (I simply forgot to add the line that actually fills the variable)

### DIFF
--- a/sbncode/CAFMaker/FillReco.cxx
+++ b/sbncode/CAFMaker/FillReco.cxx
@@ -118,10 +118,9 @@ namespace caf
     // allowEmpty does not (yet) matter here                                                           
     (void) allowEmpty;
     //srmatch.setDefault();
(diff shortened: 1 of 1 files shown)


jedori0228 (approved):
I have checked sbn::crt::CRTPMTMatching::flashGateTime is already in µs, so no need to convert it again in FillReco. Adding matched CRT position part looks also fine.

aheggest:
hi @miquelnebot, I will make a new develop PR! this develop PR will only have one of the modifications i mentioned above (filling the  Matched CRT Hit X,Y,Z positions in the CAFs, the other change is not needed for develop as there was not the same units issue)

PR #379, opened 2023-08-01
Changes to allow g4info reducer to use track ID instead of just origT…

bear-is-asleep:
…rackID through a fcl param. These changes are necessary to store the proper trackID for SBND

### DIFF
--- a/sbncode/LArG4/G4InfoReducer_module.cc
+++ b/sbncode/LArG4/G4InfoReducer_module.cc
@@ -51,6 +51,7 @@ private:
   art::InputTag fSedLabel; ///< module making the SimEnergyDeposit
   double fMinX, fMinY, fMinZ; ///< bottom left coordinate of union of all TPC active volumes
   double fVoxelSizeX, fVoxelSizeY, fVoxelSizeZ; ///< size of a voxel (cm)
+  bool fUseOrigTrackID; //Use orig track ID boolean
   //services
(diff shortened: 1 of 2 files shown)


sbncode/LArG4/G4InfoReducer_module.cc
```diff
@@ -131,20 +139,37 @@ void G4InfoReducer::produce(art::Event& e)
 
   for (size_t idx = 0; idx < sed_v.size(); ++idx) {
     auto const& sed = sed_v[idx];
+    // std::cout << "G4info - orig track ID : " << sed.OrigTrackID() 
+    // << " trackID : " << sed.TrackID()
+    // << " useOrigID : " << fUseOrigTrackID
+    // << std::endl;
```
fjnicolas, line 145: clean?

andrewmogan (approved):

fjnicolas (commented):
Looks good! Default behaviour of the G4InfoReducer should not be altered by this PR. @bear-is-asleep just for curiosity, what's the use you have in mind for this new feature?

andrewmogan:
@fjnicolas Since Bear is on vacation, I'll respond. Basically, this is a temporary fix to our labeling procedure for the machine learning reconstruction. The track ancestor IDs are not filled correctly in the SimEnergyDepositLite at the moment, and this breaks our ML training. So, for now, we use the original trackID instead.

PR #381, opened 2023-08-04
fill Matched CRT Hit positions into SRCRTPMTMatch in CAFs

aheggest:
This PR fills the Matched CRT Hit X,Y,Z positions in the CAFs (I simply forgot to add the line that actually fills the variable). This is close to a duplicate of PR #378, apart from the fix in that PR to the flashGateTime, as this was already being saved in µs in develop.


jedori0228 (approved):
Already approved at https://github.com/SBNSoftware/sbncode/pull/378

PR #384, opened 2023-09-15
exclude region feature

wjdanswjddl:
include a configurable option (`excludeRegions`) to exclude certain regions of the TPC in MCS calculation. 
- feature was added in a way that introduces minimal change to the existing code. Algorithm skips any breakpoints in defined bad regions, as it previously did for bad `segradlengths` values.
- excluded regions can be defined as volume boxes, and should be passed in the format of one list. Code will slice up the list in blocks of 6 numbers, and treat each block as values for boundaries of the region to be exluded: [x_low, x_high, y_low, y_high, z_low, z_high]. Code will check that the passed list is a multiple of 6, and if it is not, will log an error and not exclude any regions
- adding a fcl for ICARUS, `mcsproducer_icarus.fcl`, that uses an angular resolution value of 10mrad (default: 3mrad), uses Bethe-Bloch energy loss mode (default: Landau MPV), and defines exclude regions around the cathode (default: none)

### DIFF
--- a/sbncode/LArRecoProducer/LArReco/TrajectoryMCSFitter.cxx
+++ b/sbncode/LArRecoProducer/LArReco/TrajectoryMCSFitter.cxx
@@ -15,7 +15,8 @@ recob::MCSFitResult TrajectoryMCSFitter::fitMcs(const recob::TrackTrajectory& tr
   vector<size_t> breakpoints;
   vector<float> segradlengths;
   vector<float> cumseglens;
-  breakTrajInSegments(traj, breakpoints, segradlengths, cumseglens);
+  vector<bool> breakpointsgood;
+  breakTrajInSegments(traj, breakpoints, segradlengths, cumseglens, breakpointsgood);
   //
   // Fit segment directions, and get 3D angles between them
   //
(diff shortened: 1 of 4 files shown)


gputnam (approved):

sbncode/LArRecoProducer/LArReco/TrajectoryMCSFitter.cxx
```diff
@@ -88,7 +106,10 @@ void TrajectoryMCSFitter::breakTrajInSegments(const recob::TrackTrajectory& traj
   }
   //then add last segment
   if (thislen>0.) {
+    auto endpointpos = traj.LocationAtPoint(nextValid);
+    bool endpointposgood = isInVolume(fiducialVolumes, endpointpos) && !isInVolume(excludeVolumes, endpointpos);
     breakpoints.push_back(traj.LastValidPoint()+1);
+    breakpointsgood.push_back(endpointposgood);
```
PetrilloAtWork, line 112: Question: isn't `nextValid` here always an `recob::TrackTrajectory::InvalidIndex`?
The behaviour of `LocationAtPoint()` on an `InvalidIndex` is undefined, so we should make sure we are not triggering it.

PR #388, opened 2023-10-05
Cluster3D for SBN (works for sbnd and icarus)

bear-is-asleep:
Copied code from [Cluster3D in icaruscode](https://github.com/SBNSoftware/icaruscode/tree/develop/icaruscode/TPC/Tracking/cluster3D) to sbncode to be useable by both sbnd and icarus.

### DIFF
--- a/sbncode/Cluster3D/CMakeLists.txt
+++ b/sbncode/Cluster3D/CMakeLists.txt
@@ -1,0 +1,65 @@
+
+# where should the scripts/..xml file be installed?  Perhaps in bin?
+
+art_make_library(
+        LIBRARIES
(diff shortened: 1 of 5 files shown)


bear-is-asleep:
@SFBayLaser can you take a look at this when you get a chance? It's your cluster3d code but ported to be compatible with both icarus and sbnd.

SFBayLaser (approved):
This would be the first step to convert icaruscode and sbndcode to use common versions of the space point building code. These are SBN specific versions of the orginal versions in larreco. 
I agree that these versions should be common to SBN and moved to here.

PR #390, opened 2023-10-16
Feature/howard spillquality

brucehoward-physics:
Putting together PR for a couple of spill info updates to try to get things into the CAFs and in a way that we can also start to do beam quality cuts. Some questions notes:

See some info in doc-33224 slides 5-13

1. IMPORTANT (!!) -- I'm not sure if I should be making this PR against develop or v09_72_00_05 e.g. There seem to be a lot of commits present in the latter but not in develop as my changes are only the last few commits...

2. Think some additional similar changes may be needed in BNB module(s) but probably would be a PR from their side — including Joseph and Jacob L in the conversation here.

3. What else do we need for validation? (See some info at doc-33224)

4. Do we want these parts `(info.spill_time_s+(info.spill_time_ns/1.0e9))` to instead do something with more precision?

5. We probably want to talk to NOvA to remember how they handle the POT sums within CAFAna and with spill quality cuts, but this PR is more about getting the info available for use rather than the internal workings. But it's possible that the way of saving them would preclude certain usage patterns.

Tagging some reviewers that I think make sense given this touches trigger info, spill accounting, CAFs, etc.

### DIFF
--- a/sbncode/BeamSpillInfoRetriever/NuMIRetriever/NuMIRetriever_module.cc
+++ b/sbncode/BeamSpillInfoRetriever/NuMIRetriever/NuMIRetriever_module.cc
@@ -109,6 +109,13 @@ void sbn::NuMIRetriever::produce(art::Event &e)
   e.getByLabel(raw_data_label_, "ICARUSTriggerV3", raw_data_ptr);
   auto const & raw_data = (*raw_data_ptr);
 
+  // NOTE: Really we should skip the first event of each trigger type, so let's make this look at that too...
+  if ( raw_data.size()==0 ) return;
+  else {
+    icarus::ICARUSTriggerV3Fragment frag(raw_data.at(0));
+    if ( frag.getTotalTriggerNuMIMaj() <= 1 ) return;
+  }
+
   double t_current_event  = 0;
   double t_previous_event = 0;
   double number_of_gates_since_previous_event = 0;
@@ -163,8 +170,8 @@ void sbn::NuMIRetriever::produce(art::Event &e)
     // plus or minus some time padding, currently using 3.3 ms
     // which is half the Booster Rep Rate
     if(e.event() != 1){//We already addressed the "first event" above 
-      if(times_temps[i] > t_current_event){continue;}
-      if(times_temps[i] <= t_previous_event){continue;}
+      if(times_temps[i] > t_current_event+fTimePad){continue;}
+      if(times_temps[i] <= t_previous_event+fTimePad){continue;}
     }
     
     //count found spills
--- a/sbncode/BeamSpillInfoRetriever/NuMIRetriever/NuMIRetriever_module.cc
+++ b/sbncode/BeamSpillInfoRetriever/NuMIRetriever/NuMIRetriever_module.cc
@@ -110,7 +110,7 @@ void sbn::NuMIRetriever::produce(art::Event &e)
   auto const & raw_data = (*raw_data_ptr);
 
   // NOTE: Really we should skip the first event of each trigger type, so let's make this look at that too...
-  if ( raw_data.size()==0 ) return;
+  if ( raw_data.empty() ) return;
   else {
     icarus::ICARUSTriggerV3Fragment frag(raw_data.at(0));
     if ( frag.getTotalTriggerNuMIMaj() <= 1 ) return;
--- a/sbncode/CAFMaker/CAFMaker_module.cc
+++ b/sbncode/CAFMaker/CAFMaker_module.cc
@@ -148,6 +148,13 @@ namespace sbn{
 
 namespace caf {
 
+/// Function to calculate a timestamp from the spill info product
+template <typename SpillInfo>
+double spillInfoToTimestamp(SpillInfo const& info) {
+  return static_cast<double>(info.spill_time_s) +
+         static_cast<double>(info.spill_time_ns)*1.0e-9;
+}
+
 /// Module to create Common Analysis Files from ART files
 class CAFMaker : public art::EDProducer {
  public:
@@ -785,12 +792,9 @@ void CAFMaker::beginSubRun(art::SubRun& sr) {
     fHasBNBInfo = true;
     for(const sbn::BNBSpillInfo& info: *bnb_spill)
     {
-      if ( fBNBInfoEventMap.find(info.event)==fBNBInfoEventMap.end() ) {
-        fBNBInfoEventMap[info.event] = info;
-      }
-      else if ( (info.spill_time_s+(info.spill_time_ns/1.0e9)) >
-                (fBNBInfoEventMap[info.event].spill_time_s+(fBNBInfoEventMap[info.event].spill_time_ns/1.0e9)) ) {
-        fBNBInfoEventMap[info.event] = info;
+      auto& storedInfo = fBNBInfoEventMap[info.event]; // creates if needed
+      if ( (storedInfo.event == UINT_MAX) || spillInfoToTimestamp(info) > spillInfoToTimestamp(storedInfo) ) {
+        storedInfo = std::move(info);
       }
     }
   }
@@ -803,12 +807,9 @@ void CAFMaker::beginSubRun(art::SubRun& sr) {
     fHasNuMIInfo = true;
     for(const sbn::NuMISpillInfo& info: *numi_spill)
     {
-      if ( fNuMIInfoEventMap.find(info.event)==fNuMIInfoEventMap.end() ) {
-        fNuMIInfoEventMap[info.event] = info;
-      }
-      else if ( (info.spill_time_s+(info.spill_time_ns/1.0e9)) >
-                (fNuMIInfoEventMap[info.event].spill_time_s+(fNuMIInfoEventMap[info.event].spill_time_ns/1.0e9)) ) {
-        fNuMIInfoEventMap[info.event] = info;
+      auto& storedInfo = fNuMIInfoEventMap[info.event]; // creates if needed
+      if ( (storedInfo.event == UINT_MAX) || spillInfoToTimestamp(info) > spillInfoToTimestamp(storedInfo) ) {
+        storedInfo = std::move(info);
       }
     }
   }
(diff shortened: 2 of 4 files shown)


brucehoward-physics:
One other thing to check is that we probably want some way to directly check in the CAFs `nspill` from ifbeam vs `nspill` from TDAQ reporting to cut instances of disagreement if indeed that's what we want to do --- need to see if that's available or needs to be added.

brucehoward-physics:
Per @PetrilloAtWork request to make things less confusing (even if we still have to figure out about the differences between this and develop) I've changed the base branch to be `release_v09_72_00_03` which looking at my area's `git log` seemed to be what the tag I was working from pointed to.

sbncode/CAFMaker/CAFMaker_module.cc
```diff
@@ -769,10 +779,38 @@ void CAFMaker::beginSubRun(art::SubRun& sr) {
...
+
+    // Find the spill for each event and fill the event map:
+    // We take the latest spill for a given event number to be the one to keep
+    fHasBNBInfo = true;
+    for(const sbn::BNBSpillInfo& info: *bnb_spill)
+    {
+      if ( fBNBInfoEventMap.find(info.event)==fBNBInfoEventMap.end() ) {
+        fBNBInfoEventMap[info.event] = info;
+      }
+      else if ( (info.spill_time_s+(info.spill_time_ns/1.0e9)) >
```
PetrilloAtWork, line 791: Can you put this into a function, like:
```cpp
namespace {
  template <typename SpillInfo>
  double spillInfoToTimestamp(SpillInfo const& info) {
    return static_cast<double>(info.spill_time_s)
      + static_cast<double>(info.spill_time_ns)*1e-9;
  }
}
```
Even better would be to add a `timestamp()` member function to `caf::SRBNBSpillInfo` and sister.

sbncode/CAFMaker/CAFMaker_module.cc
```diff
@@ -769,10 +779,38 @@ void CAFMaker::beginSubRun(art::SubRun& sr) {
...
+                (fBNBInfoEventMap[info.event].spill_time_s+(fBNBInfoEventMap[info.event].spill_time_ns/1.0e9)) ) {
+        fBNBInfoEventMap[info.event] = info;
+      }
```
PetrilloAtWork, line 794: Minor point: optimisation to reduce lookups in the map:
```suggestion
      auto const iInfo = fBNBInfoEventMap.find(info.event);
      if ( iInfo==fBNBInfoEventMap.end() ) {
        fBNBInfoEventMap[info.event] = info;
      }
      else if ( (info.spill_time_s+(info.spill_time_ns/1.0e9)) >
                (iInfo->spill_time_s+(iInfo->spill_time_ns/1.0e9)) ) {
        *iInfo = info;
      }
```
Another approach is a bit more convoluted:
```suggestion
      auto& storedInfo = fBNBInfoEventMap[info.event];
      if (storedInfo.event == UINT_MAX) { // no info
        storedInfo = info;
      }
      else if ( (info.spill_time_s+info.spill_time_ns/1.0e9) >
                (storedInfo.spill_time_s+storedInfo.spill_time_ns/1.0e9)) ) {
        storedInfo = info;
      }
```
which relies on the fact that a default-constructed `SpillInfo` object has an event number `UINT_MAX`; or, merging with the other suggestion:
```suggestion
      auto& storedInfo = fBNBInfoEventMap[info.event]; // creates if needed
      if ((storedInfo.event == UINT_MAX) // no info
        || spillInfoToTimestamp(info) > spillInfoToTimestamp(storedInfo) // newer
      {
        storedInfo = std::move(info);
      }
```

sbncode/CAFMaker/CAFMaker_module.cc
```diff
@@ -1984,6 +2022,20 @@ void CAFMaker::produce(art::Event& evt) noexcept {
   // rec.hdr.blind = 0;
   // rec.hdr.filt = rb::IsFiltered(evt, slices, sliceID);
 
+  // Fill the header info for the given event's spill quality info
+  if ( fHasBNBInfo && fHasNuMIInfo ) {
```
PetrilloAtWork, line 2026: I am all for throw an exception, that in `CAFMaker` is actually `abort()`.
I don't think this should happen, and if it does we can probably fix it if we notice.

sbncode/CAFMaker/CAFMaker_module.cc
```diff
@@ -1984,6 +2022,20 @@ void CAFMaker::produce(art::Event& evt) noexcept {
...
+    std::cout << "Found > 0 BNBInfo size and NuMIInfo size, which seems strange. Will not fill event-specific spill quality info for this event..." << std::endl;
+  }
+  else if ( fHasBNBInfo ) {
+    if ( fBNBInfoEventMap.find(evt.id().event()) == fBNBInfoEventMap.end() )
```
PetrilloAtWork, line 2030: Or, more intelligible:
```suggestion
    unsigned int const eventNo = evt.id().event();
    if ( fBNBInfoEventMap.count(eventNo) == 0 )
```

sbncode/CAFMaker/CAFMaker_module.cc
```diff
@@ -1984,6 +2022,20 @@ void CAFMaker::produce(art::Event& evt) noexcept {
...
+      std::cout << "We think (since fHasBNBInfo) that this event should be BNB, but did not find this event in the spill info map." << std::endl;
+    else rec.hdr.spillbnbinfo = makeSRBNBInfo(fBNBInfoEventMap.at(evt.id().event()));
+  }
+  else if ( fHasNuMIInfo ) {
+    if ( fNuMIInfoEventMap.find(evt.id().event()) == fNuMIInfoEventMap.end() )
```
PetrilloAtWork, line 2035: Ditto.

PetrilloAtWork (approved):
Looks ok, comments left mostly for C++ improval plus an error that I would prefer to be fatal.
Becomes approval if you think it's ok as is.

sbncode/CAFMaker/CAFMaker_module.cc
brucehoward-physics: Fair, I can make this a loud `abort()` along with the printout.

sbncode/CAFMaker/CAFMaker_module.cc
brucehoward-physics: Wait shouldn't count > 0 if `fHasBNBInfo` ?

sbncode/CAFMaker/CAFMaker_module.cc
brucehoward-physics: Good idea! Is the `namespace { }` enclosure here just a generic template of such a function or do you prefer it live in a specific namespace? Otherwise, I was thinking to just put it in `caf` namespace in the code.

brucehoward-physics:
@PetrilloAtWork does what I just pushed address your idea with the `count()` correctly or not quite?

Tried to address other points as well, let me know if something misses the mark. Thanks for the suggestions! (Will be running a test to see that I get the same POT values or something else as with my previous test.)

sbncode/CAFMaker/CAFMaker_module.cc
PetrilloAtWork: The [unnamed namespace](https://en.cppreference.com/w/cpp/language/namespace#Unnamed_namespaces) has a specific meaning, in that its content is not visible outside the ~translation unit~ C++ source file where it is defined. For example, if I wrote a header file `cheater.h` like:
```cpp
template <typename SpillInfo>
double spillInfoToTimestamp(SpillInfo const& info);
```
and I included it in another source file:
```cpp
#include "cheater.h"

double const timeStamp = spillInfoToTimestamp(SpillInfo{});
```
this would fail linking because `spillInfoToTimestamp` would not be found anywhere.

[example faults: (1) for a template this should not be done in such way anyway (pretend it's not a template) and (2) the header where `SpillInfo` is declared needs to be included somewhere]

sbncode/BeamSpillInfoRetriever/NuMIRetriever/NuMIRetriever_module.cc
```diff
@@ -109,6 +109,13 @@ void sbn::NuMIRetriever::produce(art::Event &e)
   e.getByLabel(raw_data_label_, "ICARUSTriggerV3", raw_data_ptr);
   auto const & raw_data = (*raw_data_ptr);
 
+  // NOTE: Really we should skip the first event of each trigger type, so let's make this look at that too...
+  if ( raw_data.size()==0 ) return;
```
PetrilloAtWork, line 113: I can't believe I missed one of my favourite pet peeves!
```suggestion
  if ( raw_data.empty() ) return;
```

sbncode/CAFMaker/CAFMaker_module.cc
PetrilloAtWork: That count "counts" the number of entries in the map _for `eventNo`_: there may be some information (in which case presumably `fHasBNBInfo` is `true`), but the one for `eventNo` might be missing, in which case that `count(eventNo)` returns `0`.
I would call that situation a problem, but technically it is possible.

PetrilloAtWork (approved):
I confirm the approval.

PetrilloAtWork:
> @brucehoward-physics @PetrilloAtWork Is this still needed for the " current ICARUS production" or as been discussed can go into develop and into the soon coming new production release?

I hereby summon @SFBayLaser since I definitely don't have knowledge current enough to answer.

brucehoward-physics:
I don't know what happened with my attempted retest but I am getting the same output POT I got the first time, so I'm back to thinking this is okay... I even added the fix for @PetrilloAtWork 's preference to use `empty()`

jzennamo (approved):

brucehoward-physics:
@kjplows I see this got assigned to me. I had to go back and check, but I _think_ this PR can just be closed (🤔)

See e.g. https://github.com/SBNSoftware/sbncode/pull/407

but maybe someone else here remembers if there was a reason to keep this version around?

kjplows:
Hi @brucehoward-physics , since I haven't heard anyone object and #407 got merged, I am closing this PR now. Many thanks!

PR #397, opened 2023-11-16
Decouple storing MCParticle from storing MCParticleLite and MCParticl…

bear-is-asleep:
…eAssignment

This will ruin the default behavior but currently this module is only used by mlreco group and we will not be using `MCParticleLite` in the refactored G4 code. Therefore, the default behavior once the code migrates to refactored G4 anywats.

### DIFF
--- a/sbncode/LArG4/MergeSimSourcesSBN_module.cc
+++ b/sbncode/LArG4/MergeSimSourcesSBN_module.cc
@@ -73,6 +73,18 @@ public:
       true // default
     };
 
+    fhicl::Atom<bool> FillMCParticlesLite{
+      fhicl::Name{"FillMCParticlesLite"},
+      fhicl::Comment{"whether to merge MCParticleLite"},
+      true // default
+    };
+
+    fhicl::Atom<bool> FillMCParticlesAssociated{
+      fhicl::Name{"FillMCParticlesAssociated"},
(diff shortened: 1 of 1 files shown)


sbncode/LArG4/MergeSimSourcesSBN_module.cc
```diff
@@ -300,11 +322,12 @@ void sbn::MergeSimSourcesSBN::produce(art::Event& e)
 
   for (auto const& [i_source, input_label] : util::enumerate(fInputSourcesLabels)) {
 
+    auto const input_partCol = e.getValidHandle<std::vector<simb::MCParticle>>(input_label);
+    art::PtrMaker<simb::MCParticle> const makePartPtr{e};
```
PetrilloAtWork, line 326: It seems that this object is used only in the `if (fFillMCParticlesAssociated)` branch of the code, so I suggest it is moved into that branch.

PetrilloAtWork (approved):
I leave a very minor suggestion, but in general I can't find any issue with the code.

sbncode/LArG4/MergeSimSourcesSBN_module.cc
miquelnebot: @bear-is-asleep would you address this?

bear-is-asleep: done! thanks for the comment i just didn't see it until now..

PR #400, opened 2023-11-28
Fix bug in filling exposure information in CAFs

jacoblarkin:
Fill pot and beam information in the CAFs per subrun instead of per file and make sure it is filled in both blind and prescaled files.


etworcester (approved):
Looks good, thanks for fixing this

PR #404, opened 2023-12-19
Feature/bearc simplemerge

bear-is-asleep:
Merge two products without any fancy trackID offsets. Currently only supports MCParticles. Changes from https://github.com/SBNSoftware/sbncode/pull/397 have leaked into this PR which should be merged into develop soon


bear-is-asleep:
The changes from [PR](https://github.com/SBNSoftware/sbncode/pull/404#:~:text=MCParticles.%20Changes%20from-,%23397,-have%20leaked%20into) have been leaked which I can fix if need be.

sbncode/LArG4/CMakeLists.txt
```diff
@@ -31,6 +31,20 @@ cet_build_plugin(MergeSimSourcesSBN art::EDProducer
...
+  lardataobj::MCBase
+  larcorealg::headers
+  art::Framework_Principal
+  art::Framework_Services_Registry
+  art::Persistency_Common
+  messagefacility::MF_MessageLogger
+  canvas::canvas
+  fhiclcpp::types
+  fhiclcpp::fhiclcpp
+  BASENAME_ONLY
```
PetrilloAtWork, line 45: `BASENAME_ONLY` should not be used. It makes the library name short (`libSimpleMerge_module.so`), which is nice if you need to type it but a nightmare when some other library with the same name happens to exist anywhere.
I **recommend** to remove it.

sbncode/LArG4/CMakeLists.txt
```diff
@@ -31,6 +31,20 @@ cet_build_plugin(MergeSimSourcesSBN art::EDProducer
   BASENAME_ONLY
 )
 
+cet_build_plugin(SimpleMerge art::EDProducer
```
PetrilloAtWork, line 34: This is the first time I see `art::EDProducer` here... I have learned something.

sbncode/LArG4/CMakeLists.txt
```diff
@@ -31,6 +31,20 @@ cet_build_plugin(MergeSimSourcesSBN art::EDProducer
...
+  LIBRARIES
+  lardataobj::MCBase
+  larcorealg::headers
+  art::Framework_Principal
+  art::Framework_Services_Registry
+  art::Persistency_Common
+  messagefacility::MF_MessageLogger
+  canvas::canvas
+  fhiclcpp::types
+  fhiclcpp::fhiclcpp
```
PetrilloAtWork, line 44: Almost all of these are already included when ~`art::module`~ `art::EDProducer` is used above:
```suggestion
  messagefacility::MF_MessageLogger
```

sbncode/LArG4/SimpleMerge_module.cc
```diff
@@ -0,0 +1,103 @@
...
+  bool const fResetMotherID;
+
+};
+
+
+sbn::SimpleMerge::SimpleMerge(fhicl::ParameterSet const& p)
+  : EDProducer{p}
+  , fInputSourcesLabels(p.get<std::vector<art::InputTag>>("InputSourcesLabels"))
+  , fFillMCParticles(p.get<bool>("FillMCParticles", false))
+  , fTrackIDOffsets(p.get<std::vector<int>>("TrackIDOffsets", std::vector<int>(0)))
```
PetrilloAtWork, line 65: I suppose an empty vector is intended here. Try one of the following:
```suggestion
  , fTrackIDOffsets(p.get<std::vector<int>>("TrackIDOffsets", {}))
```
or
```suggestion
  , fTrackIDOffsets(p.get("TrackIDOffsets", std::vector<int>{}))
```
or
```suggestion
  , fTrackIDOffsets(p.get<std::vector<int>>("TrackIDOffsets", std::vector<int>{}))
```
I _recommend_ avoiding `std::vector<int>(0)` because it looks ambiguous, whether you are initialising a vector with `0` elements or if you are initialising a vector with one element with value `0`.

PetrilloAtWork (changes requested):
I apologise for taking this long to notice the request for review.

`MergeSimSourcesSBN` module: it seems that the options `FillMCParticles` and `FillMCParticlesAssociated` are not independent. I think the most reasonable approach is to require saving the particles when saving their associations. There are several ways to obtain that with FHiCL (you can look to the conditional enabling feature of `fhicl::Atom` if you are brave). The most straightforward is to add at the beginning of the constructor a configuration check:
```cpp
  if (fFillMCParticlesAssociated && !fFillMCParticles) {
    throw art::Exception{ art::errors::Configuration }
      << "To save particle associations (`FillMCParticlesAssociated`) particles must be saved too (`FillMCParticles`).";
  }
```
Or if you want to be brutal, force it right:
```cpp
  , fFillMCParticlesAssociated(params().FillMCParticlesAssociated() || fFillMCParticles)
```
in the constructor. I **request** this dependency to be enforced or resolved, in the way you deem more proper.

`SimpleMerge`: you can consider writing a `art::SharedProducer` instead of a `art::EDProducer`. That will allow _art_ to do some multithreading. [The changes](https://cdcvs.fnal.gov/redmine/projects/art/wiki/Module_threading_types#Comparison-of-module-interfaces) are in general trivial: a different header to be included, an additional argument (which you can then ignore) in some of the methods (e.g. in `produce()`), and adding the type of multithreading (`art::async<art::Event>();`) in the constructor. This is just a *suggestion*.
I don't get the point of the parameter `FillMCParticles` in this module: rather than setting it to `false`, one can as well not to run the module at all. I **recommend** the parameter to be removed.
It seems to me that the `TrackIDOffsets` parameter is not describing track ID offsets, but rather a list of IDs for particles to be forcibly orphaned. I **request** that the name of the parameter to be changed to reflect its meaning (`ResetMotherOf` maybe? I am not great with names).
Also, for thrift, I *suggest* you to remove the `ResetMotherID` parameter as it is redundant: if the list of "track offsets" is empty, no mother ID will be reset, so you can assume that if every ID in the "track offset" parameter is a request to reset its mother.

sbncode/LArG4/SimpleMerge_module.cc
bear-is-asleep: The intention is to have a vector with just 0 in there. Now that I look [here](https://github.com/SBNSoftware/sbncode/blob/3699a40018678bc7980ea111599b1a7655818660/sbncode/LArG4/SimpleMerge_module.cc#L87-L92) though, I see that if the mother ID is already 0, there's no point in setting it if the offset is set to 0. I suppose leaving it as a blank vector will be a good default!

sbncode/LArG4/CMakeLists.txt
bear-is-asleep: done!

sbncode/LArG4/CMakeLists.txt
bear-is-asleep: yeah I just copied another producer that worked upon building !

bear-is-asleep:
Moving this PR to [here](https://github.com/SBNSoftware/sbncode/pull/430) so we can combat changes to both SBN merge and simple merge, since I got comments on both.

PR #407, opened 2024-01-11
Feature/howard spillquality for develop based on j zennamo branch

brucehoward-physics:
This is a copy of the commits in PR 390: https://github.com/SBNSoftware/sbncode/pull/390

Note from the comments/discussion in that thread that the PR there is against production patch and not develop, which is in part because there were commits not in develop that it was picking up on when telling the differences between the branch to be merged and the base branch. Looking into it, I think some of those are actually important to the module that was updated by my PR. That makes switching the linked PR back to develop (the initial suggestion) seem less favorable. While develop doesn't yet have some of the changes I noticed (like TriggerV2 -> TriggerV3), I talked to Joseph who pointed me to the PR for this branch.

Plus, as discussed with some others at the ICARUS workshop in December, we will want to have a production patch at some point (for NuMI analyses) that would incorporate this, so having a PR against that can still be useful when we go to do that.

So, here is _hopefully_ a solution -- Ieft the old PR linked above active as it was, and I checked out Joseph's feature branch, cherry picked my 5 commits from the other PR, pushed and made this PR... Marking the same reviewer as linked commit, as well as Miquel as rel mgr and Joseph as the person whose PR/branch I'm piggy-backing on.

Is there something we can/should do to test this? And/or hopefully you will have suggestions how to proceed. Thanks!

### DIFF
--- a/sbncode/BeamSpillInfoRetriever/NuMIRetriever/NuMIRetriever_module.cc
+++ b/sbncode/BeamSpillInfoRetriever/NuMIRetriever/NuMIRetriever_module.cc
@@ -109,6 +109,13 @@ void sbn::NuMIRetriever::produce(art::Event &e)
   e.getByLabel(raw_data_label_, "ICARUSTriggerV3", raw_data_ptr);
   auto const & raw_data = (*raw_data_ptr);
 
+  // NOTE: Really we should skip the first event of each trigger type, so let's make this look at that too...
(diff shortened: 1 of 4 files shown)


jzennamo (approved):
Looked good to me and contained some things we should adopt in the BNB accounting!

miquelnebot (approved):

PR #411, opened 2024-01-19
Razzled

henrylay97:
Razzle**d** is a new MVA PID based on using both the Razzle & Dazzle inputs to make a determination of the likely PID of a pandora PFParticle. It is all _heavily_ based on Ed's work to implement the original Razzle and Dazzle. Discussed in docDBs: [31183](https://sbn-docdb.fnal.gov/cgi-bin/sso/ShowDocument?docid=31183), [31421](https://sbn-docdb.fnal.gov/cgi-bin/sso/ShowDocument?docid=31421) and [33367](https://sbn-docdb.fnal.gov/cgi-bin/sso/ShowDocument?docid=33367) among others.

This PR:
- adds the module that produces the training trees and / or (depending on fcl configurations) evaluates the scores and adds an object (using a pre-made object by Ed) to the event.
- adds the scores to the CAFs in the same manner as Razzle & Dazzle
- changes the calculation of the best plane in the Razzle module to be explicit as bugs have been found in the stored version of this quantity. The same calculation is used in the Razzle**d** module.
- adds the generic fcl parameters for running the module (explicit use is in sbndcode)
- tinkers with training scripts for the three PID MVAs.

The only reason it is a draft is the weights are not yet in `sbnd_data` to be accessed. This is because there is a CRUMBS + OpT0Finder PR coming next week which also requires weights updates so I'd like to do that in one go but I thought I'd get these up now so reviews could start.

Auxilliary PRs can be found in `sbnobj`, `sbndcode` and `sbnanaobj`.


sbncode/CAFMaker/FillReco.cxx
```diff
@@ -930,6 +930,22 @@ namespace caf
...
+  void FillPFPRazzled(const art::Ptr<sbn::MVAPID> razzled,
+                      caf::SRPFP& srpfp,
+                      bool allowEmpty)
+  {
+    srpfp.razzled.electronScore = razzled->mvaScoreMap.at(11);
+    srpfp.razzled.muonScore = razzled->mvaScoreMap.at(13);
+    srpfp.razzled.photonScore = razzled->mvaScoreMap.at(22);
+    srpfp.razzled.pionScore = razzled->mvaScoreMap.at(211);
+    srpfp.razzled.protonScore = razzled->mvaScoreMap.at(2212);
+
```
lynnt20, line 942: any particular reason to not include the `otherScore`? I suppose that if the `otherScore` is the best, then the pdg will be -5 and the score will be available in `bestScore` so one doesn't really need that information 😛

VCLanNguyen: I personally do not use `otherScore` in my selection but I don't see why not. 
Was wondering if it'll be confusing for analyzer.
Maybe can add a note somewhere in the code explaining the score

henrylay97: Thanks for bringing this up @lynnt20 , I hadn't noticed I had left lingering references to 'other' in the module.

The version of Razzled the three of us have all been using doesn't actually have an other score, I will push a commit that removes all references to it. 

The approach was modelled on Razzle & Dazzle which _did_ have `otherScore` categories because that could be filled with 'mis-labelled' particles (muons, protons & pions in Razzle and electrons & photons in Dazzle). So the first iteration of Razzled also contained it but I quickly realised that it wasn't really feasible because it would only contain "exotic" particles like kaons etc and there just weren't enough to train on.

lynnt20 (approved):

fjnicolas:
trigger build SBNSoftware/sbnobj#103 SBNSoftware/sbnanaobj#117 SBNSoftware/sbndcode#403

PR #413, opened 2024-01-23
OpT0 CRUMBS

henrylay97:
This PR adds OpT0 variables as an alternative flash matching input to the CRUMBS tool (already makes use of SimpleFlash). This was discussed in various docDBs, the most relevant being [31598](https://sbn-docdb.fnal.gov/cgi-bin/sso/ShowDocument?docid=31598) and [33367](https://sbn-docdb.fnal.gov/cgi-bin/sso/ShowDocument?docid=33367).

The default is being set as using OpT0 given the marginally better performance but support is retained for the training of either type. I have also swept some name changes of CRT variables into this PR such that we only need to change the weights once for this round of updates.

Finally, I am making it a draft PR for two reasons:
- ~This PR requires the (approved) CRT Clustering PRs to be merged first, until then the diff will look huge, it's not.~
- The new version of sbnd_data with new weights will need to be in place to test this PR, I will work offline with the release managers for this.

Auxilliary PRs can be found in `sbnanaobj`, `sbndcode` and `sbnobj`.

### DIFF
--- a/sbncode/CAFMaker/FillReco.cxx
+++ b/sbncode/CAFMaker/FillReco.cxx
@@ -433,6 +433,8 @@ namespace caf
       slice.crumbs_result.pds.fmtotalscore = crumbs->pds_FMTotalScore;
       slice.crumbs_result.pds.fmpe = crumbs->pds_FMPE;
       slice.crumbs_result.pds.fmtime = crumbs->pds_FMTime;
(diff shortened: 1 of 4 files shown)


lynnt20 (approved):

fjnicolas:
trigger build SBNSoftware/sbnobj#104 SBNSoftware/sbnanaobj#118 SBNSoftware/sbndcode#405

PR #416, opened 2024-01-29
update to follow art conventions

jzennamo:
The `put` command needed to be updated to follow the latest art convention. This shouldn't change functionality but resolves the `warning` that appears during build.


jzennamo:
trigger build

PetrilloAtWork (approved):
Yup.
I think.

miquelnebot (approved):

brucehoward-physics:
Glad you reviewed this @PetrilloAtWork -- are there are other places where the _art_ convention changing has not yet been caught that we should be looking for?

PR #417, opened 2024-02-05
Add Geant4Reweight to sbncode

gputnam:


sjgardiner:
This PR looks reasonable, but the build currently fails due to some compilation problems. I'm pretty sure it's because we are trying to link against geant4reweight v01_20_03, which had API changes relative to the older version (v01_00_03) used in the original code from MicroBooNE. The binary tarball I used for testing was v01_00_03e. My suggested short-term solution is to downgrade to v01_00_03 and test again. We could request an official e26:prof build of that version from SciSoft for the production release. Eventually we'll need to update to use the new API.

kjplows:
@gputnam with the merging of #555 I'm going to close this if that's okay with you?

PR #420, opened 2024-02-06
Save offbeam gate count to CAFs

jacoblarkin:
Save the number of offbeam gates 

depends on https://github.com/SBNSoftware/sbnanaobj/pull/121


sbncode/CAFMaker/CAFMaker_module.cc
```diff
@@ -773,8 +776,27 @@ void CAFMaker::beginSubRun(art::SubRun& sr) {
   fHasNuMIInfo = false;
 
   fSubRunPOT = 0;
+  fOffbeamBNBGates = 0;
+  fOffbeamNuMIGates = 0;
```
jzennamo, line 780:
```suggestion
  fOffbeamBNBGates = 0.;
  fOffbeamNuMIGates = 0.;
```

jzennamo (approved):
With my one suggestion this looks good to me. Might want to have @brucehoward-physics take a look too, just in case this collides with any changes he made.

brucehoward-physics:
I am guessing you mean the "keep the event's corresponding spill in the event record" change right @jzennamo ? From a quick look this shouldn't conflict with that (this seems to mostly be adding things and not diverting things away from where I did stuff). Thanks for letting me know to take a look though!

PR #421, opened 2024-02-07
Introduce module to respect the trigger simulation

woodtp:
The purpose of `AdjustSimForTrigger_module.cc` is to shift the time of simulated products based on the `raw::Trigger::TriggerTime` created by `emuTrigger`.

A time shift is computed by calculating the time difference between the `raw::Trigger:TriggerTime` and `detinfo::DetectorClockData::TriggerTime`, and subtracting this value from the following products:

1. `sim::SimEnergyDeposit::StartT` and `sim::EnergyDesposit::EndT`
2. `sim::SimPhotons::OnePhoton::Time`
3. `sim::BeamGateInfo::Start`
4. `raw::OpDetWaveform::TimeStamp`

An `art::errors::EventProcessorFailure` exception is thrown if any of the following is true:

1. All the product shifts are disabled
2. There is no `raw::Trigger`
3. There is no `sim::BeamGateInfo`
4. There is more than 1 `raw::Trigger`
5. There is more than 1 `sim::BeamGateInfo`

If there is an available trigger but the `TriggerTime` is invalid (outside the numerical limits of `double`), the time shift defaults to 0 us.

One can also adjust the time shift by modifying `AdditionalOffset`; see [icaruscode PR #673](https://github.com/SBNSoftware/icaruscode/pull/673) for the corresponding detsim fcl.

Related presentation from the TPC Reconstruction Meeting on February 12, 2024 [SBN-doc-34946-v1](https://sbn-docdb.fnal.gov/cgi-bin/sso/ShowDocument?docid=34946). 

Validation plots demonstrating the module performs to our expectations [SBN-doc-35018-v1](https://sbn-docdb.fnal.gov/cgi-bin/sso/ShowDocument?docid=35018).


aheggest (approved):
All looks good to me! I checked the CRT time distribution and the CRT-PMT time difference distribution in files Tony produced and the distributions look the same with and without the changes.

PR #426, opened 2024-02-15
Saving GENIE tree to caf [release/SBN2023A_NuMI]

jedori0228:


PetrilloAtWork (changes requested):
I did not look at the code too much in detail, but it's not clear to me how you avoid duplicate GENIE matching indices. Each job will start with index `0` and go on, right?

jedori0228:
> I did not look at the code too much in detail, but it's not clear to me how you avoid duplicate GENIE matching indices. Each job will start with index `0` and go on, right?

Yes, `fGenieEventCounter` is increased everytime it fills one entry in the genie tree.

PetrilloAtWork:
In general the concern is that the newly introduced index is not enforced to be unique, which demands a specific workflow in order to make the matching between entries in the new tree ("GENIE tree") and in the rest of the data ("CAF tree").
Because of this non-uniqueness, blindly using file lists or `hadd`'ing files won't do.

@jedori0228 is providing a tool that can generate a reindexed tree: as long as this is used, joining the trees is possible.

Also, the index can be used for fast lookup of the needed entry. This works only if the order of the entries in the tree is never changed (if for example the input is a file list, the order of the files must not be changed, or the lookup will silently pick a wrong entry).

My recommendation to address this fragility is to introduce a unique identifier for each entry in the CAF tree, and replicate it in all the "auxiliary" trees (in this case, the GENIE tree).
In that way there is a mean to look up an entry even when the order of the input is shuffled, and it is an independent check when doing the matching in a different way.

As a candidate for the unique identifier, I suggest the couple of entries: the original input file name, and the index of the event in that file. To make this more practical, I would use instead of the actual file name its hash code (just from `std::hash`). The two components are already stored in the CAF tree (the hash value should instead be added), and there is code that can be used in other _art_ modules to reproduce the values.

For fast lookup, nothing really beats the absolute index of the entry in the tree. Yet, a feature of ROOT trees is the ability to build [an index](https://root.cern/doc/master/classTTreeIndex.html#a08aac749ab22fd5c8ab792a0061a4b0f) that can then be used for lookup. This index can for example map the `(hash ; position)` ID described above to the location of the entry in the tree. The facility is started by loading the `TTree`/`TChain` and running [`BuildIndex()`](https://root.cern/doc/master/classTTree.html#a3f6b5bb591ff7a5bd0b06eea6c12b998) on it; any entry can then be quickly accessed with [`TTree::GetEntryWithIndex(hash, position)`](https://root.cern/doc/master/classTTree.html#a8c77a02a1323019375d273299a2c14b1).

jedori0228:
Thanks @PetrilloAtWork ! I pushed few more commits with updated structure;
- In `sbnanaobj`, I add `std::uint32_t  sourceNameHash` into `caf::SRHeader`. A 32-bit integer is chosen to make sure this value not exceeding 32-bit. This is necessary to be used as one of the ingredients of `TTree::BuildIndex`, where it [shifts the major number by 32-bit](https://root.cern/doc/master/classTTreeIndex.html#a08aac749ab22fd5c8ab792a0061a4b0f).
- Here in `sbncode`, I am now evaluate the hash using `fSourceFile` which is the input filename. Then I truncate this into 32-bit integer using `static_cast<std::uint32_t>`.
- In the GENIE tree, I am also saving this hash
- In the end, both CAF and GENIE tree has `hash of the inputfile` and `GENIE tree position`. I run `BuildIndex` on the GENIE tree so one can access to this tree using `(hash, GENIE tree position)` pair.
- A `hadd`-ed or `concat-caf`-ed file should be duplicated `GENIE tree position`, but those entry should have different hash. User can now correct match between caf::SRTrueInteraction and genie::EventRecord on these merged files.

jedori0228:
Plan is to make a copy-PR of this for develop once this PR is approved and merged.

sbncode/CAFMaker/CMakeLists.txt
```diff
@@ -41,6 +41,8 @@ art_make_library( LIBRARY_NAME sbncode_CAFMaker
                   systematicstools::interface
                   systematicstools::interpreters
                   systematicstools::utility
+                  ${GENIE_LIB_LIST}
```
PetrilloAtWork, line 44: Is this needed? where is this defined?
In the best case, this is actually empty and can be removed.

sbncode/CAFMaker/CAFMaker_module.cc
```diff
@@ -521,6 +537,9 @@ CAFMaker::~CAFMaker()
   delete fFlatTree;
   delete fFlatFile;
 
+  //delete fGenieEvtRec;
+  //delete fGenieTree;
+
```
PetrilloAtWork, line 542: Were these causing double-freeing or some other trouble?
Shouldn't the flat versions also be here?

PetrilloAtWork (commented):
A couple of details to be fixed...
The PR seems, in general, good.

sbncode/CAFMaker/CMakeLists.txt
jedori0228: This was required for
```
#include "Framework/EventGen/EventRecord.h"
#include "Framework/Ntuple/NtpMCEventRecord.h"
```
and further usage of these objects in making GENIE tree. These are defined from [nufinder](https://github.com/NuSoftHEP/nufinder/blob/develop/Modules/FindGENIE.cmake#L44-L70).

sbncode/CAFMaker/CAFMaker_module.cc
jedori0228: I pushed a commit where I now delete `fGenieEvtRec` and `fFlatGenieEvtRec`. But I found `fGenieTree` or `fFlatGenieTree` is deleted when the Record is delete.

PetrilloAtWork: After investigation, the conclusion is that the trees are deleted when _the ROOT file objects_ they are associated to (e.g. `fFile`) are deleted.
In addition, the `fGenieEvtRec` _should_ be deleted by us since, according to [ROOT documentation ("If addr is not zero, but the pointer addr points at is zero" paragraph)](https://root.cern/doc/master/classTBranchElement.html#a7b19271c3bf833e6738bc18a43d6352f), we passed an address variable that has null content at branch address setting time, the record was allocated by ROOT but it still considered owned by us, and it should be explicitly deleted with `delete`.

jedori0228: Thanks! In [1c47597](https://github.com/SBNSoftware/sbncode/pull/426/commits/1c47597b74548ba6589c79a543f99978130c9c30), I properly deleted all the GENIE objects.

PetrilloAtWork (approved):

PR #429, opened 2024-02-15
Saving non-primaries in cafs for g4 study [release/SBN2023A_NuMI]

jedori0228:
For G4 (reweight) study, we save non-primaries to truth.prim.


brucehoward-physics (approved):
We want this!

brucehoward-physics:
Though actually we included this in https://github.com/SBNSoftware/sbncode/pull/431 so really I don't think I needed to approve this... but anyway.

PR #430, opened 2024-02-16
Simple merge and SBN merge changes

bear-is-asleep:
Makes simple merge modules which combines `MCParticle` vectors, with the option to set mother IDs. Also adds a simple check to `MergeSimSourcesSBN` which checks that `MCParticle` is also merged if the associations are merged.

Replaces [this PR](https://github.com/SBNSoftware/sbncode/pull/404) so we can combine simple and sbn merge into a single PR.

### DIFF
--- a/sbncode/LArG4/SimpleMerge_module.cc
+++ b/sbncode/LArG4/SimpleMerge_module.cc
@@ -1,0 +1,93 @@
+////////////////////////////////////////////////////////////////////////
+// Class:       SimpleMerge
+// Plugin Type: producer (Unknown Unknown)
+// File:        producer_module.cc
+//
+// Generated at Tue Dec 19 08:26:07 2023 by Brinden Carlson using cetskelgen
+// from  version .
+//
+// Producer module for merging multiple input sources into a single output
+// without trackID offsets or other complications from merge sim sources.
+
+////////////////////////////////////////////////////////////////////////
+
+#include "art/Framework/Core/EDProducer.h"
+#include "art/Framework/Core/ModuleMacros.h"
+#include "art/Framework/Principal/Event.h"
+#include "art/Framework/Principal/Handle.h"
+#include "art/Framework/Principal/Run.h"
+#include "art/Framework/Principal/SubRun.h"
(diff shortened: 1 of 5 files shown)


sbncode/LArG4/SimpleMerge_module.cc
```diff
@@ -0,0 +1,93 @@
...
+  }
+  fResetMotherID = fResetMotherIDs.size() > 0;
+}
+
+void sbn::SimpleMerge::produce(art::Event& e)
+{
+  auto partCol = std::make_unique<std::vector<simb::MCParticle>>();
+  for (auto const& [i_source,input_label] : util::enumerate(fInputSourcesLabels)) {
+    auto parts = e.getValidHandle<std::vector<simb::MCParticle>>(input_label);
+    for (auto part : *parts) {
```
PetrilloAtWork, line 78: We can skip the handle:
```suggestion
    auto const& parts = e.getProduct<std::vector<simb::MCParticle>>(input_label);
    for (auto part : parts) {
```

sbncode/LArG4/SimpleMerge_module.cc
```diff
@@ -0,0 +1,93 @@
...
+      if (fResetMotherID){
+        for (auto const& offset : fResetMotherIDs) {
```
PetrilloAtWork, line 80: Given that `offset` is a fundamental type (`int`?) we can avoid the reference indirection:
```suggestion
        for (int offset : fResetMotherIDs) {
```

PetrilloAtWork (approved):
I left a couple of C++/_art_ style comments for @bear-is-asleep , but the PR can definitely go in.

bear-is-asleep:
> I left a couple of C++/_art_ style comments for @bear-is-asleep , but the PR can definitely go in.

Thanks for your suggestions! I've updated the PR with your suggestions and it builds and runs successfully.

bear-is-asleep:
trigger build

PetrilloAtWork (approved):
Plain approved, for once.

PR #433, opened 2024-02-21
Moved SystTools from icaruscode so we share this with SBND

jedori0228:
Moving `icarsucode/fcl/caf/SystTools` to `sbncode/SBNEventWeight/jobs/SystTools`. The universes are already thrown in fcl, so it is important to have same configs between SBND and ICARUS. I am creating another PR in `icaruscode` where I removed the directory.


jedori0228:
https://github.com/SBNSoftware/icaruscode/pull/697 is created.

sbncode/SBNEventWeight/jobs/SystTools/configs/config_All_icarus_v2.fcl
```diff
@@ -0,0 +1,15 @@
+#include "config_GENIEReWeightMultisigma_icarus_v2.fcl"
+#include "config_GENIEReWeightMultisim_icarus_v2.fcl"
+#include "config_MINERvA_E2p2h_icarus_v1.fcl"
+#include "config_MINERvAq0q3Tune_icarus_v1.fcl"
+#include "config_MiscInteractionSysts_icarus_v1.fcl"
+#include "config_NOvAStyleNonResPion_icarus_v1.fcl"
```
henrylay97, line 6: Do we want to change the file names to reflect `sbn` status?

sbncode/SBNEventWeight/jobs/SystTools/configs/config_GENIEReWeightMultisigma_icarus_v2.fcl
```diff
@@ -0,0 +1,348 @@
+GENIEReWeight_Multisigma_Tool_Config: {
+
+    tool_type: "GENIEReWeight"
+
+    # Updated 25 July 2023 by S. Gardiner
+    instance_name: "ICARUS_v2_multisigma"
```
henrylay97, line 6: And instance names too?

henrylay97 (commented):
Hi @jedori0228  thank you so much for doing that :D 

This looks great to me. I have one query which is this:

Do we want to rename the files & instance names to be `sbn_v1`  and similar (rather than ICAURS type names). I appreciate the actual config isn't changing and this will affect ICARUS analysis scripts which would need to adopt a new naming convention for reading the weights. I do feel it might avoid future confusion though.

What are your thoughts?

jedori0228:
@henrylay97 Thanks Henry, I just updated the filenames and strings in fhicls.

PetrilloAtWork (commented):
It looks good to me.
I would like somebody else to look at this too, though.
But ping me in case a faster approval is needed, and I will turn this into an approval.

fjnicolas:
trigger build SBNSoftware/icaruscode#697 SBNSoftware/sbndcode#417

PR #435, opened 2024-02-22
Moving to HadronHP for low energy hadron simulation.

linyan-w:
Screenshot from 2024-02-22 13-56-43
Validated via de-excitation light spectrum.


jzettle (approved):
This looks good to me! Thanks for making this change @linyan-w!

PR #437, opened 2024-02-26
Update NuMI flux file location for MeVPrtl. Update a couple HPS configs.

gputnam:
Backport to the 2023A NuMI release.


jamiecd (approved):
Just a note for prosperity: these updated flux files were simulated assuming missing concrete blocks. So, updated event weights accounting for the existing concrete need to be applied by analyzers downstream.

PR #443, opened 2024-03-05
Switch to beam time reference for CRT-PMT matched CRTHit times

aheggest:
Use the beam time reference for the CRT-PMT Matched CRTHit times. Before the CRT Hit time in the srcrthit object was correctly using the beam time reference, but the CRT Hit time in the srcrtpmtmatch was not. These times should be on the same timescale to be easily compared at CAF level


francescopoppi:
I would like to have a discussion toghether with @PetrilloAtWork  on this subject. I have to admit, I am strongly not in favour of changing this in the CAFs, possibly we can add an additional variablt at CAFs.
The reason why I have strong opinions on this matter is that we worked hard to achieve synchronization between CRT and PMT using distributed global trigger signal between the two system. The synchronization was very accurate and we were able thanks to this to precisely identify the triggering flash (-0.6 ish us w.r.t. the Global Trigger). The reason why we did not use beam-related signals (beam-gate start, early warning, trigger_timestamp, etc...) is because they are affect by several effects (8 ns discretization, 25 ish ns jitter) and we wanted to provide a timestamp which was as precise as possible. Effectively, by adding the beam gate timing reference to our CRT-PMT objects will add to our timestamp this discretization and resolution effects, which we mitigated in the first place. It is nonetheless true that it is extremely important to have the flash/crt reference w.r.t. the beam gate start, so I second the proposal to add a new timing variable, but I strongly do not endorse changing the main one.

aheggest:
Thanks for the comment @francescopoppi. I originally added this change to have the two collections of CRT Hit times (the full CRT Hit collection CRT Hit times and the CRT-PMT Matched CRT Hit times) on the same timescale, aka using the same time reference, so they can be compared. Please note that the `caf::SRCRTHit.t1` variable already uses the beam reference, and this shift can always be reversed by subtracting `srtrigger.trigger_within_gate`. Maybe I could add a new variable so both `CRT Hit time wrt trigger` and `CRT Hit time wrt beamgate timestamp` are saved in both CRT Hit collections?

PetrilloAtWork:
I believe there was an agreement that the unshifted time would be preserved into a new leaf.
@aheggest, do I recall right?

aheggest:
> I believe there was an agreement that the unshifted time would be preserved into a new leaf. @aheggest, do I recall right?
Correct! I still have yet to implement the new leaf, I think that will require a paired sbnanaobj PR to add a new variable to the [SRCRTPMTMatch](https://github.com/SBNSoftware/sbnanaobj/blob/develop/sbnanaobj/StandardRecord/SRCRTPMTMatch.h). I can make this modification!

ibsafa:
Hi! @aheggest any updates on this? I believe this is still waiting on the accompanying sbnanaobj PR and the necessary variable changes that Gianluca requested, correct?

aheggest:
hi @ibsafa! yes, this PR is still waiting on a new variable to be added to sbnanaobj. I will close this PR and eventually make a new one with these changes.

PR #447, opened 2024-03-22
Add ability to shift sim::AuxDetIDEs by the trigger time.

woodtp:
In the process of reviewing [`icaruscode` #673](https://github.com/SBNSoftware/icaruscode/pull/673), it was discovered that the CRT hits were not being shifted and therefore CRT-PMT matching was broken by this module as the CRT hit times no longer matched the OpFlash times. To rectify this, I have updated the module allowing it to now shift the `sim::AuxDetIDE` according to the emulated trigger time, which can then be used as input to the CRT DetSim. This is accomplished by recreating the collection of `std::vector<sim::AuxDetSimChannel>` and reconstructing each channel with a new `std::vector<std::AuxDetIDE>`, which has `entryT` and `exitT` shifted by the value of the trigger time.

### DIFF
--- a/sbncode/DetSim/AdjustSimForTrigger_module.cc
+++ b/sbncode/DetSim/AdjustSimForTrigger_module.cc
@@ -74,7 +74,7 @@ AdjustSimForTrigger::AdjustSimForTrigger(fhicl::ParameterSet const& p)
   , fShiftAuxDetIDEs{p.get<bool>("ShiftAuxDetIDEs", false)}
   , fShiftBeamGateInfo{p.get<bool>("ShiftBeamGateInfo", false)}
 {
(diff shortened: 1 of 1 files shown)


aheggest (approved):
These changes and the output files Tony shared look good to me. Thanks!!

woodtp:
@PetrilloAtWork requested a change in the naming convention for the fhicl parameters, to make it clearer exactly which objects are being shifted vs. not. So I have performed the following changes:

1. `InputTriggerLabel` will remain the same, as it is not being shifted, just used to compute the shift.
2. All other `Input*` labels have been prefixed with `Init*`, e.g., `InputSimEnergyDepositLabel` --> `InitSimEnergyDepositLabel`.

PR #449, opened 2024-03-30
Adding additional emuTrigger safety checks to the CAFMaker module

woodtp:
In MC, when an event doesn't fire the trigger, the `raw::Trigger`, and therefore `SRTrigger`, will be filled with the default values, which are set to the numerical limits of `double` (see [lardataobj](https://github.com/LArSoft/lardataobj/blob/develop/lardataobj/RawData/TriggerData.h)). In this case, we should set the `PMT_reference_time` to '0'. Otherwise, PMT flashes will be shifted by some garbage value of O(308).

I also added some additional logging, which prints out the value of the reference time, along with the trigger and beam gate times in the 'detector' reference. Below is an example of an event where we've caught the garbage trigger timestamp, and set the reference shift to 0:

```
%MSG-i CAFMaker:  CAFMaker:cafmaker@BeginModule 30-Mar-2024 14:00:19 CDT  run: 1 subRun: 0 event: 1
Setting PMT reference time to 0 us
    Trigger Time   = -1.79769e+308 us
    Beam Gate Time =  1499.8 us
%MSG
```


woodtp:
I've enabled the CAFMaker module to read both shifted/unshifted `raw::Trigger` in order to calculate the time difference between them, and base the PMT reference time shift off of this value. Otherwise, the shift was 200 ns larger than it should've been.

Example printout for an event generated at 4.5 us:

```
%MSG-i CAFMaker:  CAFMaker:cafmaker@BeginModule 03-Apr-2024 17:16:49 CDT  run: 1 subRun: 0 event: 1
Setting PMT reference time to 4.564 us
    Trigger Time   = 1504.56 us
    Beam Gate Time =  1499.8 us
%MSG
```

sbncode/CAFMaker/CAFMaker_module.cc
```diff
@@ -1392,9 +1392,18 @@ void CAFMaker::produce(art::Event& evt) noexcept {
   art::Handle<std::vector<raw::Trigger>> trig_handle;
   GetByLabelStrict(evt, fParams.TriggerLabel().encode(), trig_handle);
 
+  art::Handle<std::vector<raw::Trigger>> unshifted_trig_handle;
+  GetByLabelStrict(evt, fParams.UnshiftedTriggerLabel().encode(), unshifted_trig_handle);
+
+  const bool isValidTrigger = extratrig_handle.isValid() && trig_handle.isValid() && trig_handle->size() == 1;
+  const bool isValidUnshiftedTrigger = !isRealData && unshifted_trig_handle.isValid() && unshifted_trig_handle->size() == 1;
+
+  const double triggerShift = isValidUnshiftedTrigger ?
```
PetrilloAtWork, line 1401: Since you are using the trigger in this expression, it's reasonable to check for its validity too:
```suggestion
  const double triggerShift = (isValidUnshiftedTrigger && isValidTrigger)?
```

sbncode/CAFMaker/CAFMaker_module.cc
```diff
@@ -1392,9 +1392,18 @@ void CAFMaker::produce(art::Event& evt) noexcept {
   art::Handle<std::vector<raw::Trigger>> trig_handle;
   GetByLabelStrict(evt, fParams.TriggerLabel().encode(), trig_handle);
 
+  art::Handle<std::vector<raw::Trigger>> unshifted_trig_handle;
+  GetByLabelStrict(evt, fParams.UnshiftedTriggerLabel().encode(), unshifted_trig_handle);
+
+  const bool isValidTrigger = extratrig_handle.isValid() && trig_handle.isValid() && trig_handle->size() == 1;
+  const bool isValidUnshiftedTrigger = !isRealData && unshifted_trig_handle.isValid() && unshifted_trig_handle->size() == 1;
```
PetrilloAtWork, line 1399: We can save some time (and possibly some noise) by not attempting anything if this is data:
```suggestion
  art::Handle<std::vector<raw::Trigger>> unshifted_trig_handle;
  if (!isRealData)
    GetByLabelStrict(evt, fParams.UnshiftedTriggerLabel().encode(), unshifted_trig_handle);

  const bool isValidTrigger = extratrig_handle.isValid() && trig_handle.isValid() && trig_handle->size() == 1;
  const bool isValidUnshiftedTrigger = unshifted_trig_handle.isValid() && unshifted_trig_handle->size() == 1;
```

PetrilloAtWork (approved):
I have left some technical comments.
The implementation appears correct, so in principle it can go in as is, but I would like the author to consider the suggested changes.

ibsafa:
trigger build

ibsafa:
trigger build

ibsafa:
trigger build

ibsafa:
@woodtp Can you check this failed build warning for ICARUS?

PR #460, opened 2024-04-29
supera update for ICARUS production

yeonjaej:


yeonjaej:
@bear-is-asleep can review this.

bear-is-asleep (approved):
The particle validity check was turned off for all ICARUS fcls, and was already off for all SBND fcls. Additionally, the SBND fcl was updated to use the correct mpvmpr producer format. APPROVED!

PR #461, opened 2024-05-06
Update NearestWireID

bear-is-asleep:
Fixed bug where out of bounds IDs set TPC and Plane info to bad values.


SFBayLaser (approved):
What a really bizarre problem! But it looks like this solves the problem so all good!

bear-is-asleep:
trigger build

Issue #453, opened 2024-04-04
Geant4 Physics List


marcodeltutto:
With PR #435 we switched to a new high-precision (HP) physics list. We should make the old and new ones both available so we can switch between the two as needed.

Plan is to keep `SBN_QGSP_BERT_NNC` to be the old one (so would need to be reverted) and create a new one with name `SBN_QGSP_BERT_HP_NNC` which contains the HP one.

This will require both SBND and ICARUS to point to the new list here:
- SBND: [https://github.com/SBNSoftware/sbndcode/blob/252d833fd3849b82fcbbe7879870daecffe193a2/sbndcode/LArG4/larg4_services_sbnd.fcl#L11](https://github.com/SBNSoftware/sbndcode/blob/develop/sbndcode/LArG4/larg4_services_sbnd.fcl#L11)
- ICARUS: https://github.com/SBNSoftware/icaruscode/blob/ffa6d4167b9d44f86c1e79300e1c646c4eb730ce/fcl/g4/larg4_services_icarus.fcl#L11

marcodeltutto:
Looking into this, I had the following question: why are we not using Geant4's official `QGSP_BERT_HP` instead of creating our own? I think it would be better to use an official Geant4 physics unless there is motivation for changing it. @linyan-w do you think it's okay if we use `QGSP_BERT_HP`?

I see that QGSP_BERT_HP differs from the current list in the following:
- has the additional G4RadioactiveDecayPhysics
- has the additional G4IonElasticPhysics
- used G4HadronElasticPhysicsVI instead of G4HadronElasticPhysicsHP

linyan-w:
I was following the convention in `SBN_QGSP_BERT_NNC`. 
I have no objection to the official `QGSP_BERT_HP` physics list. @sjgardiner may have more thoughts there?

marcodeltutto:
Then @Li-Jiaoyang97 I think we should revert `SBN_QGSP_BERT_NNC` to what it was originally, and then switch sbndcode and icaruscode to use `QGSP_BERT_HP` in the two places I mentioned above. Thanks!

Li-Jiaoyang97:
[...]

PR #462, opened 2024-05-06
Restore SBN_QGSP_BERT_NNC to be QGSP_BERT with no neutron cut

marcodeltutto:
This PR restores the `SBN_QGSP_BERT_NNC` physics list to be "`QGSP_BERT` with no neutron cut".
Two separate PRs should follow, one for SBND and one for ICARUS, where the `QGSP_BERT_HP` physics list is adopted:

- [x] SBND: https://github.com/SBNSoftware/sbndcode/pull/468
- [ ] ICARUS: https://github.com/SBNSoftware/icaruscode/pull/737

See issue #453 for more details.

### DIFF
--- a/sbncode/LArG4/PhysicsLists/SBN_QGSP_BERT_NNC.cc
+++ b/sbncode/LArG4/PhysicsLists/SBN_QGSP_BERT_NNC.cc
@@ -18,9 +18,9 @@
 #include "Geant4/G4EmExtraPhysics.hh"
 #include "Geant4/G4IonPhysics.hh"
 #include "Geant4/G4StoppingPhysics.hh"
-#include "Geant4/G4HadronElasticPhysicsHP.hh"
+#include "Geant4/G4HadronElasticPhysics.hh"
 // #include "Geant4/G4NeutronTrackingCut.hh"
-#include "Geant4/G4HadronPhysicsQGSP_BERT_HP.hh"
+#include "Geant4/G4HadronPhysicsQGSP_BERT.hh"
 
 
 // Register this new physics list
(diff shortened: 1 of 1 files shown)


jzettle:
Hi @marcodeltutto, thanks for this. Is the idea we need to change https://github.com/SBNSoftware/icaruscode/blob/ffa6d4167b9d44f86c1e79300e1c646c4eb730ce/fcl/g4/larg4_services_icarus.fcl#L11 to "QGSP_BERT_HP"? I'm trying to understand which would become the default in that case, what is written in sbncode or what is written in icaruscode/sbndcode? I can try to make this change later today I have a couple other things to do before I can get to this and do the same comparison test that you did

marcodeltutto:
Thanks, @jzettle! Yes, I believe the only thing to be changed is that one.

The default is what is specified in `PhysicsListName` for SBND and ICARUS. In `sbncode` there is no default, but we can add custom physics lists in `sbncode/LArG4/PhysicsLists/` and then point to them in `sbndcode` and `icaruscode`.

In the future, we can think about moving `larg4_services_icarus.fcl` and `larg4_services_sbnd.fcl` to `sbncode` so we both use the same config.

jzettle:
Thanks Marco! Sounds good, I wanted to better understand the idea of this change. I agree we should figure out how to at some point move truly common pieces of the simulation to sbncode so it is more synchronized between the two detector simulations

jzettle:
@marcodeltutto I created https://github.com/SBNSoftware/icaruscode/pull/737 to address the same as you did on sbndcode and edited your initial post to include that. Looks like it is making the intended change when comparing the old and new versions and what Geant4 is loading looking at the neutron as a comparison

marcodeltutto:
Sounds great, thanks @jzettle!

ibsafa:
@cerati @sjgardiner Do we want this physics list update in the next production campaign? If so, we need folks to approve this PR ASAP. sbndcode already has this change merged. icaruscode also has a similar PR that is already approved, but not merged. Thanks!

jzettle (approved):
With the other SBND and ICARUS PRs combined mentioned in Marco's initial comment with this one it looks good. Apologies I thought I had approved this at some point

PR #463, opened 2024-05-07
last valid point for last segment validity

wjdanswjddl:
fix that addresses [comment](https://github.com/SBNSoftware/sbncode/pull/384/files#r1589798179)


PetrilloAtWork (approved):
I think it's good.
I don't grasp exactly the point in having this last step, but that's an algorithm detail I can trust the author(s) about.

ibsafa:
trigger build

PR #468, opened 2024-06-19
Add t0CRT to Calibration Ntuple

afm1g15:
Add the t0CRT to the calibration ntuple. Needs to have to equivalent pull request in sbnobj [(here)](https://github.com/SBNSoftware/sbnobj/pull/111) also present to work. Adds this extra t0 from Henry's CRT association into the ntuple next to the cathode crossing t0. Also unpicks the Pandora assumed trigger correction if only t0CRT tagged.


mrmooney (approved):
Looks good to me

ibsafa:
trigger build

PR #477, opened 2024-10-17
Fix how CRT-PMT entries are stored in CAFs

aheggest:
### Description 
This pull request fixes a bug in how the CRT-PMT Matched information was stored in the CAFs. 

When there is no CRT-PMT match (flashClassification==0), we should have no CRT Hit match information (or`PMTTimeDiff`, `time`, `sys`, `region`, `position` values equal to their defaults, but it seems that CRT Hit match values were getting filled in from later CRT Hit matches. This is fixed by adding `srmatch.matchedCRTHits.clear();` which results in the CRT Hit match variables getting set to their defaults defined in the [StandardRecord](https://github.com/SBNSoftware/sbnanaobj/blob/develop/sbnanaobj/StandardRecord/SRCRTPMTMatch.cxx). 

When we have a PMT Flash matched to more than one CRT Hits (flashClassification==3, flashClassification==6, flashClassification==7), we should have entries saved from both CRT hit matches (with two CRT-PMT Match entries being saved for a single flash). This is fixed by calling `FillCRTPMTMatch` once for every CRT Hit in the matchedCRTHits vector.


francescopoppi:
Is this PR still open? Or is it outdated and the new one is PR 480?

aheggest:
> Is this PR still open? Or is it outdated and the new one is PR 480?

Hi @francescopoppi , this PR is the develop version of PR 480

ibsafa:
trigger build

sbncode/CAFMaker/FillReco.cxx
```diff
@@ -138,11 +138,11 @@ namespace caf
   }
 
   void FillCRTPMTMatch(const sbn::crt::CRTPMTMatching &match,
+		       const sbn::crt::MatchedCRT &crtmatch,
```
PetrilloAtWork, line 141: Given that the CRT match may or may not be there, and if it's not there it should not be used, I recommend a more explicit interface. Two options:
1. a C pointer:
```suggestion
		       const sbn::crt::MatchedCRT * crtmatch,
```
that is assigned to the address of the match (`&crthitmatch`) if there is one, or to `nullptr` if there is no match.

2. a `std::optional` object (`#include <optional>`):
```suggestion
		       std::optional<sbn::crt::MatchedCRT> crtmatch,
```
which is assigned a `crthitmatch` object when available, or `std::nullopt` otherwise. A disadvantage of this one is that it copies the object, while the pointer does not. Semantically, `optional` is the most correct approach, i.e. if I read the signature of that function I _know_ that `crtmatch` may or may not be there, while with a C pointer I don't know whether it's acceptable or not for it not to be there — and with a reference, as it is now, I know it _must_ be there.

Both pointer and `optional` have a similar interface (access elements with member access operator `->`, check if it's valid or not by implicit or explicit conversion to `bool`, like in `if (crtmatch) ...`).

PetrilloAtWork (changes requested):
I feel I am missing something fundamental.

The first stated problem is that non-matches (matches with no CRT hits) are not been saved, thus breaking a 1-to-1 relation with either the original matching objects or the flash objects (not sure here... what is the protocol?). The PR restores the 1-to-1 relation.

The second stated problem is that there is a single match for flashes sharing multiple (supposedly two) hits, and here it is proposed to have as many as the hits. Thus the PR immediately breaks the 1-to-1 relation again. In addition, the matches of the same flash to different hits appear to be all identical except for the hit they are associated to, suggesting that no new information is added by the changes that was not there already.

I have left _technical_ comments in the commits, but I would like to better understand the rationale of these two changes.

aheggest:
Closing this Pull Request because I realized this is not a true bug, but instead an issue in how I was accessing the information in the CAFs. I was using ROOT's `Scan()` and `Draw()` commands to do some simple checks, but this led to some inconsistencies between the matched Flash and the associated matched CRT hit. If I instead use the `cafe` framework, it seems these issues are taken care of by using the lengths of the matched vectors directly, and the associations between the matched CRT hit and the matched Flash are as expected. 
I will open a new PR to reassign the FlashID while filling the CAFs so the flashClassification is always correct.

PR #459, opened 2024-04-26
Saving GENIE tree to caf [develop]

jedori0228:
This is a copy (cherry-picked) of https://github.com/SBNSoftware/sbncode/pull/426 against develop.

This PR requires https://github.com/SBNSoftware/sbnanaobj/pull/128

PR #478, opened 2024-10-22
Saving GENIE tree to caf [release/SBN2024A]

jedori0228:
This is a copy of https://github.com/SBNSoftware/sbncode/pull/459 (on develop) or https://github.com/SBNSoftware/sbncode/pull/426 (on release/SBN2023A_NuMI).

Saving `genie::EventRecord` as a separate `TTree` into `CAFs`. By default we do not save it, and we turn this feature by adding `physics.producers.cafmaker.SaveGENIEEventRecord: true` in fhicls.
The matching between a `SRTrueInteraction` object from `recTree` and `genie::EventRecord` is done by the index of genie tree, `genie_evtrec_idx`, which is introduced in https://github.com/SBNSoftware/sbnanaobj/pull/131.

This PR requires https://github.com/SBNSoftware/sbnanaobj/pull/131.


jacoblarkin (approved):
Thanks Jaesung!

sbncode/CAFMaker/CMakeLists.txt
```diff
@@ -43,6 +43,8 @@ art_make_library( LIBRARY_NAME sbncode_CAFMaker
                   systematicstools::interface
                   systematicstools::interpreters
                   systematicstools::utility
+                  ${GENIE_LIB_LIST}
```
PetrilloAtWork, line 46: Where is this variable defined? is it necessary?
You may also ask CMake to print its value (I think something like `message(STATUS "GENIE_LIB_LIST='${GENIE_LIB_LIST}'")`).
If this variable is empty or unnecessary, we should not put it (they often go viral).

PetrilloAtWork (changes requested):
All good except for a question on the `CMakeLists.txt`.

sbncode/CAFMaker/CMakeLists.txt
PetrilloAtWork: Thank you for finding that. It's not really a standard pattern (in fact, I question it), but it's valid.

PetrilloAtWork (approved):
Approved after clarification by the requester.

etworcester (approved):
Thanks (and sorry for the delay - I missed that I was tagged on this one)

PR #480, opened 2024-10-28
Fix how CRT-PMT entries are stored in CAFs [release/SBN2024A]

aheggest:
### Description 
This pull request fixes a bug in how the CRT-PMT Matching information was stored in the CAFs. 
When there is no CRT-PMT match (flashClassification==0), we should have no CRT Hit match information (orPMTTimeDiff, time, sys, region, position values equal to their defaults, but it seems that CRT Hit match values were getting filled in from later CRT Hit matches. This is fixed by adding srmatch.matchedCRTHits.clear(); which results in the CRT Hit match variables getting set to their defaults defined in the [StandardRecord](https://github.com/SBNSoftware/sbnanaobj/blob/develop/sbnanaobj/StandardRecord/SRCRTPMTMatch.cxx).

When we have a PMT Flash matched to more than one CRT Hits (flashClassification==3, flashClassification==6, flashClassification==7), we should have entries saved from both CRT hit matches (with two CRT-PMT Match entries being saved for a single flash). This is fixed by calling FillCRTPMTMatch once for every CRT Hit in the matchedCRTHits vector.

This sbncode pull request addresses a separate issue than the bug fix in [icaruscode PR #760](https://github.com/SBNSoftware/icaruscode/pull/760) that updates the flashTime used to assign a CRT-PMT match classification.

### DIFF
--- a/sbncode/CAFMaker/CAFMaker_module.cc
+++ b/sbncode/CAFMaker/CAFMaker_module.cc
@@ -1471,30 +1471,43 @@ void CAFMaker::produce(art::Event& evt) noexcept {
       }
     }
 
-  // Get all of the CRTPMT Matches ..
+  // Get all of the CRTPMT Matches
   std::vector<caf::SRCRTPMTMatch> srcrtpmtmatches;
-  std::cout << "srcrtpmtmatches.size = " << srcrtpmtmatches.size() << "\n";
   art::Handle<std::vector<sbn::crt::CRTPMTMatching>> crtpmtmatch_handle;
   GetByLabelStrict(evt, fParams.CRTPMTLabel(), crtpmtmatch_handle);
   if(crtpmtmatch_handle.isValid()){
-    std::cout << "valid handle! label: " << fParams.CRTPMTLabel() << "\n";
     const std::vector<sbn::crt::CRTPMTMatching> &crtpmtmatches = *crtpmtmatch_handle;
     for (unsigned i = 0; i < crtpmtmatches.size(); i++) {
+      int topen = 0, topex = 0, sideen = 0, sideex = 0;
+      size_t startIndex = srcrtpmtmatches.size();
       if(!crtpmtmatches[i].matchedCRTHits.empty()){
 	for(const auto& crthitmatch :  crtpmtmatches[i].matchedCRTHits){
 	  srcrtpmtmatches.emplace_back();
-	  FillCRTPMTMatch(crtpmtmatches[i], crthitmatch, srcrtpmtmatches.back());
+	  FillCRTPMTMatch(crtpmtmatches[i], crthitmatch, 
+			  topen, topex, sideen, sideex,
+			  srcrtpmtmatches.back());
 	}
       }
       else{
 	sbn::crt::MatchedCRT dummy_CRTmatch;
 	srcrtpmtmatches.emplace_back();
-	FillCRTPMTMatch(crtpmtmatches[i], dummy_CRTmatch, srcrtpmtmatches.back());
+	FillCRTPMTMatch(crtpmtmatches[i], dummy_CRTmatch, 
+			topen, topex, sideen, sideex,
+			srcrtpmtmatches.back());
+      }
+      if (crtpmtmatches[i].matchedCRTHits.size()>1){
+        for (size_t j = startIndex; j < srcrtpmtmatches.size(); ++j) {
+          if (topen == 1 && sideen == 0 && topex == 0 && sideex == 1)
+            srcrtpmtmatches[j].flashClassification = 3;
+          else if(topen >= 1 && sideen == 0 && topex == 0 && sideex == 0)
+            srcrtpmtmatches[j].flashClassification = 6;
+          else if (topen >= 1 && sideen == 0 && topex == 0 && sideex >= 1)
+            srcrtpmtmatches[j].flashClassification = 7;
+          else
+            srcrtpmtmatches[j].flashClassification = 9;
+        }
       }
     }
-  }
-  else{
-    std::cout << "crtpmtmatch_handle.isNOTValid!\n";
   }
 
   // Get all of the OpFlashes
--- a/sbncode/CAFMaker/CAFMaker_module.cc
+++ b/sbncode/CAFMaker/CAFMaker_module.cc
@@ -1480,8 +1480,17 @@ void CAFMaker::produce(art::Event& evt) noexcept {
     std::cout << "valid handle! label: " << fParams.CRTPMTLabel() << "\n";
     const std::vector<sbn::crt::CRTPMTMatching> &crtpmtmatches = *crtpmtmatch_handle;
     for (unsigned i = 0; i < crtpmtmatches.size(); i++) {
-      srcrtpmtmatches.emplace_back();
-      FillCRTPMTMatch(crtpmtmatches[i],srcrtpmtmatches.back());
+      if(!crtpmtmatches[i].matchedCRTHits.empty()){
+	for(const auto& crthitmatch :  crtpmtmatches[i].matchedCRTHits){
+	  srcrtpmtmatches.emplace_back();
+	  FillCRTPMTMatch(crtpmtmatches[i], crthitmatch, srcrtpmtmatches.back());
+	}
+      }
+      else{
+	sbn::crt::MatchedCRT dummy_CRTmatch;
+	srcrtpmtmatches.emplace_back();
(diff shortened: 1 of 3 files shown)


sbncode/CAFMaker/FillReco.cxx
```diff
@@ -155,16 +156,28 @@ namespace caf
...
+    caf::SRMatchedCRT matchedCRT;
+    if(!match.matchedCRTHits.empty()){
+      matchedCRT.PMTTimeDiff = crtmatch.PMTTimeDiff; 
+      matchedCRT.time = crtmatch.time;
+      matchedCRT.sys = crtmatch.sys;
+      matchedCRT.region = crtmatch.region;
+      matchedCRT.position = SRVector3D(crtmatch.position.X(), crtmatch.position.Y(), crtmatch.position.Z());
+      std::cout << "CRTPMTTimeDiff = "<< matchedCRT.PMTTimeDiff << "\n";
+    }
+    std::cout << "CRTPMTTimeDiff = "<< matchedCRT.PMTTimeDiff << "\n";
```
francescopoppi, line 169: Quick question: What would line 169 return if match.matchedCRTHits is empty?
Also in general I would move the std::cout to a mf::Log

sbncode/CAFMaker/FillReco.cxx
```diff
@@ -155,16 +156,28 @@ namespace caf
...
+     
       caf::SRMatchedCRT matchedCRT;
       matchedCRT.PMTTimeDiff = matchedCRTHit.PMTTimeDiff; 
       matchedCRT.time = matchedCRTHit.time;
       matchedCRT.sys = matchedCRTHit.sys;
       matchedCRT.region = matchedCRTHit.region;
       matchedCRT.position = SRVector3D(matchedCRTHit.position.X(), matchedCRTHit.position.Y(), matchedCRTHit.position.Z());
       srmatch.matchedCRTHits.push_back(matchedCRT);
-    }
+      }*/
```
francescopoppi, line 180: probably is better to remove outdated code and not comment out, unless it is expected to be re-addressed in the future.

francescopoppi:
I made two minor request. One is aesthetic, the second is just a clarification for the expected output. Other than this good to go

sbncode/CAFMaker/FillReco.cxx
aheggest: if match.matchedCRTHits is empty, the matchedCRT variables are set to their defaults defined in [StandardRecord/SRCRTPMTMatch.cxx](https://github.com/SBNSoftware/sbnanaobj/blob/develop/sbnanaobj/StandardRecord/SRCRTPMTMatch.cxx#L11), so matchedCRT.PMTTimeDiff would return `std::numeric_limits<double>::lowest()`

sbncode/CAFMaker/CAFMaker_module.cc
```diff
@@ -1471,22 +1471,44 @@ void CAFMaker::produce(art::Event& evt) noexcept {
...
+	}
+      }
+      else{
+	sbn::crt::MatchedCRT dummy_CRTmatch;
+	srcrtpmtmatches.emplace_back();
+	FillCRTPMTMatch(crtpmtmatches[i], dummy_CRTmatch, 
+			topen, topex, sideen, sideex,
+			srcrtpmtmatches.back());
+      }
+      if (crtpmtmatches[i].matchedCRTHits.size()>1){
```
francescopoppi, line 1498: is it supposed to be strictly >1 or >= 1?

francescopoppi:
Hi @aheggest thank you! Everything looks ok, thanks again for introducing this bug fix. I just had a question on this "if (crtpmtmatches[i].matchedCRTHits.size()>1)". Should it be >= 1 ? If you confirm me that I misinterpreted, this PR is ok to go!

francescopoppi (approved):
Approved

PetrilloAtWork (changes requested):
It looks like my review got lost.
I am not trying to rewrite everything, but I'll focus on the most important point.
I am confused by two apparently contradictory changes.
The first is the introduction of default values when a match is not successful. I am assuming that what leads the match is a flash and that the idea is that there is one entry per flash, even if there is no match.
The second is the introduction of multiple matches for the same matching object. This appears to break the protocol that the first change achieved, since now there is again not a 1-to-1 correspondence between match results and flashes. On top of this, I don't fully get the motivation for having multiple matching entries each with a different CRT hit, and otherwise identical, vs. having a single matching object with a list of CRT hits.
Please clarify.

aheggest:
hi @PetrilloAtWork , thank you for your comment! hopefully i can clarify here the purpose of these changes. 

Regarding your first point, I found that when there is no CRT-PMT Match, there is only information being saved to the flash variables, and not to the CRT hit variables in the CRT-PMT match. This seemed to work fine in the calib ntuples, and the CRT related variables are just empty if there is no CRT-PMT match. But it seems in the CAFs those variables were not getting left empty/set to defaults if there is no CRT hit match: it seemed that somehow the CRT information from the following CRT-PMT matches were getting saved, resulting in an incorrect association between the flash information and the CRT information. See, for example, here is what was getting stored before
```***********************************************************************************
*    Row   * Instance * hdr.evt * match.flashID * matchID * CRTRegion * CRT-PMTTime *
***********************************************************************************
*        0 *        0 *     28354 * -97647192 *         2 *        44 *   -7.0721 *
*        0 *        1 *     28354 * -88408192 *         1 *        30 *  -31.0721 *
*        0 *        2 *     28354 * -86476192 *         1 *        30 *  -28.0721 *
*        0 *        3 *     28354 * -74332192 *         1 *        30 *  -35.0721 *
*        0 *        4 *     28354 * -67618192 *         1 *        30 *  -17.0721 *
*        0 *        5 *     28354 * -61628192 *         1 *        32 *  -24.0721 *
*        0 *        6 *     28354 * -57480192 *         0 *        30 *  -23.0721 *
*        0 *        7 *     28354 * -51348192 *         0 *        30 *  -28.0721 *
*        0 *        8 *     28354 * -50540192 *         1 *        30 *  -29.0721 *
*        0 *        9 *     28354 * -45043192 *         0 *        42 *    4.9279 *
*        0 *       10 *     28354 * -44903192 *         1 *        30 *  -21.0721 *
```

(notice the flashes with MatchID = 0, aka no CRT-PMT Match, the CRT related variables `region` and `PMTTimeDiff` are still filled)
With my edits, below is what i now see in the CAFs:
```***********************************************************************************
*    Row   * Instance * hdr.evt * match.flashID * matchID * CRTRegion * CRT-PMTTime *
* ************************************************************************************
*        0 *        0 *     28354 * -97647192 *         2 *        44 *   -7.0721 *
*        0 *        1 *     28354 * -88408192 *         1 *        30 *  -31.0721 *
*        0 *        2 *     28354 * -86476192 *         1 *        30 *  -28.0721 *
*        0 *        3 *     28354 * -74332192 *         1 *        30 *  -35.0721 *
*        0 *        4 *     28354 * -67618192 *         1 *        30 *  -17.0721 *
*        0 *        5 *     28354 * -61628192 *         1 *        32 *  -24.0721 *
*        0 *        6 *     28354 * -57480192 *         0 *        -1 *      -inf *
*        0 *        7 *     28354 * -51348192 *         0 *        -1 *      -inf *
*        0 *        8 *     28354 * -50540192 *         1 *        30 *  -23.0721 *
*        0 *        9 *     28354 * -45043192 *         0 *        -1 *      -inf *
*        0 *       10 *     28354 * -44903192 *         1 *        30 *  -28.0721 *
```
Notice now that the flashes with `MatchID=0` now have the CRT variables set to their defaults, and `flashID=-50540192` , for example, is getting associated to the correct `PMTTimeDiff=-23.0721` (notice above the `PMTTimeDiff=-23.0721` was _incorrectly_ getting associated to `flashID=-57480192`). So this is why i included the default variables getting set when there is no CRT-PMT Match. 

Regarding your second point about having multiple matches saved for the same match object (or same flashID), i noticed that when we have >1 CRT hit matched to a flash (for ex, matchID=3,6,7), only information from the later CRT hit was getting saved. To fix this, i was first trying to introduce a vector of matched CRT hits for each match object (each unique flashID would have a vector of CRT Hit matches, and that vector would be zero if there is no CRT-PMT Match), but i had issues implementing this and found the simplest way forward was to introduce a new entry for each CRT-PMT Match. I'm trying to remember what issues i was facing exactly, I think it was just getting tricky figuring out where to call FillCRTPMTMatch, but if you prefer i could try to instead have a list (vector?) of CRT hit matches for each CRT-PMT match (maybe asking for some help from you on implementing it if i cant figure it out). This is what we have in the calibration ntuples

PR #482, opened 2024-11-11
reassign CRT-PMT MatchID while filling CAFs

aheggest:
### Description 
This pull request reassigns the CRT-PMT Match flash Classification in the CAFs. There was a bug found in the CRT-PMT Matching producer where the classification of the flash that gets assigned from the CRT-PMT Time difference value was using the incorrect flashTime. This is now fixed in [icaruscode PR #760](https://github.com/SBNSoftware/icaruscode/pull/760). Since the CRT-PMT Time difference value is also available in the CAFs, @francescopoppi  suggested we have the ability to re-evaluate the flashClassification while filling the CAFs, instead of just reading the `sbn::crt::CRTPMTMatching match.flashClassification` value into the CAFs. With this, we don't need to rerun stage0/stage1 to get the correct CRT-PMT flashClassification, only the CAFMaker needs to be ran which will now save the correct flashClassification. 

I've also moved the logic of the [flashClassification assignment](https://github.com/SBNSoftware/icaruscode/blob/develop/icaruscode/CRT/CRTUtils/CRTPMTMatchingUtils.cxx) to a function in `sbnobj`, which can also be accessed from the CRTPMTMatchingProducer in icaruscode, as suggested by Gianluca in his [review](https://github.com/SBNSoftware/sbncode/pull/477#discussion_r1826355539). This PR thus requires [sbnobj PR #114](https://github.com/SBNSoftware/sbnobj/pull/114) to be merged in so that function can be called. 

- [] Is this PR related to an open issue / project?
- [] Does this PR affect CAF data format? If so, please assign a CAF maintainer as additional reviewer.
- [x] Does this PR require merging another PR in a different repository (such as sbnanobj/sbnobj etc.)? If so, please link it in the description.
- [X] Are you submitting this PR on behalf of someone else who made the code changes? If so, please mention them in the description.


francescopoppi:
Another comment is that the Labels is wrong, it is not a bug, the but was at the producer level. This is probably an enhancement

PetrilloAtWork (approved):
Looks ok to me.
"Returning" those four/six counts seems to be an unused feature so far — I am assuming you expect it might come handy some day.
Usually I recommend using a `struct` rather than many single values, since it's easier to mess with them when calling the function.
In this case the `struct` might have looked like:
```cpp
/// Counts of matched hits and their "direction".
struct MatchedHitCounts {
  unsigned int topen = 0;    ///< Entering from top CRT.
  unsigned int sideen = 0;   ///< Entering from side CRT.
  unsigned int bottomen = 0; ///< Entering from bottom CRT.
  unsigned int topex = 0;    ///< Exiting from top CRT.
  unsigned int sideex = 0;   ///< Exiting from side CRT.
  unsigned int bottomex = 0; ///< Exiting from bottom CRT.
}; // MatchedHitCounts
```
and then you pass one of those, e.g.
```cpp
MatchedHitCounts counts;
FillCRTPMTMatch(crtpmtmatches[i], counts, srcrtpmtmatches.back());
```

PR #485, opened 2024-11-12
Misc bnb retriever changes

nathanielerowe:
Various updates to BNBRetriever modules. Some changes are relevant for both detectors. Open to discussion on the renaming of the original BNBRetriever, just felt like it would be bad to have SBNDBNBRetriever while also having a seemingly generic BNBRetriever. Requires updated PTB information from [sbndaq_artdaq_core/v01_10_03](https://github.com/SBNSoftware/sbndaq-artdaq-core/tree/release/v1_10_03) as well as [sbndaq_artdaq/v01_10_03](https://github.com/SBNSoftware/sbndaq-artdaq/tree/release/v1_10_03).

Changes:                                               
- Introduce a new module which handles PTB fragment data.
- Fix a bug in the original BNBRetriever module used in ICARUS
- Rename module used in ICARUS to distinguish between the two BNBRetrievers


nathanielerowe:
Will utilize this PR for ICARUS producer module reorganization and bug fix following sbnd producer module implementation.

PR #492, opened 2024-12-10
Added SBN (shared) GENIE configuration for BNB and NuMI

PetrilloAtWork:
## Description

This request stores in `sbncode` a configuration template for `GENIEGen` module that include common settings like the time structure of the beam, its intensity, ν-Ar cross sections, and a few sparse others. Conversely, flux and simulation volumes are intended to be specified by the derivative configurations (although the latter does have a default volume specified).

Both BNB and NuMI beams are described.

This PR is a dependency of SBNSoftware/icaruscode#785: ICARUS is going to use this PR as base configuration. 

### Choice of parameters

The configuration almost completely replicates the one from SBND (`genie_sbnd.fcl`), although the difference with the current ICARUS is almost completely insignificant beside the enabling of beam time structure, which is the issue that sparked this PR.

For the beam timing, SBND figures were used; the bunch period matches the one measured by ICARUS, and the width of the bunch is... well, probably too narrow, but I don't have a "naked" measure of it (even MicroBooNE's folds detector effects into it).

For the beam intensity, I stuck to the "official" values used in the SBN proposal, which were reasonably close to what MicroBooNE had seen but that are hopelessly higher than what SBN detector have observed, typically by a factor ~1.5. Nevertheless, changing these figures needs to be discussed by the experiments and especially by SBND, which is the most affected (as long as events include only one interaction, intensity changes can be approximated by just scaling the total exposure; but SBND does see interaction pile up).

### Details

The new configuration fragments are all in a single `genie_beam_settings.fcl`.
It is structured in a timing section where the beam time structure is described, followed by a section with templates for `GENIEGen` configuration. Each of the section is split between BNB and NuMI.
Also, some beam time structures configuration fragments there are meant to be the default/current setting (e.g. `beam_parameters_BNB`), and others to be immutable snapshots of the state at a certain time (e.g. `beam_parameters_BNB_2024`).

You may thing of `beam_parameters_NuMI_2024` as an internal, sad sad joke.

## Review

I summon for review:
 * @leoaliaga, as generator expert, member of ICARUS, and SBN simulation coordinator
 * @marcodeltutto, as member of SBND, and SBN simulation coordinator (ok, maybe also as generator expert?)

I invite the reviewers to add more reviewers as they feel fit.

I tested the BNB and NuMI configurations in ICARUS; I did not attempt to modify SBND configurations to use these configuration templates.
I would like the reviewers to also express about the parameters I removed compared to [SBND configuration](sbndcode/LArSoftConfigurations/gen/genie_sbnd.fcl), which I believe I have identified as unused (and seriously misleading; one example: `BeamCenter: [ 350, 1400, 0 ]`).

I would also like **SBND to consider to use this configuration as base for its GENIE jobs**.

One presentation about this change was given to ICARUS event selection working group ([SBN DocDB 39065](https://sbn-docdb.fnal.gov/cgi-bin/sso/ShowDocument?docid=39065)). I think I'll give a similar one in a SBN analysis infrastructure meeting (same document number), but I am not sure when the next one will be called — and I would prefer not to wait for that.


marcodeltutto:
Thanks, @PetrilloAtWork! I agree SBND should also use these common settings. I'll take a look at your settings soon so we have this PR merged quickly, but it may take a bit longer for SBND to use it. SBND should review what changes this brings us, especially waht you noted about the beam intensity, before we can adopt it.

marcodeltutto:
Hi @PetrilloAtWork, I took a deeper look and compared with the current settings uded in SBND (using `prodoverlay_corsika_cosmics_proton_genie_rockbox_sce.fcl` which is the fcl for our current CV MC). I think things looks good and there is no actual change for SBND, other than a better and cleaner fcl! Thanks very much. We will make an SBND PR to have our gen fcls be based on this one (SBND issue opened).

My only question is about the intensity change you mention in the description. We are already using 5e12 POT per spill, which seems the same as in this PR. Or are you referrerring to something else?




Parameter | Current SBND | Proposed |Comment|
-----------|---------------|--------------|-------------|
AddGenieVtxTime | true | true | OK |
BeamCenter | [-1400, -350, 0] | [ 0, 0, 0 ] | Good to change, used if we had a histogram-based flux |
BeamDirection | [0, 0, 1] | [ 0, 0, 1 ] | Good to change, used if we had a histogram-based flux |
BeamName | "booster" | "booster" | OK |
BeamRadius | 3 | | Good to change, used if we had a histogram-based flux |
DebugFlags | 0 | 0 | OK |
DefinedVtxHistRange | true | false | Good to change, used if we had a histogram-based flux| 
DetectorLocation | "MINOS-NearDet" | "MINOS-NearDet" | Shall we change it? |
Environment | [ ] | [ ] | OK | 
EventGeneratorList | "Default" | "Default" | OK |
EventsPerSpill | 0 | @erase | Is @erase == 0? |
FiducialCut | "rockbox:(-210,-210,-10)(210,210,510),0,500,0.00425,1.3" | "none" | OK, as SBND used rockbox |
FluxCopyMethod | "DIRECT" | | OK, as it should not be in common settings |
FluxFiles | "..." | | OK, as it should not be in common settings |
FluxSearchPaths | "..." | | OK, as it should not be in common settings |
FluxType | "simple_flux" | | OK, as it should not be in common settings |
FluxUpstreamZ | -18 | | OK, as it should not be in common settings |
GenFlavors | [12, 14, -12, -14] |  [ 12, 14, -12, -14 ] | OK |
GeomScan | "file: GENIE/sbnd_rock_maxpathlength_fluxI_gdmlv02_00.xml" | | OK, as SBND used rockbox |
GlobalTimeOffset | 0 | 0 | OK |
MaxFluxFileMB | 100 | | OK, as it should not be in common settings |
MixerBaseline | 0 | 0 | OK |
MixerConfig | "none" | "none" | OK |
MonoEnergy | 2 | | OK, we should not set options for FluxType = "mono" |
POTPerSpill | 5e12 | 5e12 | OK |
PassEmptySpills | false | false | OK |
ProductionMode | "yes" | "yes" | OK |
RandomTimeOffset | 10000 | | OK |
SpillTimeConfig | "evgb::EvtTimeFNALBeam  booster,  dtbucket=18.936, sigma=1.308" | " evgb::EvtTimeFNALBeam booster nperbatch  84 nfilled    81 intensity  1.0 dtbucket  18.936 sigma      1.308 global     0.0 "| OK, as nperbatch  84 nfilled    81 intensity  1.0 are defaults |
SurroundingMass | 0 | | OK, we shouldn't set this, it's for fluxes that don't know pot normalization (???)|
TopVolume | "volWorld" |  "volDetEnclosure" | OK to have volDetEnclosure as common settings, and then override in exp code |
VtxPosHistRange | [-210, 210, -210, 210, -10, 510] | [ ] | Good to change, used if we had a histogram-based flux |
XSecTable | "gxspl-FNALsmall.xml" | "gxspl-FNALsmall.xml" | OK |
GHepPrintLevel | | -1 | OK, -1 is no print |

ibsafa:
@marcodeltutto If you are satisfied with @PetrilloAtWork answer, can you approve the PR? if not, what else would you like clarification on or changes to this?

marcodeltutto:
@ibsafa I had a few more questions above but I prefer to get this going since those are minor things. Approving now.

marcodeltutto (approved):
Thank you!

PR #495, opened 2024-12-13
SBNB POT Accounting Producer Modules

nathanielerowe:
Splitting the PR for POT Accounting into two components. An additional PR will follow this which fixes a bug in  the ICARUS POT accounting and reorganizes the directory a bit.

Requires updated PTB information from [sbndaq_artdaq_core/v01_10_03](https://github.com/SBNSoftware/sbndaq-artdaq-core/tree/release/v1_10_03) as well as [sbndaq_artdaq/v01_10_03](https://github.com/SBNSoftware/sbndaq-artdaq/tree/release/v1_10_03).

This PR introduces three producer modules, summarized in presentation at [SBND Dec 2024 Collab Meeting](https://sbn-docdb.fnal.gov/cgi-bin/sso/ShowDocument?docid=39161).

Remaining issues were determined to be beam related. Behavior is stable with new BNBZeroBias stream.

### DIFF
--- a/sbncode/BeamSpillInfoRetriever/SBNDBNBEXTRetriever/SBNDBNBEXTRetriever_module.cc
+++ b/sbncode/BeamSpillInfoRetriever/SBNDBNBEXTRetriever/SBNDBNBEXTRetriever_module.cc
@@ -1,0 +1,230 @@
+////////////////////////////////////////////////////////////////////////
+// Class:       SBNDBNBEXTRetriever
+// Plugin Type: producer 
+// File:        SBNDBNBEXTRetriever_module.cc
+//
+////////////////////////////////////////////////////////////////////////
+
+#include "art/Framework/Core/EDProducer.h"
+#include "art/Framework/Core/ModuleMacros.h"
+#include "art/Framework/Principal/Event.h"
+#include "art/Framework/Principal/Handle.h"
+#include "art/Framework/Principal/Run.h"
+#include "art/Framework/Principal/SubRun.h"
+#include "canvas/Utilities/InputTag.h"
+#include "fhiclcpp/ParameterSet.h"
+#include "messagefacility/MessageLogger/MessageLogger.h"
+
+#include <memory>
+#include <bitset>
+#include <tuple>
+#include <algorithm>
+
+#include "sbndaq-artdaq-core/Overlays/SBND/PTBFragment.hh"
+#include "sbndaq-artdaq-core/Overlays/SBND/TDCTimestampFragment.hh"
+#include "artdaq-core/Data/ContainerFragment.hh"
+//#include "sbndcode/Decoders/PTB/sbndptb.h"
+#include "sbnobj/Common/POTAccounting/EXTCountInfo.h"
+
+#include "ifdh_art/IFBeamService/IFBeam_service.h"
+#include "ifbeam_c.h"
+
+#include "larcorealg/CoreUtils/counter.h"
+
+namespace sbn {
+  class SBNDBNBEXTRetriever;
+}
+
+class sbn::SBNDBNBEXTRetriever : public art::EDProducer {
+public:
+  explicit SBNDBNBEXTRetriever(fhicl::ParameterSet const & params);
+  // Required functions.
+  void produce(art::Event & e) override;
+  void beginSubRun(art::SubRun& sr) override;
+  void endSubRun(art::SubRun& sr) override;
+
+  // Plugins should not be copied or assigned.
+  SBNDBNBEXTRetriever(SBNDBNBEXTRetriever const &) = delete;
+  SBNDBNBEXTRetriever(SBNDBNBEXTRetriever &&) = delete;
+  SBNDBNBEXTRetriever & operator = (SBNDBNBEXTRetriever const &) = delete;
+  SBNDBNBEXTRetriever & operator = (SBNDBNBEXTRetriever &&) = delete;
+
+
+private:
+  // Declare member data here.
+  std::vector< sbn::EXTCountInfo > fOutExtInfos;
+  struct PTBInfo_t {
+    double currPTBTimeStamp  = 0;
+    double prevPTBTimeStamp  = 0;
+    unsigned int GateCounter = 0; // FIXME needs to be integral type
+  };
+
+  struct TriggerInfo_t {
+    double t_current_event  = 0;
+    double t_previous_event = 0;
+    unsigned int number_of_gates_since_previous_event = 0; // FIXME needs to be integral type
+  };
+
+  TriggerInfo_t extractTriggerInfo(art::Event const& e) const;
+  PTBInfo_t extractPTBInfo(art::Handle<std::vector<artdaq::Fragment> > cont_frags) const;
+  double extractTDCTimeStamp(art::Handle<std::vector<artdaq::Fragment> > cont_frags) const;
+
+  // input labels
+  std::string raw_data_label_;
+  float TotalEXTCounts;  
+  float totalMinBias;
+  float evtCount;
+};
+
+sbn::SBNDBNBEXTRetriever::SBNDBNBEXTRetriever(fhicl::ParameterSet const & params)
+  : EDProducer{params} {
+  produces< std::vector< sbn::EXTCountInfo >, art::InSubRun >();
+  TotalEXTCounts = 0;
+  totalMinBias = 0;
+  evtCount = 0;
+}
+
+int eventNum =0;
+int _run;
+int _subrun;
+int _event;
+
+void sbn::SBNDBNBEXTRetriever::produce(art::Event & e)
+{
+
+
+  TriggerInfo_t const triggerInfo = extractTriggerInfo(e);
+
+  TotalEXTCounts += triggerInfo.number_of_gates_since_previous_event;
+
+  if(triggerInfo.number_of_gates_since_previous_event > 0){
+    evtCount++;  
+    totalMinBias += triggerInfo.number_of_gates_since_previous_event;
+  }
+   
+  //Store everything in our data-product
+  sbn::EXTCountInfo extInfo;
+  extInfo.gates_since_last_trigger = triggerInfo.number_of_gates_since_previous_event;
+  fOutExtInfos.push_back(extInfo);
+}
+
+sbn::SBNDBNBEXTRetriever::PTBInfo_t sbn::SBNDBNBEXTRetriever::extractPTBInfo(art::Handle<std::vector<artdaq::Fragment> > cont_frags) const {
+  int HLT_count = 0;
+
+  int numcont = 0;
+  PTBInfo_t PTBInfo;
+  for (auto const& cont : *cont_frags)
+  { 
+    artdaq::ContainerFragment cont_frag(cont);
+    numcont++;
+    int numfrag = 0;
+    for (size_t fragi = 0; fragi < cont_frag.block_count(); ++fragi)
+    {
+      numfrag++;
+      artdaq::Fragment frag = *cont_frag[fragi];
+      sbndaq::CTBFragment ctb_frag(frag);   // somehow the name CTBFragment stuck
+      for(size_t word_i = 0; word_i < ctb_frag.NWords(); ++word_i)
+      {
+        if(ctb_frag.Trigger(word_i)){
+          uint32_t wt = 0;
+          uint32_t word_type = ctb_frag.Word(word_i)->word_type;
+          wt = word_type;
+	  if (wt == 2 && ctb_frag.Trigger(word_i)->IsTrigger(4))
+	  {
+            HLT_count++;
+
+	    uint64_t RawprevPTBTimeStamp = ctb_frag.PTBWord(word_i)->prevTS * 20; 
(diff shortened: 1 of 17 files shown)


sbncode/BeamSpillInfoRetriever/SBNDBNBEXTRetriever/SBNDBNBEXTRetriever_module.cc
```diff
@@ -0,0 +1,218 @@
...
+
+
+  TriggerInfo_t const triggerInfo = extractTriggerInfo(e);
+
+  TotalEXTCounts += triggerInfo.number_of_gates_since_previous_event;
+
+  if(triggerInfo.number_of_gates_since_previous_event > 0){
+    evtCount++;  
+    totalMinBias += triggerInfo.number_of_gates_since_previous_event;
+  }
```
jzennamo, line 103: these are unused, you could remove them.

sbncode/BeamSpillInfoRetriever/SBNDBNBEXTRetriever/SBNDBNBEXTRetriever_module.cc
```diff
@@ -0,0 +1,218 @@
...
+    artdaq::ContainerFragment cont_frag(cont);
+    for (size_t fragi = 0; fragi < cont_frag.block_count(); ++fragi)
+    {
+      artdaq::Fragment frag = *cont_frag[fragi];
+      sbndaq::TDCTimestampFragment tdc_frag(frag); 
+      TDCTimeStamp = static_cast<double>(tdc_frag.getTDCTimestamp()->timestamp_ns())/1e9;
+    } //End of loop over the number of fragments per container
+  } //End of loop over the number of containers
+  return TDCTimeStamp;
+}
```
jzennamo, line 156: I think this is unused too, right?

sbncode/BeamSpillInfoRetriever/SBNDBNBRetriever/MWRData.cpp
```diff
@@ -0,0 +1,71 @@
+#include <iostream>
```
jzennamo, line 1: since this is a copy of the code in the other directory, could we move it to a shared directory that sits inside sbncode/BeamSpillInfoRetriever and everyone references that shared code?

sbncode/BeamSpillInfoRetriever/SBNDBNBRetriever/SBNDBNBRetriever_module.cc
```diff
@@ -0,0 +1,624 @@
...
+  std::vector<int> matched_MWR;
+  matched_MWR.resize(3);
+  
+  // NOTE: for now, this is dead code because we don't
+  // do anything for the first event in a run. We may want to revisit 
+  // this later to understand if there is a way we can do the POT
+  // accounting in the first event.
+  //
+  // Need to handle the first event in a run differently
+  if(isFirstEventInRun){
```
jzennamo, line 384: this doesn't look to be set anywhere, I might remove it for safety

sbncode/BeamSpillInfoRetriever/SBNDBNBRetriever/SBNDBNBRetriever_module.cc
```diff
@@ -0,0 +1,624 @@
...
+      if(times_temps[i] > (triggerInfo.t_current_event+fTimePad)){
+	spills_after_our_target++;
+      }
+    }//end loop through spill times 	 
+    
+    // Remove the spills after our trigger
+    times_temps.erase(times_temps.end()-spills_after_our_target,times_temps.end());
+    
+    // Remove the spills before the start of our Run
+    times_temps.erase(times_temps.begin(), times_temps.end() - std::min(int(triggerInfo.number_of_gates_since_previous_event), int(times_temps.size())));
```
jzennamo, line 399: I am sure I wrote this :face_palm: do you understand this logic? What is it is doing?

sbncode/BeamSpillInfoRetriever/SBNDBNBRetriever/SBNDBNBRetriever_module.cc
```diff
@@ -0,0 +1,624 @@
...
+  // mf::LogDebug("SBNDBNBRetriever") << std::setprecision(19) << "Upper Limit : " << (triggerInfo.t_current_event)+fTimePad <<  std::endl;
+  // mf::LogDebug("SBNDBNBRetriever") << std::setprecision(19) << "Lower Limit : " << (triggerInfo.t_previous_event)+fTimePad <<  std::endl;
+  
+  // Iterating through each of the beamline times
+  for (size_t i = 0; i < times_temps.size(); i++) {
+    
+    // Only continue if these times are matched to our DAQ time
+    // mf::LogDebug("SBNDBNBRetriever") << std::setprecision(19) << "Time # : " <<  i << std::endl;
+
+    if(!isFirstEventInRun){//We already addressed the "first event" above
```
jzennamo, line 416: I might drop this

sbncode/BeamSpillInfoRetriever/SBNDBNBRetriever/SBNDBNBRetriever_module.cc
```diff
@@ -0,0 +1,624 @@
...
+      DocDB 33155 provides documentation of this
+    */
+
+    // mf::LogDebug("SBNDBNBRetriever") << std::setprecision(19) << "matchMultiWireData:: trigger type : " << get_trigger_type_matching_gate(db, callback_trigger_type, run_number, times_temps[i]*1.e9-triggerInfo.WR_to_Spill_conversion+3.6e7, 40.) << " times : spill " << times_temps[i]*1.e9 << " - " << triggerInfo.WR_to_Spill_conversion << " + " << 3.6e7 <<  std::endl;
+    
+    // if(get_trigger_type_matching_gate(db, callback_trigger_type, run_number, times_temps[i]*1.e9-triggerInfo.WR_to_Spill_conversion+3.6e7, 40.) == 1){
+    //      mf::LogDebug("SBNDBNBRetriever") << std::setprecision(19)  << "matchMultiWireData:: Skipped a MinBias gate at : " << times_temps[i]*1000. << std::endl;
+
+    //  continue;
+    //}
```
jzennamo, line 440: I think it is safe to clean up all the commented out code.

sbncode/BeamSpillInfoRetriever/SBNDBNBZEROBIASRetriever/MWRData.cpp
```diff
@@ -0,0 +1,71 @@
+#include <iostream>
```
jzennamo, line 1: I would definitely move this code up and reference it in one place so we don't have to maintain the copies

sbncode/BeamSpillInfoRetriever/SBNDBNBZEROBIASRetriever/SBNDBNBZEROBIASRetriever_module.cc
```diff
@@ -0,0 +1,597 @@
...
+
+void sbn::SBNDBNBRetriever::produce(art::Event & e)
+{
+
+  // If this is the first event in the run, then ignore it
+  // We do not currently have the ability to figure out the first
+  // spill that the DAQ was sensitive to, so don't try to save any
+  // spill information
+
+  if (e.event() == 1) {
```
jzennamo, line 135: is this how SBND events work? Would the first event in the run be `e.event() ==  1`?

jzennamo (commented):
It would would be good to verify that this is doing what we expect it to do with data. Both running the modules and seeing if we can co-normalize the samples but also what happens when we run modules on the "wrong" sample?

sbncode/BeamSpillInfoRetriever/SBNDBNBEXTRetriever/SBNDBNBEXTRetriever_module.cc
nathanielerowe: Sorry for the convolution in some of these cases. They are trying to catch some issues which got passed along from the DAQ side of things. This function gets called in line 171 of SBNDBNBEXTRetriever.

The way things are currently laid out, there are two kinds of timestamps used. By default, the TDC timestamp is used for the current event and the PTB timestamps are used for the previous event. This is due to some issues with the PTB timestamps being off and was the result of some discussions from a few months ago which I can dig up from slack if necessary.

If the TDC timestamps are not present, then we use the PTB timestamp and try to correct for it being off. I was told that this bug was fixed, but to keep the catch for this failure mode in place. If this is no longer preferred let me know, and I can remove anything about using the PTB time for the current timestamp.

sbncode/BeamSpillInfoRetriever/SBNDBNBRetriever/MWRData.cpp
nathanielerowe: That would be great, and was my intention long term as well. I was under the impression that anything that interacts with the ICARUS modules should be left to a separate PR. I can rearrange it such that SBNDBNBRetriever and SBNDBNBEXTRetriever use the same file from sbncode/BeamSpillInfoRetriever. 

Would it be alright to include the change to make BNBRetreiever (which I was hoping to rename as ICARUSBNBRetriever) later?

sbncode/BeamSpillInfoRetriever/SBNDBNBRetriever/SBNDBNBRetriever_module.cc
nathanielerowe: As is, isFirstEventInRun is set in line 142 based on the result of e.event() == 1. Let me know if this is not the cleanest way to do this or if I am misunderstanding what you are pointing to here.

sbncode/BeamSpillInfoRetriever/SBNDBNBRetriever/SBNDBNBRetriever_module.cc
nathanielerowe: Gotcha, I had used some LogDebug for my own tests and had left it behind as I thought it might be of use to someone who had come along after. It does make it harder to look at though, so I will remove them.

sbncode/BeamSpillInfoRetriever/SBNDBNBZEROBIASRetriever/SBNDBNBZEROBIASRetriever_module.cc
nathanielerowe: Yeah, taking this chunk did cause issues for me in the past. In those cases, the previous event timestamp just defaults to zero when e.event() == 1.

sbncode/BeamSpillInfoRetriever/SBNDBNBRetriever/MWRData.cpp
nathanielerowe: Marking this as resolved, will include a similar change for BNBRetriever in the ICARUS update

sbncode/BeamSpillInfoRetriever/SBNDBNBRetriever/SBNDBNBRetriever_module.cc
nathanielerowe: Cleaned this up, marking as resolved.

sbncode/BeamSpillInfoRetriever/SBNDBNBRetriever/SBNDBNBRetriever_module.cc
nathanielerowe: In this function, times_temps is an vector of times in the cache associated with the specified timing device. Later in the function, we are going to match up the times in this vector with the times in the MWRdata to produce the data product that goes into the subrun.

In that times_temps, we only need to look at spills within the range of the previous event to the current event. This specific line you have highlighted is removing the spills before the start of the run. My impression was that you (or whoever) wrote this to decrease the time the loop through times_temps takes.

I would also note that it looks like this segment was left in the original module for the possibility of adding a way to handle the first event in a run. It seems like this may not be something we end up doing. If you would like, I can remove this case from the matchMultiWire function in the module and we can just have a return statement for these events in the module. I can do something similar in the ICARUS module in that PR as well.

kjplows:
Hi @nathanielerowe, just to check - did the checks @jzennamo mentioned get done? I note in DocDBs [39679](https://sbn-docdb.fnal.gov/cgi-bin/sso/ShowDocument?docid=39679) and [39601](https://sbn-docdb.fnal.gov/cgi-bin/sso/ShowDocument?docid=39601) there were some minor issues - were those part of those checks? Thanks!

nathanielerowe:
@kjplows The only outstanding check is the gate normalized top-hat plot. This required a fix to the PTB which should be going in some time today.

jzennamo (approved):
This looks fantastic! Thank you so much for all your work on this!

justinjmueller (approved):
This looks great! Thanks!

kjplows:
Brilliant, huge thanks to everyone for all your hard work! CI tests happy so prepping this for release now. 💪

PR #502, opened 2025-02-06
Lepton Meson HNL Anisotropies and HNL Polarization fix

LuisPelegrina:
In this request two main issues in the HNL part of MeVPrtlGen have been tackled:
1. HNL Polarisation calculations have been revised:
   - Polarisation calculation are now also included in Tau decays into HNLs, using the same formula as the Kaon decays, and are implemented inside Tau2HNLFlux_tool.cc
   - A bug in the polarisation calculation in the Kaon case is now solved. K^{+}->l^{+}N and K^{-} -> l^{-N} where programmed to have the same polarisation instead of opposite. This now has been fixed inside the module Kaon2HNLFlux_tool.cc.
2- The implementation of anisotropies in the l^{-/+}M^{+/-} decay is now implemented:
   - AnisotropicTwoBodyDecay.h and AnisotropicTwoBodyDecay.cpp have been added to handle anisotropic decay. 
       - The main function is AnisotropicTwoBodyDist, which samples the integral that makes the decay anisotropic. 
       - The integral sampled is given by: I = (1 -Pol) * I_{1}^{-}/2 + (1 + Pol) * I_{1}^{+)/2. 
       - The function also changes the reference frame to correctly match the original formula in 1905.00284 in which the angle is given in respect to the HNL direction.
   - HNLMakeDecay_tool.cc has changed to include the anisotropic sampling. The boosting of the daughter particles is performed in this part of the code, the AnisotropicTwoBodyDist only gives the distribution in the HNL rest frame.
   - hnl.fcl (the config fhicl for HNL decays) has been changed to include a new boolean: DecayIsTwoBodyAnisotropic. This boolean lets the user choose between the old implementation (done by Gray) and the new one. It is turned true by default.
   
This pull request is linked to sbndcode pull request with link:
https://github.com/SBNSoftware/sbndcode/pull/657


sbncode/EventGenerator/MeVPrtl/Tools/HNL/AnisotropicTwoBodyDecay.cpp
```diff
@@ -0,0 +1,181 @@
...
+  //Get the messon mass, by default choose the meson to be a pion
+  double m_m = evgen::ldm::Constants::Instance().piplus_mass;
+  if (abs(MesonPDG) == 211) {
+    m_m = evgen::ldm::Constants::Instance().piplus_mass;
+  } else if (abs(MesonPDG) == 321) {
+    m_m = evgen::ldm::Constants::Instance().kplus_mass;
+  }
+
+  //Initialize the seed to apply rejection-acceptance method. Use 0 to automatically compute the seed via a TUUID object.
+  TRandom3 *rand = new TRandom3(0);
```
gputnam, line 129: It would be better to use the TRandom object available in the HNLMakeDecay class (fRandom). Could you use that object instead?

sbncode/EventGenerator/MeVPrtl/Tools/HNL/HNLMakeDecay_tool.cc
```diff
@@ -567,53 +569,65 @@ HNLMakeDecay::DecayFinalState HNLMakeDecay::LepPi(const MeVPrtlFlux &flux, bool
...
-  double this_dalitz = 0.;
-  double p = evgen::ldm::twobody_momentum(flux.mass, lep_mass, piplus_mass);
+  }
   TLorentzVector LB;
   TLorentzVector PI;
-  do {
-    TVector3 dir = RandomUnitVector(); 
-    LB = TLorentzVector(p*dir, sqrt(p*p + lep_mass*lep_mass));
-    PI = TLorentzVector(-p*dir, sqrt(p*p + piplus_mass*piplus_mass));
+  if(fDecayIsTwoBodyAnisotropic) {
```
gputnam, line 575: This bool is selecting between including or not the interference in the amplitude between the different polarizations. Your (new) implementation does not include it, while the old one does. Could you change the variable name to be more communicative of that distinction? Maybe something like fTwoBodyAnisotropyIncludeInterference.

sbncode/EventGenerator/MeVPrtl/config/HNL/hnl.fcl
```diff
@@ -46,9 +48,10 @@ decay_hnl: {
 decay_tau_hnl: {
   tool_type: HNLMakeDecay
   Decays: ["nu_mu_mu"]
-  WidthDecays: ["nu_nu_nu", "nu_mu_mu", "nu_e_e", "nu_pi0", "nu_eta", "nu_etap", "nu_rho0", "nu_mu_e"]
+  WidthDecays: ["mu_pi", "e_pi", "nu_mu_mu", "nu_e_e", "nu_nu_nu", "nu_pi0", "nu_eta", "nu_etap", "nu_rho0", "nu_mu_e"]
   Majorana: false
   DecayIsThreeBodyAnisotropic: false
```
gputnam, line 53: Shouldn't we include the three-body anisotropies by default?

gputnam (commented):
Looks good! Made some comments about configuration.

sbncode/EventGenerator/MeVPrtl/Tools/HNL/HNLMakeDecay_tool.cc
LuisPelegrina: Sure! If thats correct I should get the old implementation when fTwoBodyAnisotropyIncludeInterference is true, right?

sbncode/EventGenerator/MeVPrtl/Tools/HNL/AnisotropicTwoBodyDecay.cpp
LuisPelegrina: I'll make the change

jberger7 (approved):

gputnam (approved):

kjplows:
trigger build

PR #531, opened 2025-04-10
This commit introduces the CRTHitT0 tagging tracks in the sbncode IF …

jzennamo:
For @kjplows!


kjplows:
Thanks @jzennamo !! Holding off on approving for now as @francescopoppi is moving this to sbnobj - huge thanks to you both! 🙌

PR #532, opened 2025-04-11
Feature/cerati ng2caf

cerati:
### Description 
This PR adds a few NuGraph-related variables to CAFs, along the lines of what presented on [docdb 40585](https://sbn-docdb.fnal.gov/cgi-bin/sso/ShowDocument?docid=40585). Once icaruscode [PR815](https://github.com/SBNSoftware/icaruscode/pull/815) is merged the needed data products can be produced, but it won't break CAF making if they are not there. Requires `sbnanaobj` [PR137](https://github.com/SBNSoftware/sbnanaobj/pull/137).

### DIFF
--- a/sbncode/CAFMaker/CAFMaker_module.cc
+++ b/sbncode/CAFMaker/CAFMaker_module.cc
@@ -1766,9 +1766,15 @@ void CAFMaker::produce(art::Event& evt) noexcept {
 
     std::vector<art::Ptr<anab::FeatureVector<1>>> ng2_filter_vec;
     std::vector<art::Ptr<anab::FeatureVector<5>>> ng2_semantic_vec;
-    art::fill_ptr_vector(ng2_filter_vec,ng2_filter_handle[producer]);
-    art::fill_ptr_vector(ng2_semantic_vec,ng2_semantic_handle[producer]);
-    FillSliceNuGraph(slcHits,*ng2_slice_hit_map_handle[producer],ng2_filter_vec,ng2_semantic_vec,recslc);
+    if (ng2_filter_handle[producer].isValid()) {
+      art::fill_ptr_vector(ng2_filter_vec,ng2_filter_handle[producer]);
+    }
+    if (ng2_semantic_handle[producer].isValid()) {
+      art::fill_ptr_vector(ng2_semantic_vec,ng2_semantic_handle[producer]);
+    }
+    if (ng2_slice_hit_map_handle[producer].isValid()) {
+      FillSliceNuGraph(slcHits,*ng2_slice_hit_map_handle[producer],ng2_filter_vec,ng2_semantic_vec,recslc);
+    }
 
     art::FindManyP<sbn::OpT0Finder> fmOpT0 =
       FindManyPStrict<sbn::OpT0Finder>(sliceList, evt, fParams.OpT0Label() + slice_tag_suff);
@@ -2142,7 +2148,9 @@ void CAFMaker::produce(art::Event& evt) noexcept {
         FillCNNScores(thisParticle, cnnScores, pfp);
       }
 
-      FillPPFNuGraph(*ng2_slice_hit_map_handle[producer], ng2_filter_vec, ng2_semantic_vec, fmPFPartHits.at(iPart), pfp);
+      if (ng2_slice_hit_map_handle[producer].isValid()) {
+	FillPPFNuGraph(*ng2_slice_hit_map_handle[producer], ng2_filter_vec, ng2_semantic_vec, fmPFPartHits.at(iPart), pfp);
+      }
 
       if (!thisTrack.empty())  { // it has a track!
         assert(thisTrack.size() == 1);
--- a/sbncode/CAFMaker/CAFMaker_module.cc
+++ b/sbncode/CAFMaker/CAFMaker_module.cc
@@ -95,6 +95,7 @@
 #include "lardataobj/RecoBase/Vertex.h"
 #include "lardataobj/RecoBase/Shower.h"
 #include "lardataobj/RecoBase/MCSFitResult.h"
+#include "lardataobj/RecoBase/Cluster.h"
 
 #include "nusimdata/SimulationBase/MCFlux.h"
 #include "nusimdata/SimulationBase/MCTruth.h"
@@ -1696,6 +1697,17 @@ void CAFMaker::produce(art::Event& evt) noexcept {
     }
   }
 
+  // nu graph
+  std::vector< art::Handle<std::vector<unsigned int>> > ng2_slice_hit_map_handle(pandora_tag_suffixes.size());
+  std::vector< art::Handle<std::vector<anab::FeatureVector<1>>> > ng2_filter_handle(pandora_tag_suffixes.size());
+  std::vector< art::Handle<std::vector<anab::FeatureVector<5>>> > ng2_semantic_handle(pandora_tag_suffixes.size());
+  for (unsigned i_tag = 0; i_tag < pandora_tag_suffixes.size(); i_tag++) {
+    const std::string &pandora_tag_suffix = pandora_tag_suffixes[i_tag];
+    GetByLabelIfExists(evt, fParams.NuGraphSliceHitLabel().encode() + pandora_tag_suffix, ng2_slice_hit_map_handle[i_tag]);
+    GetByLabelIfExists(evt, fParams.NuGraphFilterLabel().label() + pandora_tag_suffix + ":" + fParams.NuGraphFilterLabel().instance(), ng2_filter_handle[i_tag]);
+    GetByLabelIfExists(evt, fParams.NuGraphSemanticLabel().label() + pandora_tag_suffix + ":" + fParams.NuGraphSemanticLabel().instance(), ng2_semantic_handle[i_tag]);
+  }
+
   // The Standard Record
   // Branch entry definition -- contains list of slices, CRT information, and truth information
   StandardRecord rec;
@@ -1752,6 +1764,12 @@ void CAFMaker::produce(art::Event& evt) noexcept {
       }
     }
 
+    std::vector<art::Ptr<anab::FeatureVector<1>>> ng2_filter_vec;
+    std::vector<art::Ptr<anab::FeatureVector<5>>> ng2_semantic_vec;
+    art::fill_ptr_vector(ng2_filter_vec,ng2_filter_handle[producer]);
+    art::fill_ptr_vector(ng2_semantic_vec,ng2_semantic_handle[producer]);
+    FillSliceNuGraph(slcHits,*ng2_slice_hit_map_handle[producer],ng2_filter_vec,ng2_semantic_vec,recslc);
+
     art::FindManyP<sbn::OpT0Finder> fmOpT0 =
       FindManyPStrict<sbn::OpT0Finder>(sliceList, evt, fParams.OpT0Label() + slice_tag_suff);
     std::vector<art::Ptr<sbn::OpT0Finder>> slcOpT0;
@@ -1797,6 +1815,25 @@ void CAFMaker::produce(art::Event& evt) noexcept {
 
     art::FindManyP<recob::PFParticle> fmSpacePointPFPs =
       FindManyPStrict<recob::PFParticle>(slcSpacePoints, evt, fParams.PFParticleLabel() + slice_tag_suff);
+
+    art::FindManyP<recob::Cluster> fmPFPClusters =
+      FindManyPStrict<recob::Cluster>(fmPFPart, evt, fParams.PFParticleLabel() + slice_tag_suff);
+
+    std::vector<std::vector<art::Ptr<recob::Hit>>> fmPFPartHits;
+    // make Ptr's to clusters for cluster -> other object associations
+    if (fmPFPClusters.isValid()) {
+      for (size_t ipf=0; ipf<fmPFPart.size();++ipf) {
+	std::vector<art::Ptr<recob::Hit>> pfphits;
+	std::vector<art::Ptr<recob::Cluster>> pfclusters = fmPFPClusters.at(ipf);
+	art::FindManyP<recob::Hit> fmCluHits = FindManyPStrict<recob::Hit>(pfclusters, evt, fParams.PFParticleLabel() + slice_tag_suff);
+	for (size_t icl=0; icl<fmCluHits.size();icl++) {
+	  for (auto hit : fmCluHits.at(icl)) {
+	    pfphits.push_back(hit);
+	  }
+	}
+	fmPFPartHits.push_back(pfphits);
+      }
+    }
 
     art::FindManyP<recob::Shower> fmShower =
       FindManyPStrict<recob::Shower>(fmPFPart, evt, fParams.RecoShowerLabel() + slice_tag_suff);
@@ -2104,6 +2141,8 @@ void CAFMaker::produce(art::Event& evt) noexcept {
         const sbn::PFPCNNScore *cnnScores = fmCNNScores.at(iPart).at(0).get();
         FillCNNScores(thisParticle, cnnScores, pfp);
       }
+
+      FillPPFNuGraph(*ng2_slice_hit_map_handle[producer], ng2_filter_vec, ng2_semantic_vec, fmPFPartHits.at(iPart), pfp);
 
       if (!thisTrack.empty())  { // it has a track!
         assert(thisTrack.size() == 1);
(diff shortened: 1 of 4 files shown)


sbncode/CAFMaker/CAFMaker_module.cc
```diff
@@ -95,6 +95,7 @@
 #include "lardataobj/RecoBase/Vertex.h"
 #include "lardataobj/RecoBase/Shower.h"
 #include "lardataobj/RecoBase/MCSFitResult.h"
+#include "lardataobj/RecoBase/Cluster.h"
```
brucehoward-physics, line 98: I'm only leaving this as comments for now as I have more pieces to look at and so don't want to leave an actual "review" result yet, but I'll put a couple thoughts here.

I guess the inclusion of `FillReco` prevents this from being an actual problem, but I'm trying to recall if our style guidelines suggest that you should actually put
```
#include "lardataobj/AnalysisBase/MVAOutput.h"
```
here as an include too since `FeatureVector` is directly used in the CAFMaker module.

sbncode/CAFMaker/FillReco.cxx
```diff
@@ -544,6 +544,26 @@ namespace caf
...
+			const std::vector<unsigned int> &sliceHitsMap,
+			const std::vector<art::Ptr<anab::FeatureVector<1>>> &ngFilterResult,
+			const std::vector<art::Ptr<anab::FeatureVector<5>>> &ngSemanticResult,
+			caf::SRSlice &slice)
+  {
+
+    //need to double check that the slice processed by NuGraph is the same under consideration
+    //std::cout << "sizes=" << inputHits.size() << " " << sliceHitsMap.size() << " " << ngFilterResult.size() << " " << ngSemanticResult.size() << std::endl;
+    unsigned int nHits = inputHits.size();
+    if (nHits==0 || nHits!=sliceHitsMap.size() || inputHits[0].key()!=sliceHitsMap[0]) return;//not the same slice!
```
brucehoward-physics, line 558: should this be a silent error or would it be a sign of a bigger problem?

sbncode/CAFMaker/FillReco.cxx
```diff
@@ -950,6 +970,59 @@ namespace caf
...
+		      const std::vector<art::Ptr<anab::FeatureVector<1>>> &ngFilterResult,
+		      const std::vector<art::Ptr<anab::FeatureVector<5>>> &ngSemanticResult,
+		      const std::vector<art::Ptr<recob::Hit>> &pfpHits,
+		      caf::SRPFP& srpfp,
+		      bool allowEmpty)
+  {
+
+    auto arg_max = [](std::vector<float> const& vec) {
+      return std::distance(vec.begin(), std::max_element(vec.begin(), vec.end()));
+    };
```
brucehoward-physics, line 983: Another question on the SBN coding conventions: are lambdas okay within other methods? (If @cerati himself isn't sure then maybe @PetrilloAtWork is the best person to answer this (especially as he is also tagged as the other reviewer))

sbncode/CAFMaker/FillReco.cxx
```diff
@@ -950,6 +970,59 @@ namespace caf
...
+      std::vector<float> ng2sempfpcounts(5,0);
+      size_t ng2bkgpfpcount = 0;
+      for (size_t pos : mappedhits) {
+	auto scores = ngSemanticResult.at(pos);
+	std::vector<float> ng2semscores;
+	for (size_t i=0;i<scores->size();i++) ng2semscores.push_back(scores->at(i));
+	size_t sem_label = arg_max(ng2semscores);
+	ng2sempfpcounts[sem_label]++;
+	auto bkgscore = ngFilterResult.at(pos);
+	if (bkgscore->at(0)<0.5) ng2bkgpfpcount++;
```
brucehoward-physics, line 1005: This seems fine, but I was thinking: is there an argument to be made that one should only fill the semantic label fractions for hits that are not tagged as background?

sbncode/CAFMaker/FillReco.cxx
cerati: It's not an error. We just need to skip the slices that were not considered by NuGraph

sbncode/CAFMaker/FillReco.cxx
cerati: Yes, that's a good point. Ideally this should be studied, to find what works best. In general, the semantic labels can be meaningful also when the filter does not pass (e.g. for cosmics). But since noise is a bigger issues than cosmics (unlike MicroBooNE), your suggestion makes sense and I will implement it

sbncode/CAFMaker/CAFMaker_module.cc
cerati: sure, I'll add this include!

sbncode/CAFMaker/FillReco.cxx
brucehoward-physics: Oh right, since NuGraph only runs on one slice!

sbncode/CAFMaker/CAFMaker_module.cc
brucehoward-physics: Thanks for adding!

sbncode/CAFMaker/FillReco.cxx
cerati: yes, the current training is considering them signal but this is going to change soon!

brucehoward-physics: Ah then I guess I'm okay with either way, though I suppose in this updated schema the background fraction would still be meaningful to distinguish a cosmic PFP from others? And the semantic label fractions would help distinguish if there is some overlapping particle what it was, no?

cerati: right. My point is that for ICARUS noise is likely more of an issue than cosmics, so the semantic labels for the background hits are more likely to be meaningless (as they are more likely to be noise and not real cosmic particles). So the new version makes more sense as we don't want to pollute the semantic counts with noise. But of course the background fraction is still useful to reject cosmics!

sbncode/CAFMaker/CAFMaker_module.cc
```diff
@@ -1798,6 +1823,25 @@ void CAFMaker::produce(art::Event& evt) noexcept {
...
+	art::FindManyP<recob::Hit> fmCluHits = FindManyPStrict<recob::Hit>(pfclusters, evt, fParams.PFParticleLabel() + slice_tag_suff);
+	for (size_t icl=0; icl<fmCluHits.size();icl++) {
+	  for (auto hit : fmCluHits.at(icl)) {
+	    pfphits.push_back(hit);
+	  }
+	}
+	fmPFPartHits.push_back(pfphits);
+      }
+    }
+
```
brucehoward-physics, line 1844: I believe we had a quick chat once about using index directly vs using the _art_ Ptr `key()` and that both of us are more used to `key`, but you found that index was working okay (and seems to be used elsewhere here). Please remind me if I'm missing details? I wanted to ask just to double check that this is guaranteed to work as desired!

sbncode/CAFMaker/FillReco.cxx
```diff
@@ -950,6 +970,58 @@ namespace caf
     srpfp.cnnscore.nclusters = cnnscore->nClusters;
   }
 
+  void FillPPFNuGraph(const std::vector<unsigned int> &sliceHitsMap,
```
brucehoward-physics, line 973: I just noticed this is `PPF` not `PFP`

sbncode/CAFMaker/FillReco.cxx
```diff
@@ -950,6 +970,58 @@ namespace caf
...
+
+    // the nugraph elements are ordered the same as the sliceHitsMap
+    std::vector<size_t> mappedhits;
+    for (auto& hit : pfpHits) {
+      auto it = std::find(sliceHitsMap.begin(), sliceHitsMap.end(), hit.key());
+      if (it != sliceHitsMap.end()) {
+	size_t index = std::distance(sliceHitsMap.begin(), it);
+	mappedhits.push_back(index);
+      }
+    }
```
brucehoward-physics, line 989: Just double checking my understanding, I assume that the NuGraph producer is filling up the map that you grab above with the hit keys for the slice and so then this is why you grab things in this way from the map?

sbncode/CAFMaker/CAFMaker_module.cc
cerati: when looping plainly over a collection, the way one would use key is to first create an art::Ptr based on the handle and the index, which guarantees that the key is the same as the index

sbncode/CAFMaker/FillReco.cxx
cerati: Yes. See icaruscode/NuGraphIcarus/IcarusNuSliceHitsProducer_module.cc being added in the icaruscode PR. NuGraph runs over a hit collection. This hit collection is created based on the hits in the best neutrino slice. This means that NuGraph labels are applied to a hit collection which is a subset of the one used by Pandora. So in order to get the NuGraph label of hits in Pandora PFPs, a mapping is needed

sbncode/CAFMaker/FillReco.cxx
```diff
@@ -1004,6 +1024,58 @@ namespace caf
...
+      }
+      srpfp.ng_sem_cat = std::distance(ng2sempfpcounts.begin(), std::max_element(ng2sempfpcounts.begin(), ng2sempfpcounts.end()));//arg_max(ng2sempfpcounts);
+      srpfp.ng_mip_frac = (pfpHits.size()>ng2bkgpfpcount ? float(ng2sempfpcounts[0])/(pfpHits.size()-ng2bkgpfpcount) : -1.);
+      srpfp.ng_hip_frac = (pfpHits.size()>ng2bkgpfpcount ? float(ng2sempfpcounts[1])/(pfpHits.size()-ng2bkgpfpcount) : -1.);
+      srpfp.ng_shr_frac = (pfpHits.size()>ng2bkgpfpcount ? float(ng2sempfpcounts[2])/(pfpHits.size()-ng2bkgpfpcount) : -1.);
+      srpfp.ng_mhl_frac = (pfpHits.size()>ng2bkgpfpcount ? float(ng2sempfpcounts[3])/(pfpHits.size()-ng2bkgpfpcount) : -1.);
+      srpfp.ng_dif_frac = (pfpHits.size()>ng2bkgpfpcount ? float(ng2sempfpcounts[4])/(pfpHits.size()-ng2bkgpfpcount) : -1.);
+      srpfp.ng_bkg_frac = float(ng2bkgpfpcount)/pfpHits.size();
+    } else {
+      srpfp.ng_sem_cat = -1;
```
PetrilloAtWork, line 1068: See the `enum` suggestion on the other PR.

sbncode/CAFMaker/FillReco.cxx
```diff
@@ -1004,6 +1024,58 @@ namespace caf
...
+	size_t index = std::distance(sliceHitsMap.begin(), it);
+	mappedhits.push_back(index);
+      }
+    }
+
+    if (mappedhits.size()>0) {
+      std::vector<float> ng2sempfpcounts(5,0);
+      size_t ng2bkgpfpcount = 0;
+      for (size_t pos : mappedhits) {
+	auto bkgscore = ngFilterResult.at(pos);
```
PetrilloAtWork, line 1049: Avoid unnecessary copies:
```suggestion
	auto const& bkgscore = ngFilterResult.at(pos);
```
(this is probably still fine, since it's implemented as a static array).

sbncode/CAFMaker/FillReco.h
```diff
@@ -105,6 +106,12 @@ namespace caf
                            const std::vector<art::Ptr<recob::SpacePoint>> &inputPoints,
                            caf::SRSlice &slice);
 
+  void FillSliceNuGraph(const std::vector<art::Ptr<recob::Hit>> &inputHits,
+			const std::vector<unsigned int> &sliceHitsMap,
+			const std::vector<art::Ptr<anab::FeatureVector<1>>> &ngFilterResult,
+			const std::vector<art::Ptr<anab::FeatureVector<5>>> &ngSemanticResult,
+			caf::SRSlice &slice);
```
PetrilloAtWork, line 113: Can you doxument this interface, in particular because of the `ngSemanticResult` parameter?
```cpp
  /**
   * @brief Fills the results from semantic segmentation analysis into a `slice`.
   * @param inputHits (pointers to) the hits associated to the slice
   * @param sliceHitsMap maps position of element in semantic segmentation result array -> index/key of its slice hit pertaining that element
   * @param ngFilterResult filter result on each hit (`0.0` didn't pass for sure, up to `1.0` passed for sure)
   * @param ngSemanticResult result of a hit for each category: MIP track, HIP, shower, Michel electron, I don't remember
   * @param[out] slice the destination slice object
   *
   * Hits with filter value (`ngFilterResult`) lower than `0.5` are counted as background.
   * The fraction assigned for each category is the number of non-background hits with score larger than `0.5` for that category.
   */
```

PetrilloAtWork (commented):
No mistakes spotted, asking for some documentation and a bit of code cleanup (in particular, documentation about the hard-coded values and what they do).

brucehoward-physics:
I think I am happy to consider this approved, based on Giuseppe's responses and changes e.g., once Gianluca is satisfied. I don't know if me hitting "approve" and being otherwise silent will somehow supersede Gianluca's comments in the eyes of @kjplows or other manager, so remarking this here :)

brucehoward-physics (approved):
Approved with the comment given about pending Gianluca's satisfaction as well.

kjplows:
trigger build SBNSoftware/sbnanaobj#137

sbncode/CAFMaker/FillReco.cxx
```diff
@@ -1004,6 +1024,58 @@ namespace caf
...
+	  size_t sem_label = std::distance(ng2semscores.begin(), std::max_element(ng2semscores.begin(), ng2semscores.end()));//arg_max(ng2semscores);
+	  ng2sempfpcounts[sem_label]++;
+	}
+      }
+      srpfp.ngscore.sem_cat = SRNuGraphScore::NuGraphCategory(std::distance(ng2sempfpcounts.begin(), std::max_element(ng2sempfpcounts.begin(), ng2sempfpcounts.end())));//arg_max(ng2sempfpcounts);
+      srpfp.ngscore.mip_frac = (pfpHits.size()>ng2bkgpfpcount ? float(ng2sempfpcounts[0])/(pfpHits.size()-ng2bkgpfpcount) : -1.);
+      srpfp.ngscore.hip_frac = (pfpHits.size()>ng2bkgpfpcount ? float(ng2sempfpcounts[1])/(pfpHits.size()-ng2bkgpfpcount) : -1.);
+      srpfp.ngscore.shr_frac = (pfpHits.size()>ng2bkgpfpcount ? float(ng2sempfpcounts[2])/(pfpHits.size()-ng2bkgpfpcount) : -1.);
+      srpfp.ngscore.mhl_frac = (pfpHits.size()>ng2bkgpfpcount ? float(ng2sempfpcounts[3])/(pfpHits.size()-ng2bkgpfpcount) : -1.);
+      srpfp.ngscore.dif_frac = (pfpHits.size()>ng2bkgpfpcount ? float(ng2sempfpcounts[4])/(pfpHits.size()-ng2bkgpfpcount) : -1.);
```
PetrilloAtWork, line 1065: More maintenance- and reader-friendly:
```suggestion
      size_t hitExcess = pfpHits.size()>ng2bkgpfpcount ? pfpHits.size()-ng2bkgpfpcount: 0;
      srpfp.ngscore.mip_frac = hitExcess > 0 ? float(ng2sempfpcounts[0])/hitExcess : -1.;
      srpfp.ngscore.hip_frac = hitExcess > 0 ? float(ng2sempfpcounts[1])/hitExcess : -1.;
      srpfp.ngscore.shr_frac = hitExcess > 0 ? float(ng2sempfpcounts[2])/hitExcess : -1.;
      srpfp.ngscore.mhl_frac = hitExcess > 0 ? float(ng2sempfpcounts[3])/hitExcess : -1.;
      srpfp.ngscore.dif_frac = hitExcess > 0 ? float(ng2sempfpcounts[4])/hitExcess : -1.;
```

PetrilloAtWork (approved):

kjplows:
trigger build SBNSoftware/sbnanaobj#137

sbncode/CAFMaker/FillReco.cxx
PetrilloAtWork: There is nothing about this case in our guidelines.
Personally I do recommend the use of lambdas in methods where it makes sense (typically, to avoid the repetition of simple code). However, if the same lambda appears in different methods, then it should likely be a member function if possible.

PR #536, opened 2025-04-24
This PR introduces CRT Tagging usage into cafs for analyzers

francescopoppi:
This Pull request introduces the CRTHit tagging into CAF, so that ICARUS analyzers can use relevant information for out of time rejection.
This PR uses an existing SR and Fill function which have been modified accordingly.
This PR needs a new dependence on this PR for sbnanaobj:
https://github.com/SBNSoftware/sbnanaobj/pull/139

### DIFF
--- a/sbncode/CAFMaker/CAFMakerParams.h
+++ b/sbncode/CAFMaker/CAFMakerParams.h
@@ -275,7 +275,13 @@ namespace caf
     Atom<string> CRTHitMatchLabel {
       Name("CRTHitMatchLabel"),
       Comment("Base label of track to CRT hit matching producer."),
-      "pandoraTrackCRTHit"
+      "CRTT0Tagging"
+    };
+
+    Atom<string> CRTHitMatchInfoLabel {
+      Name("CRTHitMatchInfoLabel"),
+      Comment("Base label of additional information on track to CRT hit matching producer."),
+      "CRTT0Tagging"
     };
 
     Atom<string> CRTTrackMatchLabel {
--- a/sbncode/CAFMaker/CAFMakerParams.h
+++ b/sbncode/CAFMaker/CAFMakerParams.h
@@ -275,7 +275,7 @@ namespace caf
     Atom<string> CRTHitMatchLabel {
       Name("CRTHitMatchLabel"),
       Comment("Base label of track to CRT hit matching producer."),
-      "CRTT0Tagging"
+      "pandoraTrackCRTHit"
     };
 
     Atom<string> CRTHitMatchInfoLabel {
--- a/sbncode/CAFMaker/CAFMaker_module.cc
+++ b/sbncode/CAFMaker/CAFMaker_module.cc
@@ -1926,9 +1926,21 @@ void CAFMaker::produce(art::Event& evt) noexcept {
 
     // NOTE: The sbn::crt::CRTHit is associated to the T0. It's a bit awkward to
     // access that here, so we do it per-track (see code where fmCRTHitMatch is accessed below)
+   
+    // note: in the ICARUS case we do not need a slice_tag_stuff for CRTHitMatch as it is
+    // common to East and West
+    /*
     art::FindManyP<anab::T0> fmCRTHitMatch =
       FindManyPStrict<anab::T0>(slcTracks, evt,
                fParams.CRTHitMatchLabel() + slice_tag_suff);
+    */
+    art::FindManyP<anab::T0> fmCRTHitMatch =
+      FindManyPStrict<anab::T0>(slcTracks, evt,
+               fParams.CRTHitMatchLabel());
+
+    art::FindManyP<sbn::crt::CRTHitT0TaggingInfo> fmCRTHitMatchInfo =
+      FindManyPStrict<sbn::crt::CRTHitT0TaggingInfo>(slcTracks, evt,
+               fParams.CRTHitMatchInfoLabel());
 
     // TODO: also save the sbn::crt::CRTTrack in the matching so that CAFMaker has access to it
     art::FindManyP<anab::T0> fmCRTTrackMatch =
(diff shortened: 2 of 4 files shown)


sbncode/CAFMaker/FillReco.cxx
```diff
@@ -557,7 +558,16 @@ namespace caf
     if (t0match.size()) {
       assert(t0match.size() == 1);
       srtrack.crthit.distance = t0match[0]->fTriggerConfidence;
+      srtrack.crthit.region = t0match[0]->fID;
+      srtrack.crthit.sys = t0match[0]->fTriggerBits;      
+      srtrack.crthit.deltaX = hitmatchinfo[0]->DeltaX;
```
gputnam, line 563: This would crash len(hitmatchinfo) == 0 and len(t0match) == 1, as I believe would be the case in SBND.

Could you change this to include a separate if statement of: if (hitmatchinfo.size())

gputnam (changes requested):
Approved, modulo this change.

francescopoppi:
Request from @gputnam addressed

sbncode/CAFMaker/CAFMaker_module.cc
```diff
@@ -1926,9 +1926,21 @@ void CAFMaker::produce(art::Event& evt) noexcept {
 
     // NOTE: The sbn::crt::CRTHit is associated to the T0. It's a bit awkward to
     // access that here, so we do it per-track (see code where fmCRTHitMatch is accessed below)
+   
+    // note: in the ICARUS case we do not need a slice_tag_stuff for CRTHitMatch as it is
+    // common to East and West
+    /*
     art::FindManyP<anab::T0> fmCRTHitMatch =
       FindManyPStrict<anab::T0>(slcTracks, evt,
                fParams.CRTHitMatchLabel() + slice_tag_suff);
```
kjplows, line 1935: I'm hesitant to just get rid off the `slice_tag_suff`, as I suspect SBND may need it. Summoning @henrylay97 to comment on this

PetrilloAtWork: Could anybody explain how SBND stores the tracks? ICARUS splits them in two data products by their cryostat, which I assume SBND does not.

sbncode/CAFMaker/CAFMaker_module.cc
```diff
@@ -2167,14 +2179,20 @@ void CAFMaker::produce(art::Event& evt) noexcept {
...
+              (fmCRTHitMatch.at(iPart), evt, fParams.CRTHitMatchLabel());
 
           std::vector<art::Ptr<sbn::crt::CRTHit>> crthitmatch;
           if (CRTT02Hit.isValid() && CRTT02Hit.size() == 1) crthitmatch = CRTT02Hit.at(0);
 
-          FillTrackCRTHit(fmCRTHitMatch.at(iPart), crthitmatch, fParams.CRTUseTS0(), CRT_T0_reference_time, CRT_T1_reference_time, trk);
+          std::vector<art::Ptr<sbn::crt::CRTHitT0TaggingInfo>> crthittagginginfo;
+          if(CRTT02Hit.isValid() && CRTT02Hit.size() == 1){
+            crthittagginginfo = fmCRTHitMatchInfo.at(iPart);
+          }
```
PetrilloAtWork, line 2193: It looks like these two if branches are related and can be merged:
```cpp
          std::vector<art::Ptr<sbn::crt::CRTHit>> crthitmatch;
          std::vector<art::Ptr<sbn::crt::CRTHitT0TaggingInfo>> crthittagginginfo;
          if(CRTT02Hit.isValid() && CRTT02Hit.size() == 1){
            crthitmatch = CRTT02Hit.at(0);
            crthittagginginfo = fmCRTHitMatchInfo.at(iPart);
          }
```

sbncode/CAFMaker/CAFMaker_module.cc
PetrilloAtWork, line 1: Add
```cpp
#include "sbnobj/Common/CRT/CRTHitT0TaggingInfo.hh"
```

sbncode/CAFMaker/FillReco.cxx
```diff
@@ -548,6 +548,7 @@ namespace caf
 
   void FillTrackCRTHit(const std::vector<art::Ptr<anab::T0>> &t0match,
                        const std::vector<art::Ptr<sbn::crt::CRTHit>> &hitmatch,
+                       const std::vector<art::Ptr<sbn::crt::CRTHitT0TaggingInfo>> &hitmatchinfo,
```
PetrilloAtWork, line 551: The implementation strictly requires one `hitmatchinfo` element.
If this is necessary, then the argument should be a pointer rather than a vector of pointers. If instead this is left more general for future extensions, I request a comment on why if there is more than one hit no information is saved, and that reason should be documented in the code too.
At the moment it seems that the output data record accommodates only one matched hit though.

PetrilloAtWork (changes requested):
It looks like there are three points to address:
1. Whether SBND needs the suffix on CRT data product tag.
2. The strategy for identifying the location of a hit.
3. The support for multiple CRT hits associated to the same track.

Possibly some off-GitHub discussion is needed too.

sbncode/CAFMaker/CAFMakerParams.h
```diff
@@ -275,7 +275,13 @@ namespace caf
     Atom<string> CRTHitMatchLabel {
       Name("CRTHitMatchLabel"),
       Comment("Base label of track to CRT hit matching producer."),
-      "pandoraTrackCRTHit"
+      "CRTT0Tagging"
```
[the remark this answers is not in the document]
henrylay97: So I agree with Gianluca on the general principle.

Although for this scenario it is okay for SBND. We use the CRTSpacePoint matching rather than CRTHit we also have different methods for the track matching products so I think we are totally factorised from each other here.

sbncode/CAFMaker/CAFMaker_module.cc
henrylay97: At SBND we store our tracks as a single collection - due to the fact we have just one cryostat. So we don't use the suffix approach.

henrylay97 (commented):
In general this looks fine from an SBND point of view. Our CRT reconstruction uses different objects to ICARUS and therefore the CAF filling is completely factorised for the two experiments. Nothing Francesco does here will impact us. I would like to see that verified in the CI tests once this PR is ready to be tested.

I will leave others (as Gianluca already has) to comment on the implementation details.

sbncode/CAFMaker/CAFMakerParams.h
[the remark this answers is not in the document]
francescopoppi: silent change reverted.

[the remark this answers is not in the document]
francescopoppi: but I will also have to change the carmaker icarus definitions (which I should do anyway)

francescopoppi:
Addressed the requests from @PetrilloAtWork, but I would like clarification on the point:
The strategy for identifying the location of a hit.
What do you mean exactly?

PetrilloAtWork (approved):
Thank you for the clarifications.
I would just remove the code commented out, since it is not good for SBND and it is not good for ICARUS, and it never was.

kjplows:
trigger build

PetrilloAtWork (approved):
Approved!

kjplows:
trigger build LArSoft/lar*@LARSOFT_SUITE_v10_05_00 SBNSoftware/sbnanaobj#139

kjplows:
trigger build LArSoft/lar*@LARSOFT_SUITE_v10_05_00 https://github.com/SBNSoftware/sbnanaobj/pull/139

kjplows:
checked merge diff manually and no impact is made on the CRT Tagging info. Weird how the CI didn't seem to pick up sbnanaobj/#139 . 

Built on sbndbuild03 with both branches so given this passed CI before merge commit this is approved

PR #557, opened 2025-08-21
Calib NTuple CRT Timing Additions

henrylay97:
### Description 
Part of a pair of PRs to enhance the SBND CRT contribution to the calibration ntuples. (See also SBNSoftware/sbnobj#139).

This PR:
- Adds a new T0 source (SBND-only) from CRTSpacePoint matching. This falls in lowest priority (after Pandora (Both), CRTTrack (SBND-only) and CRTHit (ICARUS-only) and will be denoted by `whicht0==3`.
- Applies appropriate x shifts to TPC track end points for this new T0.
- Adds in x shifting of hit / spacepoint using either of the SBND-only CRT T0 tags - previously only implemented for CRTHit.
- Skips shower-like PFPs now that all PFPs are characterised as tracks, this returns to original behaviour (before ~2years ago) and removes a sea of deltas and michels from the calib ntuples.

Presenting today at the calibration meeting - will link slides after.

- [X] Is this PR related to an open issue / project?
- [ ] Does this PR affect CAF data format? If so, please assign a CAF maintainer as additional reviewer.
- [ ] Are you submitting this PR on behalf of someone else who made the code changes? If so, please mention them in the description.


kjplows:
trigger build ci_ref=v10_06_02 LArSoft/lar*@LARSOFT_SUITE_v10_09_00 SBNSoftware/sbnobj#139

kjplows:
@mrmooney @gputnam @francescopoppi just pinging here - from a CI perspective this and #564 (production branch version, on the larsoft v10_06 line) are good to go. Any chance you could look at this? Thanks!

sbncode/Calibration/TrackCaloSkimmer_module.cc
```diff
@@ -313,6 +315,9 @@ void sbn::TrackCaloSkimmer::analyze(art::Event const& e)
   for (art::Ptr<recob::PFParticle> p_pfp: PFParticleList) {
     const recob::PFParticle &pfp = *p_pfp;
 
+    if(p_pfp->PdgCode() == 11)
```
gputnam, line 318: Why is this being added in? I don't like having selection code inside the module itself. Selection code should go into one of the selection tools.

sbncode/Calibration/TrackCaloSkimmer_module.cc
```diff
@@ -1400,13 +1457,29 @@ sbn::TrackHitInfo sbn::TrackCaloSkimmer::MakeHit(const recob::Hit &hit,
...
+        time = t0Info.t0CRTTrack*1e-3;
+      else if(t0Info.hasT0CRTHit)
+        time = t0Info.t0CRTHit*1e-3;
+      else if(t0Info.hasT0CRTSpacePoint)
+        time = t0Info.t0CRTSpacePoint*1e-3;
+
+      double anodeDistance = (hit.PeakTime()-dclock.Time2Tick(dclock.TriggerTime())-time/dclock.TPCClock().TickPeriod())*dclock.TPCClock().TickPeriod()*driftv;
+      double wirePlaneX = wireReadout->Plane(geo::PlaneID(hit.WireID().Cryostat, hit.WireID().TPC, hit.WireID().Plane)).GetCenter().X();
+
+      recoX = wirePlaneX - driftDir*anodeDistance;
```
gputnam, line 1480: Since xshift is now being passed as a parameter, shouldn't you just apply it instead of calculating recoX here?

Backing up -- tp.x and sp.x should be shifted by the same amount. They shouldn't have different calculations.

sbncode/Calibration/TrackCaloSkimmer_module.cc
gputnam: That's fine then, this can stay

sbncode/Calibration/TrackCaloSkimmer_module.cc
henrylay97: @gputnam @francescopoppi I am seeing ~0mm, ~3mm and ~6mm differences between this technique and pure use of the xshift. Clearly one of the methods doesn't / double accounts for interplane drift time.

I will try and investigate this this afternoon but given you implemented it I would welcome your input - do you know for example, whether this is over corrected?

henrylay97: But clearly one of these instances is wrong so it's worth understanding which.

I would expect that the interplane drift is already accounted for in the creation of the spacepoints given they're already 3D constructs. I'm not familiar with the trajectory points so I would need to understand how they work.

henrylay97: I have looked into this more - SPs explicitly account for the interplane drift as would be expected, TPs are based on the trajectory which itself is a fit to the SPs.

Correcting for interplane drift here is wrong. I've made a couple of plots that show that the SP x's give and agreed x range (with the first induction location being the limit) whilst the TP x's with their current correction give shifted distributions.
hits_all_tpc_0_sp_x
hits_all_tpc_0_tp_x

henrylay97: I am therefore going to remove this section and correct all locations by the xshift calculated previously, avoiding double correcting.

henrylay97: Commit now pushed.

@mrmooney @linyan-w just making you aware of this too!

kjplows:
trigger build LArSoft/lar*@LARSOFT_SUITE_v10_10_02 SBNSoftware/sbnobj#139

mrmooney (approved):

PR #586, opened 2025-09-30
 Add truth matching info to CRT hits

gputnam:
Adds in G4 matching for ICARUS-style CRT hits in CAFMaker.

Copy of https://github.com/SBNSoftware/sbncode/pull/585, for develop.


francescopoppi (approved):
All good!

PR #598, opened 2025-10-30
Feature/pot q fix

nathanielerowe:
Query was very slow after latest updates to POT modules for FOM calculation. This tests a potential fix.


nathanielerowe:
trigger build

PR #616, opened 2025-12-05
Handle CAF making when Pandora runs after the NuGraph2 filter

rtriozzi:
This PR adds a mode governed by the `UsePandoraAfterNuGraph` FHiCL flag to handle NuGraph2 variables in the CAF maker, when using a second-pass Pandora after filtering noise with NuGraph2, based on work by summer intern Leonardo Lena. 

This PR also provides code to fill new NuGraph2 slice-level variables in CAF files.

Relevant information about the NuGraph2 chain is provided in [icaruscode PR #868](https://github.com/SBNSoftware/icaruscode/pull/868). 
___
### Associated PRs
- https://github.com/SBNSoftware/sbnanaobj/pull/183
- https://github.com/SBNSoftware/icaruscode/pull/868

___
### Review
Tagging for review @acampani and @PetrilloAtWork as reconstruction and infrastructure experts. Thanks a lot!
___
### Description 
Please provide a detailed description of the changes this pull request introduces. If available, also link to a docdb link where the issue/change have been presented on/discussed.

- [x] Have you added a label? (bug/enhancement/physics etc.)
- [x] Have you assigned at least 1 reviewer?
- [] Is this PR related to an open issue / project?
- [x] Does this PR affect CAF data format? If so, please assign a CAF maintainer as additional reviewer.
- [x] Does this PR require merging another PR in a different repository (such as sbnanobj/sbnobj etc.)? If so, please link it in the description.
- [x] Are you submitting this PR on behalf of someone else who made the code changes? If so, please mention them in the description.

### DIFF
--- a/sbncode/CAFMaker/CAFMakerParams.h
+++ b/sbncode/CAFMaker/CAFMakerParams.h
@@ -350,10 +350,10 @@ namespace caf
       "" //Empty by default, configured in icaruscode cafmaker_defs
     };
 
-    Atom<art::InputTag> NuGraphSliceHitLabel {
-      Name("NuGraphSliceHitLabel"),
-      Comment("Label of NuGraph slice hit map."),
-      "" //Empty by default, please set to e.g. art::InputTag("nuslhits")
+    Atom<art::InputTag> NCCSlicesLabel {
+      Name("NCCSlicesLabel"),
+      Comment("Label of NotClearCosmic slices."),
+      "" //Empty by default, please set to e.g. art::InputTag("NCCSlices")
     };
 
     Atom<art::InputTag> NuGraphFilterLabel {
--- a/sbncode/CAFMaker/CAFMakerParams.h
+++ b/sbncode/CAFMaker/CAFMakerParams.h
@@ -374,6 +374,12 @@ namespace caf
       false
     };
     
+    Atom<float> NuGraphFilterCut {
+      Name("NuGraphFilterCut"),
+      Comment("Cut on the NuGraph2 filter score to define hit as signal or noise."),
+      0.5
+    };
+
     Atom<float> NuGraphHIPTagWireDist {
       Name("NuGraphHIPTagWireDist"),
       Comment("TPC wire distance from the vertex used to count NuGraph2–tagged HIP hits."),
--- a/sbncode/CAFMaker/CAFMakerParams.h
+++ b/sbncode/CAFMaker/CAFMakerParams.h
@@ -368,6 +368,12 @@ namespace caf
       "" //Empty by default, please set to e.g. art::InputTag("NuGraph","semantic")
     };
 
+    Atom<bool> UsePandoraAfterNuGraph {
+      Name("UsePandoraAfterNuGraph"),
+      Comment("Whether to use the second pass Pandora outputs for NuGraph reco."),
+      false
+    };
+
     Atom<string> OpFlashLabel {
       Name("OpFlashLabel"),
       Comment("Label of PMT flash."),
--- a/sbncode/CAFMaker/CAFMakerParams.h
+++ b/sbncode/CAFMaker/CAFMakerParams.h
@@ -373,6 +373,18 @@ namespace caf
       Comment("Whether to use the second pass Pandora outputs for NuGraph reco."),
       false
     };
+    
+    Atom<float> NuGraphHIPTagWireDist {
+      Name("NuGraphHIPTagWireDist"),
+      Comment("TPC wire distance from the vertex used to count NuGraph2–tagged HIP hits."),
+      10
+    };
+
+    Atom<float> NuGraphHIPTagTickDist {
+      Name("NuGraphHIPTagTickDist"),
+      Comment("TPC tick distance from the vertex used to count NuGraph-2–tagged HIP hits."),
+      50
+    };
 
     Atom<string> OpFlashLabel {
       Name("OpFlashLabel"),
--- a/sbncode/CAFMaker/CAFMakerParams.h
+++ b/sbncode/CAFMaker/CAFMakerParams.h
@@ -350,9 +350,9 @@ namespace caf
       "" //Empty by default, configured in icaruscode cafmaker_defs
     };
 
-    Atom<art::InputTag> NCCSlicesLabel {
-      Name("NCCSlicesLabel"),
-      Comment("Label of NotClearCosmic slices."),
+    Atom<art::InputTag> NuGraphSlicesLabel {
+      Name("NuGraphSlicesLabel"),
+      Comment("Label of slices that have NuGraph inference."),
       "" //Empty by default, please set to e.g. art::InputTag("NCCSlices")
     };
 
--- a/sbncode/CAFMaker/CAFMaker_module.cc
+++ b/sbncode/CAFMaker/CAFMaker_module.cc
@@ -1797,39 +1797,6 @@ void CAFMaker::produce(art::Event& evt) noexcept {
       }
     }
     
-    // if (std::find(nuGraphSlices.begin(), nuGraphSlices.end(), slice) != nuGraphSlices.end()) {
-    //   std::vector<art::Ptr<anab::FeatureVector<1>>> ng2_filter_vec;
-    //   std::vector<art::Ptr<anab::FeatureVector<5>>> ng2_semantic_vec;
-    //   art::FindOneP<anab::FeatureVector<1>> findOneFilter(slcHits, evt, fParams.NuGraphFilterLabel().label() + slice_tag_suff + ":" + fParams.NuGraphFilterLabel().instance());
-    //   art::FindOneP<anab::FeatureVector<5>> findOneSemantic(slcHits, evt, fParams.NuGraphSemanticLabel().label() + slice_tag_suff + ":" + fParams.NuGraphSemanticLabel().instance());
-
-    //   if (findOneFilter.isValid()) {
-    //     for (size_t hitIdx = 0; hitIdx < slcHits.size(); ++hitIdx) {
-    //       if (findOneFilter.at(hitIdx).isNull()) {
-    //         slcHits.erase(slcHits.begin()+hitIdx);
-    //         hitIdx--;
-    //         continue; 
-    //       }
-    //       ng2_filter_vec.emplace_back(findOneFilter.at(hitIdx));
-    //     }
-    //   }
-
-    //   if (findOneSemantic.isValid()) {
-    //     for (size_t hitIdx = 0; hitIdx < slcHits.size(); ++hitIdx) {
-    //       if (findOneSemantic.at(hitIdx).isNull()) {
-    //         slcHits.erase(slcHits.begin()+hitIdx);
-    //         hitIdx--;
-    //         continue;
-    //       }
-    //       ng2_semantic_vec.emplace_back(findOneSemantic.at(hitIdx));
-    //     }
-    //   }
-
-    //   if (ng2_filter_vec.size() > 0 || ng2_semantic_vec.size() > 0) {
-    //     FillSliceNuGraph(slcHits, ng2_filter_vec, ng2_semantic_vec, recslc);
-    //   }
-    // }
-
     art::FindManyP<sbn::OpT0Finder> fmOpT0 =
       FindManyPStrict<sbn::OpT0Finder>(sliceList, evt, fParams.OpT0Label() + slice_tag_suff);
     std::vector<art::Ptr<sbn::OpT0Finder>> slcOpT0;
--- a/sbncode/CAFMaker/CAFMaker_module.cc
+++ b/sbncode/CAFMaker/CAFMaker_module.cc
@@ -1717,7 +1717,7 @@ void CAFMaker::produce(art::Event& evt) noexcept {
 
   // collect the TPC slices
   std::vector<art::Ptr<recob::Slice>> slices;
-  std::vector<art::Ptr<recob::Slice>> nccSlices;
+  std::vector<art::Ptr<recob::Slice>> nuGraphSlices;
   std::vector<std::string> slice_tag_suffixes;
   std::vector<unsigned> slice_tag_indices;
   for (unsigned i_tag = 0; i_tag < pandora_tag_suffixes.size(); i_tag++) {
@@ -1727,8 +1727,8 @@ void CAFMaker::produce(art::Event& evt) noexcept {
     GetByLabelStrict(evt, fParams.PFParticleLabel() + pandora_tag_suffix, thisSlices);
     if (thisSlices.isValid()) {
       art::fill_ptr_vector(slices, thisSlices);
-      const std::vector<art::Ptr<recob::Slice>>& tempNCCSlices = evt.getProduct<std::vector<art::Ptr<recob::Slice>>>(fParams.NCCSlicesLabel().label() + pandora_tag_suffix);
-      nccSlices.insert(nccSlices.end(), tempNCCSlices.begin(), tempNCCSlices.end());
+      const std::vector<art::Ptr<recob::Slice>>& tempNuGraphSlices = evt.getProduct<std::vector<art::Ptr<recob::Slice>>>(fParams.NuGraphSlicesLabel().label() + pandora_tag_suffix);
+      nuGraphSlices.insert(nuGraphSlices.end(), tempNuGraphSlices.begin(), tempNuGraphSlices.end());
       for (unsigned i = 0; i < thisSlices->size(); i++) {
         slice_tag_suffixes.push_back(pandora_tag_suffix);
         slice_tag_indices.push_back(i_tag);
@@ -1792,9 +1792,7 @@ void CAFMaker::produce(art::Event& evt) noexcept {
       }
     }
     
-    if (std::find(nccSlices.begin(), nccSlices.end(), slice) != nccSlices.end()) {
-      // this check is required: if FindOneP does not find an associated item instead of returning nullptr it throws an exception.
-      // I believe a change of this behaviour could be handy.
+    if (std::find(nuGraphSlices.begin(), nuGraphSlices.end(), slice) != nuGraphSlices.end()) {
       std::vector<art::Ptr<anab::FeatureVector<1>>> ng2_filter_vec;
       std::vector<art::Ptr<anab::FeatureVector<5>>> ng2_semantic_vec;
       art::FindOneP<anab::FeatureVector<1>> findOneFilter(slcHits, evt, fParams.NuGraphFilterLabel().label() + slice_tag_suff + ":" + fParams.NuGraphFilterLabel().instance());
@@ -1802,12 +1800,22 @@ void CAFMaker::produce(art::Event& evt) noexcept {
       // iteration is the only way to get something out of a FindOne or FindMany...
       if (findOneFilter.isValid()) {
         for (size_t hitIdx = 0; hitIdx < slcHits.size(); ++hitIdx) {
+          if (findOneFilter.at(hitIdx).isNull()) {
+            slcHits.erase(slcHits.begin()+hitIdx);
+            hitIdx--;
+            continue; 
+          }
           ng2_filter_vec.emplace_back(findOneFilter.at(hitIdx));
         }
       }
 
       if (findOneSemantic.isValid()) {
         for (size_t hitIdx = 0; hitIdx < slcHits.size(); ++hitIdx) {
+          if (findOneSemantic.at(hitIdx).isNull()) {
+            slcHits.erase(slcHits.begin()+hitIdx);
+            hitIdx--;
+            continue;
+          }
           ng2_semantic_vec.emplace_back(findOneSemantic.at(hitIdx));
         }
       }
@@ -2194,8 +2202,8 @@ void CAFMaker::produce(art::Event& evt) noexcept {
         FillCNNScores(thisParticle, cnnScores, pfp);
       }
 
-      if (std::find(nccSlices.begin(), nccSlices.end(), slice) != nccSlices.end()) {
-        const std::vector<art::Ptr<recob::Hit>>& PFPHits = fmPFPartHits.at(iPart);
+      if (std::find(nuGraphSlices.begin(), nuGraphSlices.end(), slice) != nuGraphSlices.end()) {
+        std::vector<art::Ptr<recob::Hit>>& PFPHits = fmPFPartHits.at(iPart);
         art::FindOneP<anab::FeatureVector<1>> findOneFilter(PFPHits, evt, fParams.NuGraphFilterLabel().label() + slice_tag_suff + ":" + fParams.NuGraphFilterLabel().instance());
         art::FindOneP<anab::FeatureVector<5>> findOneSemantic(PFPHits, evt, fParams.NuGraphSemanticLabel().label() + slice_tag_suff + ":" + fParams.NuGraphSemanticLabel().instance());
         std::vector<art::Ptr<anab::FeatureVector<1>>> ng2_filter_vec;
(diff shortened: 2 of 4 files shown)


cerati (changes requested):
Thanks, overall it looks great -- especially removing the need for the hit mapping makes things much simpler.

The two things I would suggest to change are:
1. Avoid erasing elements from vectors. I think it should be much easier to simply add a check that the element is non-null in the Fill*NuGraph method. Or is there another reason for erasing the elements with null nugraph score?
2. There is no need to get the NuGraph slices from the event if then we just use `slices` (when `UsePandoraAfterNuGraph=true`)

sbncode/CAFMaker/FillReco.cxx
```diff
@@ -581,22 +579,80 @@ namespace caf
...
+        );
+
+        // HIP tagging
+        float dwire = std::abs(float(hit.WireID().Wire) - vtx_wire[plane]);
+        float dtick = std::abs(hit.PeakTime() - vtx_tick[plane]);
+        if ((highestScoreIdx == 1) && (dwire <= vtx_wire_dist) && (dtick <= vtx_tick_dist))
+          nHIPHits += 1;
+
+        // shower hits
+        if (highestScoreIdx == 2) {
```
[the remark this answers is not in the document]
rtriozzi: Right, thanks! Fixed (and sneaked in some more variables)!

cerati (approved):
Thanks for addressing my comments!

sbncode/CAFMaker/CAFMaker_module.cc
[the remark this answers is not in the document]
acampani: Thanks for the clarification.

sbncode/CAFMaker/CAFMakerParams.h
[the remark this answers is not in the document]
acampani: Thanks for the details, this is very interesting.

sparshitadey:
Hi! 
Small addition to fix `ng_filt_pass_frac` (currently always 1.0 in the NuGraph workflow):
In the workflow, `slcHits` comes from hits that already passed the filter by construction, so `ng_filt_pass_frac` computed from `ngFilterResult` is trivially 1.0. 

The fix is to compute it directly from collection sizes - see commit [363b5120](https://github.com/sparshitadey/sbncode/commit/363b5120) on my branch. Gives mean ~0.81, varying event to event, consistent with Stage 1 filter retention (and the printouts). Happy to open a separate PR if preferred!

sbncode/CAFMaker/CAFMaker_module.cc
```diff
@@ -1945,33 +1717,30 @@ void CAFMaker::produce(art::Event& evt) noexcept {
...
     art::Handle<std::vector<recob::Slice>> thisSlices;
     GetByLabelStrict(evt, fParams.PFParticleLabel() + pandora_tag_suffix, thisSlices);
+
     if (thisSlices.isValid()) {
       art::fill_ptr_vector(slices, thisSlices);
+      if (fParams.UsePandoraAfterNuGraph()) {
+        nuGraphSlices = slices;
+      } else {
+        nuGraphSlices = evt.getProduct<std::vector<art::Ptr<recob::Slice>>>(fParams.NuGraphSlicesLabel().label() + pandora_tag_suffix);
+      }
```
PetrilloAtWork, line 1736: I don't understand this logic.
If `fParams.UsePandoraAfterNuGraph()` is `true`, `nugraphSlices` will be the same as `slices` (so that case could easily live outside the loop); in the other case, `nuGraphSlices` is overwritten with the content of each tag until the last one, hence holding only slices from the data product with the last tag.
I suspect what was meant to _add_ slices, something like
```cpp
      if (!fParams.UsePandoraAfterNuGraph()) {
        auto const& moreSlices = evt.getProduct<std::vector<art::Ptr<recob::Slice>>>(fParams.NuGraphSlicesLabel().label() + pandora_tag_suffix);
        nuGraphSlices.insert(nuGraphSlices.end(), moreSlices.begin(), moreSlices.end());
      }
```

sbncode/CAFMaker/CAFMaker_module.cc
```diff
@@ -2102,15 +1850,15 @@ void CAFMaker::produce(art::Event& evt) noexcept {
...
-	std::vector<art::Ptr<recob::Cluster>> pfclusters = fmPFPClusters.at(ipf);
-	art::FindManyP<recob::Hit> fmCluHits = FindManyPStrict<recob::Hit>(pfclusters, evt, fParams.PFParticleLabel() + slice_tag_suff);
-	for (size_t icl=0; icl<fmCluHits.size();icl++) {
-	  for (auto hit : fmCluHits.at(icl)) {
-	    pfphits.push_back(hit);
-	  }
-	}
-	fmPFPartHits.push_back(pfphits);
+        std::vector<art::Ptr<recob::Hit>> pfphits;
+        std::vector<art::Ptr<recob::Cluster>> pfclusters = fmPFPClusters.at(ipf);
```
PetrilloAtWork, line 1854: Avoid copies:
```suggestion
        std::vector<art::Ptr<recob::Cluster>> const& pfclusters = fmPFPClusters.at(ipf);
```

sbncode/CAFMaker/CAFMaker_module.cc
```diff
@@ -2102,15 +1850,15 @@ void CAFMaker::produce(art::Event& evt) noexcept {
...
+        art::FindManyP<recob::Hit> fmCluHits = FindManyPStrict<recob::Hit>(pfclusters, evt, fParams.PFParticleLabel() + slice_tag_suff);
+        for (size_t icl=0; icl<fmCluHits.size();icl++) {
+          for (auto hit : fmCluHits.at(icl)) {
```
PetrilloAtWork, line 1857: Avoid copies:
```suggestion
          for (auto const& hit : fmCluHits.at(icl)) {
```

sbncode/CAFMaker/CAFMaker_module.cc
```diff
@@ -2304,12 +2045,72 @@ void CAFMaker::produce(art::Event& evt) noexcept {
...
     const recob::Vertex *vertex = (iPart == fmPFPart.size() || !fmVertex.at(iPart).size()) ? NULL : fmVertex.at(iPart).at(0).get();
 
     //#######################################################
     // Add slice info.
     //#######################################################
+    if (std::find(nuGraphSlices.begin(), nuGraphSlices.end(), slice) != nuGraphSlices.end()) {
+      std::vector<art::Ptr<anab::FeatureVector<1>>> ng2_filter_vec;
+      std::vector<art::Ptr<anab::FeatureVector<5>>> ng2_semantic_vec;
+      art::FindOneP<anab::FeatureVector<1>> findOneFilter(slcHits, evt, fParams.NuGraphFilterLabel().label() + slice_tag_suff + ":" + fParams.NuGraphFilterLabel().instance());
+      art::FindOneP<anab::FeatureVector<5>> findOneSemantic(slcHits, evt, fParams.NuGraphSemanticLabel().label() + slice_tag_suff + ":" + fParams.NuGraphSemanticLabel().instance());
```
PetrilloAtWork, line 2059: It's hard to see whether the labels are the same. Define the tag first:
```suggestion
      art::InputTag const filtTag{
        fParams.NuGraphFilterLabel().label() + slice_tag_suff,
        fParams.NuGraphFilterLabel().instance()
        };
      art::FindOneP<anab::FeatureVector<1>> findOneFilter(slcHits, filtTag);
      art::FindOneP<anab::FeatureVector<5>> findOneSemantic(slcHits, filtTag);
```

sbncode/CAFMaker/CAFMaker_module.cc
```diff
@@ -2304,12 +2045,72 @@ void CAFMaker::produce(art::Event& evt) noexcept {
...
+        for (size_t hitIdx = 0; hitIdx < slcHits.size(); ++hitIdx) {
+          ng2_semantic_vec.emplace_back(findOneSemantic.at(hitIdx));
+        }
+      }
+
+      // vertex projection onto the three wire planes
+      float vtx_wire[3];
+      float vtx_tick[3];
+
+      if (vertex != NULL) {
```
PetrilloAtWork, line 2081: What kind of garbage is going to end into the tree if the vertex is null? The arrays need to be initialised somehow.
This is a **request**, as we don't want uninitialised values to end into the trees as it compromises tree comparisons.
Also, `NULL` is C, and we need to keep a façade here.
```suggestion
      if (vertex) {
```

sbncode/CAFMaker/CAFMaker_module.cc
```diff
@@ -2304,12 +2045,72 @@ void CAFMaker::produce(art::Event& evt) noexcept {
...
+        auto const& tpcID = geom->FindTPCAtPosition(vertex->position());
+        if (tpcID.isValid) {
+          for (geo::PlaneID const& p : wireReadout.Iterate<geo::PlaneID>()) {
+            auto const& planeID = geo::PlaneID{tpcID, p.Plane};
+            const geo::PlaneGeo& planeGeo = wireReadout.Plane(planeID);
```
PetrilloAtWork, line 2086: This should work:
```suggestion
          for (auto const& planeGeo : wireReadout.Iterate<geo::PlaneGeo>(tpcID)) {
```
(plane number is `planeGeo.ID().Plane`).

sbncode/CAFMaker/CAFMaker_module.cc
```diff
@@ -2304,12 +2045,72 @@ void CAFMaker::produce(art::Event& evt) noexcept {
...
+            vtx_wire[p.Plane] = planeGeo.WireCoordinate(vertex->position()); ///< wire projection
+            vtx_tick[p.Plane] = dprop.ConvertXToTicks(vertex->position().X(), planeID); ///< drift projection
+          }
+        }
+      }
+
+      if (ng2_filter_vec.size() > 0 || ng2_semantic_vec.size() > 0) {
```
PetrilloAtWork, line 2093: Semantics:
```suggestion
      if (!ng2_filter_vec.empty() || !ng2_semantic_vec.empty()) {
```

sbncode/CAFMaker/CAFMaker_module.cc
```diff
@@ -2304,12 +2045,72 @@ void CAFMaker::produce(art::Event& evt) noexcept {
...
+        FillSliceNuGraph(slcHits, ng2_filter_vec, ng2_semantic_vec, fmPFPartHits, 
+                         vtx_wire, vtx_tick, fParams.NuGraphHIPTagWireDist(), fParams.NuGraphHIPTagTickDist(), 
+                         fParams.NuGraphFilterCut(), recslc);
+      }
+
+      // overwrite the `ng_filt_pass_frac` variable with post-NuGraph2-filtering 
+      if (fParams.UsePandoraAfterNuGraph()) {
+        art::Handle<std::vector<anab::FeatureVector<1>>> ngFilterVecHandle;
+        evt.getByLabel(art::InputTag("NGMultiSlice" + slice_tag_suff, "filter"), ngFilterVecHandle);
```
PetrilloAtWork, line 2102: Smoother (less repetitions, less maintenance):
```suggestion
        auto ngFilterVecHandle = evt.getHandle<std::vector<anab::FeatureVector<1>>>
          (art::InputTag("NGMultiSlice" + slice_tag_suff, "filter"));
```

sbncode/CAFMaker/CAFMaker_module.cc
```diff
@@ -2304,12 +2045,72 @@ void CAFMaker::produce(art::Event& evt) noexcept {
...
+        art::Handle<std::vector<recob::Hit>> ngFilteredHitsHandle;
+        GetByLabelStrict(evt, "ngfilteredhits" + slice_tag_suff, ngFilteredHitsHandle);
```
PetrilloAtWork, line 2104: Why is this "strict" and the previous is not?

sbncode/CAFMaker/CAFMaker_module.cc
```diff
@@ -2304,12 +2045,72 @@ void CAFMaker::produce(art::Event& evt) noexcept {
...
+        if (ngFilterVecHandle.isValid() && 
+            ngFilteredHitsHandle.isValid() &&
+            ngFilterVecHandle->size() > 0) {
+          recslc.ng_filt_pass_frac = float(ngFilteredHitsHandle->size()) / float(ngFilterVecHandle->size());
```
PetrilloAtWork, line 2108: Single precision `floats` have only <7 digits, it may lose information when there are millions of hits. Probably not a problem.
Anyway, since the hardware likely converts to intermediate double anyway...

```suggestion
          recslc.ng_filt_pass_frac = double(ngFilteredHitsHandle->size()) / double(ngFilterVecHandle->size());
```

sbncode/CAFMaker/CAFMaker_module.cc
```diff
@@ -2427,32 +2227,39 @@ void CAFMaker::produce(art::Event& evt) noexcept {
...
+          }
+        }
+
+        // semantic tagging
+        if (findOneSemantic.isValid()) {
+          ng2_semantic_vec.reserve(PFPHits.size());
+          for (size_t hitIdx = 0; hitIdx < PFPHits.size(); ++hitIdx) {
+            ng2_semantic_vec.emplace_back(findOneSemantic.at(hitIdx));
+          }
+        }
```
PetrilloAtWork, line 2258: This feels quite familiar... same comments as for twin code above.
Also, consider moving into a member function 
```cpp
std::tuple<
  std::vector<art::Ptr<anab::FeatureVector<1>>>,
  std::vector<art::Ptr<anab::FeatureVector<5>>>
  >
CAFMaker::readNuGraphHitFeatures(
  std::vector<art::Ptr<recob::Hit>> const& hits,
  art::Event const& event, std::string const& suffix
) const;
```

sbncode/CAFMaker/FillReco.cxx
```diff
@@ -743,22 +579,107 @@ namespace caf
...
   void FillSliceNuGraph(const std::vector<art::Ptr<recob::Hit>> &inputHits,
-			const std::vector<unsigned int> &sliceHitsMap,
 			const std::vector<art::Ptr<anab::FeatureVector<1>>> &ngFilterResult,
 			const std::vector<art::Ptr<anab::FeatureVector<5>>> &ngSemanticResult,
+      const std::vector<std::vector<art::Ptr<recob::Hit>>> &fmPFPartHits,
+      const float vtx_wire[3], 
+      const float vtx_tick[3],
+      const float vtx_wire_dist, 
+      const float vtx_tick_dist,
+      const float filter_cut,
```
PetrilloAtWork, line 589: From GitHub it seems the indentation of these lines is different from the previous ones. Check and fix as appropriate.

sbncode/CAFMaker/FillReco.cxx
```diff
@@ -743,22 +579,107 @@ namespace caf
...
     unsigned int nHits = inputHits.size();
-    if (nHits==0 || nHits!=sliceHitsMap.size() || inputHits[0].key()!=sliceHitsMap[0]) return;//not the same slice!
-
     unsigned int npass = 0;
+
     for ( unsigned int i = 0; i < nHits; i++ ) {
-      if (ngFilterResult.at(i)->at(0)>=ng_filter_cut) npass++;
+      if (ngFilterResult.at(i).isNull()) continue;
+      if (ngFilterResult.at(i)->at(0) >= filter_cut) npass++;
+    }
```
PetrilloAtWork, line 600: Avoid indices since they are not used. Also note that at this point it is not obvious that the vectors have the same size. The assertion explicitly "documents" that assumption.
```suggestion
    assert(ngFilterResult.size() == nHits);
    for ( auto const& ngFilt: ngFilterResult) {
      if (!ngFilt) continue;
      if (ngFilt->at(0) >= filter_cut) npass++;
    }
```

sbncode/CAFMaker/FillReco.cxx
```diff
@@ -743,22 +579,107 @@ namespace caf
...
+    // NuGraph2 plane-by-plane slice variables
+    for (unsigned int plane = 0; plane < 3; ++plane) {
+
+      int nMIPHits = 0;             ///< `MIP` hits in the slice
+      int nHIPHits = 0;             ///< `HIP` hits in the slice
+      int nShrHits = 0;             ///< `Shower` hits in the slice
+      int nMhlHits = 0;             ///< `Michel` hits in the slice
+      int nDifHits = 0;             ///< `Diffuse` hits in the slice
+      int nVtxHIPHits = 0;          ///< `HIP` hits in the slice around the vertex
+      int nUnclusteredShrHits = 0;  ///< `Shower` hits in the slice not belonging to a PFP
```
PetrilloAtWork, line 620: Avoid Doxygen comments in the middle of the implementation.

```suggestion
      int nMIPHits = 0;             // `MIP` hits in the slice
      int nHIPHits = 0;             // `HIP` hits in the slice
      int nShrHits = 0;             // `Shower` hits in the slice
      int nMhlHits = 0;             // `Michel` hits in the slice
      int nDifHits = 0;             // `Diffuse` hits in the slice
      int nVtxHIPHits = 0;          // `HIP` hits in the slice around the vertex
      int nUnclusteredShrHits = 0;  // `Shower` hits in the slice not belonging to a PFP
```

sbncode/CAFMaker/FillReco.cxx
```diff
@@ -743,22 +579,107 @@ namespace caf
...
+
+      for ( unsigned int i = 0; i < nHits; i++ ) {
+        const recob::Hit& hit = *inputHits.at(i);
+        if (hit.WireID().Plane != plane) continue;
+
+        if (ngSemanticResult.at(i).isNull()) continue;
+
+        auto const& sem = *ngSemanticResult.at(i);
+        std::vector<float> semVec;
+        for (size_t k = 0; k < sem.size(); ++k) semVec.push_back(sem.at(k));
```
PetrilloAtWork, line 630: If `sem` supports iterators, then this should work:

```suggestion
        std::vector<float> semVec{ sem.begin(), sem.end() };
```

sbncode/CAFMaker/FillReco.cxx
```diff
@@ -743,22 +579,107 @@ namespace caf
...
+        auto highestScoreIdx = std::distance(
+          semVec.begin(), 
+          std::max_element(semVec.begin(), semVec.end())
+        );
+
+        // `MIP` hits
```
PetrilloAtWork, line 636: This sequence may be better served with a `switch` block.

sbncode/CAFMaker/FillReco.cxx
```diff
@@ -743,22 +579,107 @@ namespace caf
...
+        }
+      }
+
+      slice.ng_plane[plane].mip_hits = nMIPHits;
+      slice.ng_plane[plane].hip_hits = nHIPHits;
+      slice.ng_plane[plane].shr_hits = nShrHits;
+      slice.ng_plane[plane].mhl_hits = nMhlHits;
+      slice.ng_plane[plane].dif_hits = nDifHits;
+      slice.ng_plane[plane].ng_vtx_hip_hits = nVtxHIPHits;
+      slice.ng_plane[plane].unclustered_shr_hits = nUnclusteredShrHits;
```
PetrilloAtWork, line 681:
```suggestion
      auto& hitInfo = slice.ng_plane[plane];
      hitInfo.mip_hits = nMIPHits;
      hitInfo.hip_hits = nHIPHits;
      hitInfo.shr_hits = nShrHits;
      hitInfo.mhl_hits = nMhlHits;
      hitInfo.dif_hits = nDifHits;
      hitInfo.ng_vtx_hip_hits = nVtxHIPHits;
      hitInfo.unclustered_shr_hits = nUnclusteredShrHits;
```

sbncode/CAFMaker/FillReco.cxx
```diff
@@ -1239,39 +1107,34 @@ namespace caf
...
-	if (bkgscore->at(0)<ng_filter_cut) {
-	  ng2bkgpfpcount++;
-	} else {
-	  auto const& scores = ngSemanticResult.at(pos);
-	  std::vector<float> ng2semscores;
-	  for (size_t i=0;i<scores->size();i++) ng2semscores.push_back(scores->at(i));
-	  size_t sem_label = std::distance(ng2semscores.begin(), std::max_element(ng2semscores.begin(), ng2semscores.end()));//arg_max(ng2semscores);
-	  ng2sempfpcounts[sem_label]++;
-	}
+        if (ngFilterResult.at(pos).isNull()) continue;
```
PetrilloAtWork, line 1123: Indentation seems off.

sbncode/CAFMaker/FillReco.cxx
```diff
@@ -1239,39 +1107,34 @@ namespace caf
...
+
+	      auto const& bkgscore = ngFilterResult.at(pos);
+	      if (bkgscore->at(0) < filter_cut) {
+	        ng2bkgpfpcount++;
+	      } else {
+          if (ngSemanticResult.at(pos).isNull()) continue;
+          auto const& scores = ngSemanticResult.at(pos);
+          std::vector<float> ng2semscores;
+          for (size_t i=0;i<scores->size();i++) ng2semscores.push_back(scores->at(i));
```
PetrilloAtWork, line 1132:
```suggestion
          std::vector<float> ng2semscores{ scores.begin(), scores.end() };
```

sbncode/CAFMaker/FillReco.h
```diff
@@ -179,15 +163,16 @@ namespace caf
...
    * @param[out] srpfp the destination PFParticle object
    *
    * Hits with filter value (`ngFilterResult`) lower than `ng_filter_cut` are counted as background.
    */
-  void FillPFPNuGraph(const std::vector<unsigned int> &sliceHitsMap,
+  void FillPFPNuGraph(const std::vector<art::Ptr<recob::Hit>> &pfpHits,
 		      const std::vector<art::Ptr<anab::FeatureVector<1>>> &ngFilterResult,
 		      const std::vector<art::Ptr<anab::FeatureVector<5>>> &ngSemanticResult,
-		      const std::vector<art::Ptr<recob::Hit>> &pfpHits,
+          const float filter_cut,
```
PetrilloAtWork, line 175: Fix the indentation.

PetrilloAtWork (changes requested):
My usual burst of comments.
There is at least one which must be addressed, about uninitialised variables.

sbncode/CAFMaker/CAFMaker_module.cc
rtriozzi: going through the comments right now! about this: every slice has to have a vertex, sort of by definition in Pandora (e.g., even if the reconstruction is sure that the slice is a cosmic, the vertex will just be the highest-y hit)

I think there are also protections for failed vertexing (which should be _very_ rare: we really just need a hit to define a vertex...) within Pandora, so that we always get something when we grab the vertex in the CAF maker...

anyway, I agree it's bad practice from the CAF maker stand point (+ the `NULL` check was therefore strange), I'll fix this! thanks!

sbncode/CAFMaker/CAFMaker_module.cc
rtriozzi: Mh. I do not have recollection of me writing this code, as this was a year ago. It doesn't look like an issue however, so I'm marking this as resolved!

sbncode/CAFMaker/FillReco.cxx
rtriozzi: True. Fixed.

sbncode/CAFMaker/FillReco.h
rtriozzi: Looks OK on my code editor.

sbncode/CAFMaker/CAFMaker_module.cc
rtriozzi: This is right. Didn't spot the `false` case bug - thaaaaanksss!!

sbncode/CAFMaker/CAFMaker_module.cc
PetrilloAtWork: If you strongly believe in that (and assuming that only Pandora and like-minded code execute this code, which I would know if I went back to read everything that I have forgotten about it), the best approach is to remove the `if` branch and replace it with an assertion (`assert(vertex);`), which "documents" the requirement.

PetrilloAtWork (approved):
Thank you! Approved.

PR #619, opened 2025-12-10
Light Calorimetry: add sim energy deposit info and light calo info into cafs

lynnt20:
### Description 
Adds info from `SimEnergyDeposits` and light calorimetry into cafmaker. 

SBND is currently keeping `SimEnergyDeposits` throughout the workflow. We can take better advantage of this information (truth level number of electrons, photons, and true energy deposits for true neutrino interactions) by adding it to the cafs in a new class called `SRTrueDeposit`. Default will be empty if no `SimEnergyDeposit`s are available in the input files. Only adds 3 floats per neutrino interaction (obtained from length of `MCTruth`).

Accompanying PRs: SBNSoftware/sbndcode#878, SBNSoftware/sbnanaobj#181, SBNSoftware/sbnobj#158

- [ ] Have you assigned at least 1 reviewer?
- [X] Is this PR related to an open issue / project?
- [x] Does this PR affect CAF data format? If so, please assign a CAF maintainer as additional reviewer.


kjplows:
@lynnt20  could you nominate reviewers please?

lynnt20:
@PetrilloAtWork would you mind taking a look at this PR? It's the last of the light calorimetry stack of PRs 😃 Thank you advance!

sbncode/CAFMaker/CAFMakerParams.h
```diff
@@ -483,6 +483,12 @@ namespace caf
       "simdrift"
     };
 
+    Atom<art::InputTag> SimEnergyDepositLabel {
+      Name("SimEnergyDepositLabel"),
+      Comment("Label of input sim::SimEnergyDeposit objects."),
+      art::InputTag("ionandscint", "priorSCE","G4")
```
henrylay97: What's the motivation for explicitly naming the process name as "G4"? Are we expecting scenarios where multiple fcl processes have created SimEnergyDeposits? Definitely not asking for a change here as this is nice and fcl flexible. Just interested in why this came up!

henrylay97 (commented):
Looking really good. Couple of small queries before I approve :D

sbncode/CAFMaker/CAFMaker_module.cc
```diff
@@ -1637,6 +1648,34 @@ void CAFMaker::produce(art::Event& evt) noexcept {
     } // end for fm
   } // end for i (mctruths)
 
+
+  if (!isRealData && sed_handle.isValid()){
+    art::ServiceHandle<cheat::ParticleInventoryService> pi_serv;
+
+    srtruthbranch.dep.reserve(mctruths.size());
+    for (size_t n=0; n<mctruths.size();n++){
+      SRTrueDeposit init;
```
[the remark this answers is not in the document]
lynnt20: I could go either way about it, not sure that I feel very strongly as long as you think it's documented well in the corresponding header. It may be confusing if one expects it to be exactly what's in `SimEnergyDeposit`s, but I think what the branch is actually accessible with (which is `rec.mc.dep.energy`, etc., is not too bad?

henrylay97 (approved):

sbncode/CAFMaker/CAFMaker_module.cc
```diff
@@ -1637,6 +1648,34 @@ void CAFMaker::produce(art::Event& evt) noexcept {
...
+      init.electrons = 0;
+      init.photons = 0;
+      init.energy = 0; 
+      srtruthbranch.dep.push_back(init);
+    }
+
+    for (size_t n_dep=0; n_dep < seds.size(); n_dep++){
+      auto sed = seds[n_dep];
+      const auto trackID = sed->TrackID();
```
PetrilloAtWork: Removing the vector of pointers:
```suggestion
    for (sim::SimEnergyDeposit const& sed: *sed_handle){
      const auto trackID = sed.TrackID();
```
The three dereference operators below should also be changed: `sed->` → `sed.`.

sbncode/CAFMaker/CAFMaker_module.cc
```diff
@@ -2049,6 +2088,13 @@ void CAFMaker::produce(art::Event& evt) noexcept {
      if (fmCorrectedOpFlash.isValid())
       slcCorrectedOpFlash = fmCorrectedOpFlash.at(0);
 
+    art::FindOneP<sbn::LightCalo> foLightCalo =
+      FindOnePStrict<sbn::LightCalo>(sliceList,evt,
+        fParams.LightCaloLabel() + slice_tag_suff);
+    const sbn::LightCalo *slcLightCalo = nullptr;
+    if (foLightCalo.isValid()) {
+      slcLightCalo = foLightCalo.at(0).get();
+    }
```
PetrilloAtWork: You can compact this using the ternary operator:

```suggestion
    const sbn::LightCalo *slcLightCalo = foLightCalo.isValid()? foLightCalo.at(0).get(): nullptr;
```

sbncode/CAFMaker/CAFMakerParams.h
PetrilloAtWork: Not a lover of either 🤷

PetrilloAtWork (changes requested):
The PR is generally ok.
I requested the removal of the intermediate vector of art pointers, which is a bad practice being propagated since ages that I am trying to extinguish.

In addition, there are version number changes that should be left to the release managers. I am not sure what is going on, since it appears a release branch was merged here, and then this PR is requested to be merged with `develop`. If this is really what is needed, the merge of the release should happen before this PR is merged. Maybe that's the plan and I don't know it.

PetrilloAtWork (approved):
Approved.

PR #623, opened 2026-01-14
CAFMaker Truth Bugfix

nathanielerowe:
The CI was picking up on some non-deterministic behavior at CAF level. It turns out we had some loops which were hardcoded to values which were not appropriate for SBND. This should resolve some of them (there may be more).


sbncode/CAFMaker/FillTrue.cxx
```diff
@@ -599,7 +599,7 @@ namespace caf {
       }
 
       // Set the cryostat of the position
-      for (int icryo = 0; icryo < 2; icryo++) {
+      for (int icryo = 0; icryo < (int)active_volumes.size(); icryo++) {
```
PetrilloAtWork: Technically it would be more correct to have `icryo` of index type (`std::size_t`) since its name suggest it being an index, and then converted in the assignment on line 604.
But I will let it pass; it's mostly to prove that I can make a comment also to a one-liner.

PR #649, opened 2026-05-07
WireMod for SBND Gen 2 Production

tbwester:
### Description 
Add WireMod utility functions for SBND Gen 2. This also brings in refactor of RecoUtils. The WireMod utility should be identical to the SBND Gen 1 production based on v10_06_00_10. Develop PR to follow.

### DIFF
--- a/sbncode/CAFMaker/RecoUtils/RecoUtils.cc
+++ b/sbncode/CAFMaker/RecoUtils/RecoUtils.cc
@@ -107,3 +107,151 @@ int CAFRecoUtils::GetShowerPrimary(const int g4ID)
 
   return primary_id;
 }
+
+std::map<int, std::vector<sbn::ReadoutIDE>> sbn::PrepSimChannels(const std::vector<art::Ptr<sim::SimChannel>> &simchannels, const geo::WireReadoutGeom &wireReadout) {
+    std::map<int, std::vector<sbn::ReadoutIDE>> ret;
+
+    for (const art::Ptr<sim::SimChannel> sc : simchannels) {
+      // Lookup the wire of this channel
+      raw::ChannelID_t channel = sc->Channel();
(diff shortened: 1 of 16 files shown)


sbncode/CAFMaker/RecoUtils/RecoUtils.cc
```diff
@@ -107,3 +107,151 @@ int CAFRecoUtils::GetShowerPrimary(const int g4ID)
...
+
+std::map<int, std::vector<sbn::ReadoutIDE>> sbn::PrepSimChannels(const std::vector<art::Ptr<sim::SimChannel>> &simchannels, const geo::WireReadoutGeom &wireReadout) {
+    std::map<int, std::vector<sbn::ReadoutIDE>> ret;
+
+    for (const art::Ptr<sim::SimChannel> sc : simchannels) {
+      // Lookup the wire of this channel
+      raw::ChannelID_t channel = sc->Channel();
+      std::vector<geo::WireID> maybewire = wireReadout.ChannelToWire(channel);
+      geo::WireID thisWire; // Default constructor makes invalid wire
+      if (maybewire.size()) thisWire = maybewire[0];
```
hausnerh, line 119: How does this handle when `ChannelToWire` returns multiple `WireID`s? This can happen near the middle of ICARUS where the wire geometry is split into two "logical" TPC for North and South and the wires crossing there get split into two `WireID`s despite reading out from the same channel. See chapter 5 [here](https://sbn-docdb.fnal.gov/cgi-bin/sso/RetrieveFile?docid=25068&filename=ICARUSTPCchannelMapping.pdf&version=1) for some details.

tbwester: My understanding is as long as either wire correctly maps to the same sim::IDE, it shouldn't matter which one is used here. Maybe @gputnam can comment on this, since it's not something I've considered!

PR #650, opened 2026-05-07
WireMod for Develop

tbwester:
### Description 
Add WireMod utility functions to `develop`. This also brings in refactor of RecoUtils. The WireMod utility should be identical to the SBND Gen 1 production based on v10_06_00_10. This is the corresponding PR to sbncode #649.

- [x] Have you added a label? (bug/enhancement/physics etc.)
the description.
- [X] Are you submitting this PR on behalf of someone else who made the code changes? If so, please mention them in the description.

### DIFF
--- a/sbncode/CAFMaker/FillTrue.cxx
+++ b/sbncode/CAFMaker/FillTrue.cxx
@@ -5,7 +5,6 @@
 #include "larcorealg/Geometry/WireReadoutGeom.h"
 #include "larevt/SpaceCharge/SpaceCharge.h"
 #include "larcore/CoreUtils/ServiceUtil.h"
-#include "RecoUtils/RecoUtils.h"
 
 #include "CLHEP/Random/RandGauss.h"
 
@@ -136,7 +135,7 @@ namespace caf {
   }//FillTrackTruth
 
   // Assumes truth matching and calo-points are filled
-  void FillTrackCaloTruth(const std::map<int, std::vector<std::pair<geo::WireID, const sim::IDE*>>> &id_to_ide_map,
+  void FillTrackCaloTruth(const std::map<int, std::vector<sbn::ReadoutIDE>> &id_to_ide_map,
                           const std::vector<simb::MCParticle> &mc_particles,
                           const geo::GeometryCore& geometry,
                           const geo::WireReadoutGeom& wireReadout,
@@ -163,10 +162,10 @@ namespace caf {
 
     // Load the hits
     // match on the channel, which is unique
-    const std::vector<std::pair<geo::WireID, const sim::IDE*>> &match_ides = id_to_ide_map.at(srtrack.truth.p.G4ID);
+    const std::vector<sbn::ReadoutIDE> &match_ides = id_to_ide_map.at(srtrack.truth.p.G4ID);
     std::map<unsigned, std::vector<const sim::IDE *>> chan_2_ides;
-    for (auto const &ide_pair: match_ides) {
-      chan_2_ides[wireReadout.PlaneWireToChannel(ide_pair.first)].push_back(ide_pair.second);
+    for (auto const &ide_p: match_ides) {
+      chan_2_ides[wireReadout.PlaneWireToChannel(ide_p.wire)].push_back(ide_p.ide);
     }
 
     // pre-compute partial ranges
@@ -634,15 +633,15 @@ namespace caf {
   void FillTrueG4Particle(const simb::MCParticle &particle,
         const std::vector<geo::BoxBoundedGeo> &active_volumes,
         const std::vector<std::vector<geo::BoxBoundedGeo>> &tpc_volumes,
-        const std::map<int, std::vector<std::pair<geo::WireID, const sim::IDE *>>> &id_to_ide_map,
+        const std::map<int, std::vector<sbn::ReadoutIDE>> &id_to_ide_map,
         const std::map<int, std::vector<art::Ptr<recob::Hit>>> &id_to_truehit_map,
         const cheat::BackTrackerService &backtracker,
         const cheat::ParticleInventoryService &inventory_service,
         const std::vector<art::Ptr<simb::MCTruth>> &neutrinos,
                           caf::SRTrueParticle &srparticle) {
 
-    std::vector<std::pair<geo::WireID, const sim::IDE *>> empty;
-    const std::vector<std::pair<geo::WireID, const sim::IDE *>> &particle_ides = id_to_ide_map.count(particle.TrackId()) ? id_to_ide_map.at(particle.TrackId()) : empty;
+    std::vector<sbn::ReadoutIDE> empty;
+    const std::vector<sbn::ReadoutIDE> &particle_ides = id_to_ide_map.count(particle.TrackId()) ? id_to_ide_map.at(particle.TrackId()) : empty;
 
     std::vector<art::Ptr<recob::Hit>> emptyHits;
     const std::vector<art::Ptr<recob::Hit>> &particle_hits = id_to_truehit_map.count(particle.TrackId()) ? id_to_truehit_map.at(particle.TrackId()) : emptyHits;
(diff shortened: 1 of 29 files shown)


gputnam (approved):
Looks good! Insofar as it intersects with stuff I've contributed. It looks like a SinglePhoton analysis commit got in here somehow though? Bit confused by that. But I don't see that as holding this up.

hausnerh:
The WireMod utility portions look good to me. However, I do still have some concerns about the `geo::WireID` as commented above. Perhaps @PetrilloAtWork is the expert who could elucidate if my concerns about mapping the `geo::WireID`s and the `sim::SimChannel`s are warranted.

tbwester:
Converted to a draft while waiting for #654 to be merged.

sbncode/CAFMaker/FillTrue.cxx
```diff
@@ -662,9 +661,9 @@ namespace caf {
...
 
-    for (auto const &ide_pair: particle_ides) {
-      const geo::WireID &w = ide_pair.first;
-      const sim::IDE *ide = ide_pair.second;
+    for (auto const &ide_p: particle_ides) {
+      const geo::WireID &w = ide_p.wire;
+      const sim::IDE *ide = ide_p.ide;
 
       if(w.Plane >= 0 && w.Plane < 3 && w.Cryostat < 2){
         srparticle.plane[w.Cryostat][w.Plane].visE += ide->energy / 1000. /* MeV -> GeV*/;
```
PetrilloAtWork, line 669: To help addressing @hausnerh's concern: here the wire number, which is the only ambiguous part when going from a channel to a wire ID, is not used. I suggest this code makes it explicit:
```suggestion
      const geo::PlaneID &p = ide_p.wire.asPlaneID();
      const sim::IDE *ide = ide_p.ide;

if(p.Plane >= 0 && p.Plane < 3 && p.Cryostat < 2){
srparticle.plane[p.Cryostat][p.Plane].visE += ide->energy / 1000. /* MeV -> GeV*/;
```
(`.asPlaneID()` is not necessary since conversion can be implicit, but it helps the reader)

sbncode/Calibration/TrackCaloSkimmer_module.cc
PetrilloAtWork, line 620: Here only the plane number is used:
```suggestion
  for (auto const &ide_p: particle_ides) {
    const geo::PlaneID::Plane_t p = ide_p.wire.Plane;
    const sim::IDE *ide = ide_p.ide;
    switch (p) {
      case 0:
        trueparticle.plane0VisE += ide->energy / 1000. /* MeV -> GeV*/;
        break;
      case 1:
        trueparticle.plane1VisE += ide->energy / 1000. /* MeV -> GeV*/;
        break;
      case 2:
        trueparticle.plane2VisE += ide->energy / 1000. /* MeV -> GeV*/;
        break;
    }
  }
```
If you don't care of that level of pedantry, an `int` instead of the format `geo::PlaneID::Plane_t` will do just as well.
I find a `switch` more appropriate here, but that's not really the point of this comment.

sbncode/Calibration/TrackCaloSkimmer_module.cc
```diff
@@ -791,10 +792,10 @@ sbn::TrueParticle TrueParticleInfo(const simb::MCParticle &particle,
   }
 
   // Compute widths
-  for (auto const &ide_pair: particle_ides) {
-    const geo::WireID &w = ide_pair.first;
+  for (auto const &ide_part: particle_ides) {
+    const geo::WireID &w = ide_part.wire;
```
PetrilloAtWork, line 796: Here:
 * `PlaneWireToChannel()` will correctly track back to the channel, no matter which of its wire was chosen, so it's not a problem;
 * `WireToTrajectoryPosition()` (whose name is wrong) uses only the TPC information, so one can make that explicit too:
 
```suggestion
  for (auto const &ide_part: particle_ides) {
    const geo::WireID &w = ide_part.wire;
    const geo::TPCID &t = w.asTPCID();
```
and then use `t` in the following call.

sbncode/Calibration/TrackCaloSkimmer_module.cc
```diff
@@ -748,19 +746,20 @@ sbn::TrueParticle TrueParticleInfo(const simb::MCParticle &particle,
...
   trueparticle.G4ID = particle.TrackId();
   trueparticle.parent = particle.Mother();
 
   // Organize deposition info into per-wire true "Hits" -- key is the Channel Number
   std::map<unsigned, sbn::TrueHit> truehits; 
 
-  for (auto const &ide_pair: particle_ides) {
-    const geo::WireID &w = ide_pair.first;
+  for (auto const &ide_part: particle_ides) {
+    const geo::WireID &w = ide_part.wire;
```
PetrilloAtWork, line 759: This can be treated as the case below (sorry for the back-and-forth)...
```suggestion
    const geo::WireID &w = ide_part.wire;
    const geo::TPCID &t = w.asTPCID();
```
with one exception.

sbncode/Calibration/TrackCaloSkimmer_module.cc
PetrilloAtWork, line 768: This one looks like the only case where the wrong wire number may have an effect.
If this is unacceptable, then one (expensive) workaround is to extracted the wire again:
```cpp
    // track down the actual TPC and wire for the IDE
    geo::Point_t const IDEpos{ ide->x, ide->y, ide->z };
    geo::TPCID const IDETPC = wireReadout->FindTPCatPosition(IDEpos);
    geo::WireID const IDEwire = IDETPC.isValid
      ? wireReadout->NearestWireID(IDEpos, { IDETPC, w.Plane }): geo::WireID{};
```
Note that we still select the relevant plane number from the cached wire.
Do this above, and then you can use `IDETPC` for every time you need a TPC ID (if you need a plane ID too, better to define it with `geo::planeID const IDEplane{ IDETPC, w.Plane };`).
However, note that this returns the real TPC and wire of the IDE; if that IDE was shifted to simulate some displacement, its wire may end up being a different one than `w`, and on a different channel than `c`.

sbncode/CAFMaker/RecoUtils/RecoUtils.cc
```diff
@@ -107,3 +107,151 @@ int CAFRecoUtils::GetShowerPrimary(const int g4ID)
...
+
+std::map<int, std::vector<sbn::ReadoutIDE>> sbn::PrepSimChannels(const std::vector<art::Ptr<sim::SimChannel>> &simchannels, const geo::WireReadoutGeom &wireReadout) {
+    std::map<int, std::vector<sbn::ReadoutIDE>> ret;
+
+    for (const art::Ptr<sim::SimChannel> sc : simchannels) {
+      // Lookup the wire of this channel
+      raw::ChannelID_t channel = sc->Channel();
+      std::vector<geo::WireID> maybewire = wireReadout.ChannelToWire(channel);
+      geo::WireID thisWire; // Default constructor makes invalid wire
+      if (maybewire.size()) thisWire = maybewire[0];
```
PetrilloAtWork, line 119: Considering that
* often this value is not checked downstream
* a `SimChannel` should not be associated to wireless channels

I suggest to make this function loudly complain in case of inconsistency:
```suggestion
      if (maybewire.empty()) {
        if (sc->TDCIDEMap().empty()) continue;
        throw cet::exception("sbn::PrepSimChannels")
          << "Data on sim::SimChannel CH=" << channel
          << ", which has no associated wires!\n";
      }
      thisWire = maybewire[0];
```

PetrilloAtWork (commented):
Since @hausnerh asked for a comment, I have decorated the module with a quick analysis of the issue.
The question is whether any ambiguity from `ChannelToWire()` result might create trouble.
My findings:
 * there _is_ the potential for ambiguity, since that function can and will return multiple wires. The wires will all belong to the same cryostat and plane, but will have different wire number and may have (usually do have) different TPC number;
 * the code uses `sbn::ReadoutIDE` to carry around the information, and that one does not have room for a channel number, only fora wire ID;
 * most of the occurrences of the use of this wire ID are to select the plane _number_, which is expected to be always correct;
 * one exception is the use of `WireToTrajectoryPosition()` which, despite the name, just corrects a 3D point for field distortions; this function uses only the TPC ID, not the wire number. If I recall correctly, the TPC ID is used only to determine the drift direction (after all, we know the exact location of the point). While the TPC ID may be wrong, in practice in SBND all TPCs sharing a channel have the same drift direction, so this does not become an issue (DUNE APA would be a different story).
 * the other exception is the wire and TPC number values being stored in the output. **This is the only instance where the effects of the potentially wrong information will be observable**. I have left a possible workaround.

Do with it what you will :-D

tbwester:
Superseded by PR #663

PR #659, opened 2026-07-03
Adding "safe-guards" to avoid segfault in the GaussHitFinderSBN module

mattiasotgia:
# The issue

The introduction of the `NextADCThreshold` parameter in LArSoft/larreco#92 highlighted some minor bugs in the `GaussHitFinderSBN` module that made the code segfault for some edge case. 

This PR aims at adding some safeguards to the code preventing the segfault to happen. 

# Description

The changes happens in three points of the code

1. After the `peakFitterTool` returns the peak parameters for each "candidate hit". Here we added a check for all "not finite" bounds, whereas the old version of the code was only checking for the positive side. Since the $\chi^2/\mathrm{ndf}$ is computed from all the fitted hits, $-\infty$ values were possible, and treated as "not pathological". This fixes those cases. 
2. After all candidates have been fitted and failed fits treated accordingly, when computing the boundaries of the hits we added a check (line 501) making sure the start/end points where still withing the ROI range. 
3. Finally after all the safe-guards at the end of the module are applied, we make sure (line 555) that no "boundary flip" has happened (i.e. start > end)

### DIFF
--- a/sbncode/HitFinder/GaussHitFinderSBN_module.cc
+++ b/sbncode/HitFinder/GaussHitFinderSBN_module.cc
@@ -388,14 +388,8 @@ namespace hit {
                 fPeakFitterTool->findPeakParameters(
                   range.data(), mergedCands, peakParamsVec, chi2PerNDF, NDF);
 
-                // If the chi2 is infinite then there is a real problem so we bail
-                if (!(chi2PerNDF < std::numeric_limits<double>::infinity())) {
-                  chi2PerNDF = 2. * fChi2NDF.at(plane);
-                  NDF = 2;
-                }
-
-                // If the chi2 is also low (low-er than -inf) we bail
-                if (!(chi2PerNDF > -std::numeric_limits<double>::infinity())) {
+                // If the chi2 is not finite then there is a real problem so we bail
+                if (!std::isfinite(chi2PerNDF)) {
                   chi2PerNDF = 2. * fChi2NDF.at(plane);
                   NDF = 2;
                 }
--- a/sbncode/HitFinder/GaussHitFinderSBN_module.cc
+++ b/sbncode/HitFinder/GaussHitFinderSBN_module.cc
@@ -497,7 +497,8 @@ namespace hit {
                 }
 
                 // Another protection: if peak is outside of the range, 
-                // we skip this hit: this should however be avoided before... see LL407-411
+                // skip the hit creation
+                // For details on this issue refer to SBN DocDB 47092
                 if ((peakMean < startT) || (peakMean >= endT)) continue;
 
                 // Extract errors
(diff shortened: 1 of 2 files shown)


mattiasotgia:
Adding @linyan-w as reviewer for SBND, adding this to https://github.com/orgs/SBNSoftware/projects/51

cerati (approved):
Please remove the " // THIS IS NOT USED" lines? or if there is a specific need for them, please explain. Everything else seems fine so I will approve in the mean time

mattiasotgia:
Hi @cerati I addressed your comment. However I wanted to highlight that the comment was present since even if the hit filtering is enabled, filtered hits are not saved (since the lines 645--649 are commented out), so maybe we might need to check how this might be enabled correctly, to align it to the `larreco` module

sbncode/HitFinder/GaussHitFinderSBN_module.cc
```diff
@@ -388,8 +388,8 @@ namespace hit {
                 fPeakFitterTool->findPeakParameters(
                   range.data(), mergedCands, peakParamsVec, chi2PerNDF, NDF);
 
-                // If the chi2 is infinite then there is a real problem so we bail
-                if (!(chi2PerNDF < std::numeric_limits<double>::infinity())) {
+                // If the chi2 is not finite then there is a real problem so we bail
+                if (!std::isfinite(chi2PerNDF)) {
```
PetrilloAtWork, line 392: 💯

sbncode/HitFinder/GaussHitFinderSBN_module.cc
```diff
@@ -550,6 +551,9 @@ namespace hit {
 
                 if (HitsumEndItr > sumEndItr) HitsumEndItr = sumEndItr;
 
+                // This prevents L577 and L579 to make any possible boundaty flip
```
PetrilloAtWork: Lines 577 and 579 are two arguments of a function call...
That is: make this description less fragile (as a last resort, you make even consider putting comment markers, `// COMMENT [1] below` or equivalent).
Also, there is a typo on "boundaty".

sbncode/HitFinder/GaussHitFinderSBN_module.cc
```diff
@@ -499,6 +496,10 @@ namespace hit {
                   continue;
                 }
 
+                // Another protection: if peak is outside of the range, 
+                // we skip this hit: this should however be avoided before... see LL407-411
```
PetrilloAtWork: Same comment on the number of lines as below (I had missed this first because of GitHub wrapping the line).
If you have a presentation describing the problem (which I think you have plenty), pick the more appropriate and cite it here (like `// see SBN DocDB 12345`).

PetrilloAtWork (commented):
A couple of comments to help documentation and future interpretation.
Technically it looks good to me.

kjplows:
I remember the gotchas addressed by points 2 and 3.. The changes look good to me, approved! Thanks @mattiasotgia  :)

kjplows (approved):

sbncode/HitFinder/GaussHitFinderSBN_module.cc
mattiasotgia: 🫡

mattiasotgia:
Addressed @PetrilloAtWork's comments (thank you 🙏🏾). I've added a couple of lines regarding this issue (and a reference to the presentation where the issue is presented, [SBN DocDB-47092](https://sbn-docdb.fnal.gov/cgi-bin/sso/ShowDocument?docid=47092)

SFBayLaser (approved):
I'm happy to approve, have seen multiple presentations on this, the code changes are consistent with those presentations and, finally, it looks like Mattia has addressed other reviewer requests.

PetrilloAtWork (approved):
Thank you, @mattiasotgia.

PR #660, opened 2026-07-09
Updates to Enable ICARUS WireMod Closure

gputnam:


sbncode/WireMod/Utility/WireModUtility.cc
```diff
@@ -692,45 +1130,37 @@ void sys::WireModUtility::ModifyROI(std::vector<float> & roi_data,
...
     // do some sanity checks
-    if        (isnan(q_orig))
-    {
-      if (verbose)
-        std::cout << "WARNING: obtained q_orig = NaN... setting scale to 1" << std::endl;
-      scale_ratio = 1.0;
-    } else if (isnan(q_mod)) {
+    if (isnan(q_mod)) {
       if (verbose)
         std::cout << "WARNING: obtained q_mod = NaN... setting scale to 0" << std::endl;
```
francescopoppi, line 1144: do we want a guard that sets scaling factors to 1.0 instead of 0? Probably we do

francescopoppi (changes requested):
" I think there is the need to add a guard to nan scaling factors with fallback scaling factors to 1. The finger points to Francesco to do it. Ok, I am on it, early US morning I will fix it. "
This was my old comment, but after going through the rabbit hole, q_nan is an implicit fallback to scaling factors of one, simply it does not scale the rois.
Besides this, I really would like to wait Thomas or SBND review to make sure we are not butchering their WireMod implementation. We should not, as the biggest changes come from additional parameters which are defaulted to current SBNcode implementation and are updated only through icarus fcls.

francescopoppi (approved):

PR #663, opened 2026-07-15
WireMod Closure Update for sbncode develop

gputnam:


sbncode/HitFinder/ChannelROIToWire_module.cc
```diff
@@ -77,6 +78,7 @@ void ChannelROIToWire::reconfigure(fhicl::ParameterSet const& pset)
     fWireModuleLabelVec    = pset.get<std::vector<art::InputTag>>("WireModuleLabelVec",   std::vector<art::InputTag>()={"decon1droi"});
     fOutInstanceLabelVec   = pset.get<std::vector<std::string>>  ("OutInstanceLabelVec",                            {"PHYSCRATEDATA"});
     fDiagnosticOutput      = pset.get< bool                     >("DiagnosticOutput",                                           false);
+    fUseFloatPrecision     = pset.get< bool                     >("UseFloatPrecision",                                          false);
```
hausnerh, line 81: Not a stopping comment, but do we want the float precision to default to `false`? I like having it as a fhicl parameter, but we should make sure that the defaults match the most common use-case

hausnerh (approved):
looks good to me

PR #667, opened 2026-07-21
Add elastic dials for all g4reweight particles

kyjung123:
Just add elastic dials in g4 reweight fcl for all particles.

### DIFF
--- a/sbncode/SBNEventWeight/jobs/geant4/kminus_reweight_parameters.fcl
+++ b/sbncode/SBNEventWeight/jobs/geant4/kminus_reweight_parameters.fcl
@@ -1,12 +1,20 @@
 BEGIN_PROLOG
 
-## KMinus available process keys: "total"
+## KMinus available process keys: "total inelastic", "elastic"
 
 KMinusParameters: [
   {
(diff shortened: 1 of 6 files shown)


pgreen135:
Another change that could be made at the same time - for the pions, there's a split in energy range for the inelastic fPiMinusInelLow/fPiMinusInelHigh. I copied this from the configuration in MicroBooNE as we needed these in the code ASAP at the time,  but it was never clear to me _why_ this was done; I wasn't able to find any records in old uboone slides. It may be worth investigating further and fixing this at the same time. Or at least, noting it as a potential issue to be addressed in the future

kyjung123:
Merge the two pi+ inel params into one, [10, 2005] MeV/c, replacing fPiPlusInelLow ([10, 200]) and fPiPlusInelHigh ([700, 2005]).
The original split came from the Geant4Reweight paper, they only put params where pi-nucleus data existed, so low and high momentum with nothing in ~200-700 MeV/c. For our reweighting that means no inel param covers the middle region.
##Note
I kept Sigma at 0.2 here. We might want to revisit it later based on the LArIAT and ProtoDUNE data.

